PR #7, opened 2018-05-18
Adds "no __author__" validator

vergeev:
Implements the rule `no __author__`.

Note that it does not protect against the case
```
if True:
    __author__ = 'John'
```
because this is simply a strange thing to write (and it would complicate the logic for no good reason).

I've created a set `BAD_MODULE_METADATA_VARIABLES`, just in case other variables need to be added (such as `__version__` or `__copyright__`).

### DIFF
--- a/tests/fixtures/wrong_variable.py
+++ b/tests/fixtures/wrong_variable.py
@@ -49,4 +49,4 @@ val = Fixture()  # error here
 print(val.var)  # no error here
 
 if val:
-    __author__ = 'John'  # no error here since it's a rare use of module metsta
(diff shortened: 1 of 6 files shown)


sobolevn:
@vergeev thanks for implementing this! 

I feel a little bit ashamed for such a low quality docs and tasks for this project.
Since it is an internal beta tool. But, I am happy that you are interested in contributing.

I also agree that `__author__` inside any other blocks should be allowed. 
I can see perfectly valid cases for this variable inside `class`es.

Thanks for contributing!

PR #29, opened 2018-07-01
Add configuration file parser

### LINKED ISSUE #9: Create configuration
Currently our linter is full of magic constants like: 
- https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/visitors/high_complexity.py#L44
- https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/visitors/high_complexity.py#L58
- and others marked with `TODO: config` in the source code

What we need is:
0. create a file with default configuration values
1. allow users to override these values with `setup.cfg` or any other standard `flake8` tools
2. test that when configuration is applied linter respects it

mcproger:
Closes #9


sobolevn:
@mcproger awesome, thanks!

Can you please rebase your branch and fix all the linting issues?
After I will love to review this feature.

mcproger:
Sorry for this mistake. Now it should be OK.

wemake_python_styleguide/config.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+
+import configparser
+from pathlib import Path
+from typing import Optional
+
+BASE_DIR = Path(__file__).resolve().parent.parent
```
sobolevn: Sorry, but that's not how it is supposed to work.
You can have different setups and `flake8` has API to get the job done.

See, how it is implemented in a similar project: https://github.com/zheller/flake8-quotes/blob/master/flake8_quotes/__init__.py#L78-L146

README.md
```diff
@@ -38,3 +38,19 @@ poetry run doc8 -q docs
 ```
 
 It's OK if some tests are skipped.
+
+
+## Configuration file
```
sobolevn: We also need to support `cli` usage. See https://github.com/zheller/flake8-quotes/blob/master/flake8_quotes/__init__.py#L99

setup.cfg
```diff
@@ -22,6 +22,13 @@ inline-quotes = single
 accept-encodings = utf-8
 
 
+[wemake_python_styleguide]
```
sobolevn: I feel like we don't need these values here.
Because we should use default values by default in a project itself.

There's no need to duplicate them.

tests/test_config.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
+
+from wemake_python_styleguide.config import ConfigFileParser
+
+
+def test_config_parsing(absolute_path):
```
sobolevn: Nice! But we also need to test `cli` values.
That can be done in the similar way as we test our code with `Popen`.

tests/test_config.py
```diff
@@ -0,0 +1,11 @@
...
+    """Test to check that config parsing is correct."""
+    filename = absolute_path('fixtures', 'setup.cfg')
+    config_parser = ConfigFileParser(filename)
+    some_option = config_parser.get_option('some-option')
```
sobolevn: Let's make it not `some_option`, but something realistic, like: `max-local-variables`.

wemake_python_styleguide/config.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class ConfigFileParser(object):
+    """This class parses configs from `setup.cfg` file."""
+
+    CONFIG_FILE_NAME = 'setup.cfg'
```
sobolevn: We should not rely on this value, so it should be removed. We do not take part inn parsing config file ourself, `flake8` does it for us.

mcproger:
Thank you for the review. I will work on this.

wemake_python_styleguide/checker.py
```diff
@@ -56,6 +69,8 @@ def run(self) -> Generator[CheckResult, None, None]:
         """
         for visitor_class in self._visitors:
             visiter = visitor_class()
+            if hasattr(visiter, 'provide_options'):
+                visiter.provide_options(self.options)
```
sobolevn: The only thing that bothers me here is: why this API is not required?

We can define a basic `provide_options` call in `BaseVisiter` class.
And just call it.

setup.cfg
sobolevn: Please, don't forget to remove these values.

sobolevn: @mcproger

mcproger:
My mistake, I'll fix it as soon as I can.

sobolevn:
Awesome, thanks!

Issue #173, opened 2018-10-01
Feature: test async list comprehensions 


sobolevn:
Currently [`WrongListComprehensionVisitor`](https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/visitors/ast/keywords.py#L69) is not tested with `async for` comprehensions. 
It might work, it might not work.

What needs to be done?
1. We need to provider new unit tests for the `async for` cases here: https://github.com/wemake-services/wemake-python-styleguide/blob/master/tests/test_visitors/test_ast/test_general/test_wrong_keyword/test_comprehensions.py
2. I am pretty sure that we do not need to copy paste all data, just use something like: `'nodes = [node {0}for node in 'abc']'` and then format this string with either `'async '` or `''`

AlwxSin:
It is not that simple.
Ast can't parse single async comprehension without async function wrapped around. 
```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "python3.6/ast.py", line 35, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 2
    nodes = [node async for node in 'abc' if node != 'a' if node != 'b' if node != 'c']
                      ^
SyntaxError: invalid syntax
```

@sobolevn how about wrap all comprehensions with something like
```python
async_wrapper = """
async def d():
    {comprehension}
    """

list_ifs_multiple = """
nodes = [node {async}for node in 'abc' if node != 'a' if node != 'b' if node != 'c']
"""
```

sobolevn:
@AlwxSin hm, I have totally missed it! 

Yes, your solution seems reasonable. Let's do that. Are you willing to put this together in a PR?

AlwxSin:
Yeah, I'll do it.

PR #179, opened 2018-10-02
Test async list comprehensions

AlwxSin:
wemake-services/wemake-python-styleguide#173


AlwxSin:
@sobolevn because I used `.format` I escaped literals `{` and `}` in dict and set comprehensions.
Is it ok? Or I should use any other solution?

AlwxSin:
Sorry, I forgot to sync repo before making a pull-request. Close and open a new one.

PR #181, opened 2018-10-02
Test async list comprehensions

AlwxSin:
Closes wemake-services/wemake-python-styleguide#173

### DIFF
--- a/tests/test_visitors/test_ast/test_general/test_wrong_keyword/test_comprehensions.py
+++ b/tests/test_visitors/test_ast/test_general/test_wrong_keyword/test_comprehensions.py
@@ -10,75 +10,79 @@ from wemake_python_styleguide.visitors.ast.keywords import (
 )
 
 # Lists:
+async_wrapper = """
+async def f():
+    {comprehension}
+"""
 
 list_ifs_multiple = """
-nodes = [node for node in 'abc' if node != 'a' if node != 'b' if node != 'c']
+nodes = [node {async_stmt}for node in 'abc' if node != 'a' if node != 'b' if node != 'c']
 """
 
 list_ifs_twice = """
-nodes = [node for node in 'abc' if node != 'a' if node != 'b']
+nodes = [node {async_stmt}for node in 'abc' if node != 'a' if node != 'b']
 """
 
 list_ifs_single = """
-nodes = [node for node in 'abc' if node != 'a']
+nodes = [node {async_stmt}for node in 'abc' if node != 'a']
 """
 
 list_without_ifs = """
-nodes = [node for node in 'abc']
+nodes = [node {async_stmt}for node in 'abc']
 """
 
 # Dicts:
 
 dict_ifs_multiple = """
-nodes = {xy: xy for xy in 'abc' if xy != 'a' if xy != 'b' if xy != 'c'}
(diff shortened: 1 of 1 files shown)


AlwxSin:
@sobolevn There are several questions:
1) About style guide. Three comprehensions exceeded 80 symbols limit. I'm not sure how It can changed. Or can I set `noqa` here?
2) About passing 'async ' and '' directly. I'm not sure how do you want me to do that. If it's async comprehension we should do two format operations, not single one.
3) About fixture. Should it be fixture that returns text as is or should it be some kind of factory?

sobolevn:
1. Nope, there's a valid way to rewrite the following lines:

```python
list_ifs_multiple = """
nodes = [
     node 
     {async_stmt}for node in 'abc' 
     if node != 'a' if node != 'b' if node != 'c'
]
"""
```

2. Yes, I have forgotten about the wrapper again 😕 
3. Let's leave the fixture out of scope for now. I am not sure about it yet.

tests/test_visitors/test_ast/test_general/test_wrong_keyword/test_comprehensions.py
```diff
@@ -92,13 +96,21 @@
...
     assert_errors,
     parse_ast_tree,
     code,
+    is_async,
     default_options,
 ):
     """Testing that using `if` keyword is allowed."""
+    if is_async:
+        code = code.format(async_stmt='async ')
+        code = code.lstrip()  # remove heading \n symbol
```
sobolevn: Why do you think that this line is required?
We are already using `dedent` here: https://github.com/wemake-services/wemake-python-styleguide/blob/master/tests/test_visitors/test_ast/conftest.py#L33

Or is there any specific problem that I can not see?

AlwxSin: `dedent` doesn't remove empty lines.
Example
```python
>>> from textwrap import dedent
>>> example = """
... node = ...
... """
>>> dedent(example)
'\nnode = ...\n'
```

If you want to avoid empty line there should be backslash. 
Example from [docs](https://docs.python.org/3.7/library/textwrap.html#textwrap.dedent)
```python
def test():
    # end first line with \ to avoid the empty line!
    s = '''\
    hello
      world
    '''
    print(repr(s))          # prints '    hello\n      world\n    '
    print(repr(dedent(s)))  # prints 'hello\n  world\n'
```

sobolevn: Why should we care about the new lines? Does it break something?

AlwxSin: Yeah.
There are two new lines after `def`
```python
async_wrapper.format(comprehension=dedent(list_ifs_multiple))
"\nasync def f():\n    \nnodes = [\n    node\n    {async_stmt}for node in 'abc'\n    if node != 'a' if node != 'b' if node != 'c'\n]\n\n"
```

AlwxSin: Also for functions we need proper indentations. And all options that I tried were ugly.

sobolevn:
@AlwxSin thanks. Awesome!

PR #182, opened 2018-10-02
Closes #177: count async function as a module member

### LINKED ISSUE #177: Feature: count async function as a module member
Currently we count only regular functions and classes as module members: https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/visitors/ast/complexity/counts.py#L40-L60

What needs to be done?
1. We need to count `AsyncFunctionDef` as well
2. We need to improve `ModuleMembers` type to include `AsyncFunctionDef`
3. We need to refactor how handler for each node works: there's no need to duplicate the logic, so it can just be: `visit_AsyncFunctionDef = visit_FunctionDef = visit_ClassDef`
4. We need to improve unit tests to include `async` functions as well: https://github.com/wemake-services/wemake-python-styleguide/blob/master/tests/test_visitors/test_ast/test_complexity/test_counts/test_module_counts.py

novikovfred:
1. Count `AsyncFunctionDef` in Modules.
2. Improved `ModuleMembers` type, added `AsyncFunctionDef`.
3. Make base function for all types: `visit_AsyncFunctionDef = visit_FunctionDef = visit_ClassDef = visit_module_members`
4. Added to unit tests async functions.

### DIFF
--- a/tests/test_visitors/test_ast/test_complexity/test_counts/test_module_counts.py
+++ b/tests/test_visitors/test_ast/test_complexity/test_counts/test_module_counts.py
@@ -13,6 +13,12 @@ def first(): ...
 class Second(object): ...
 """
 
+module_with_async_function_and_class = """
+async def first(): ...
+
+class Second(object): ...
+"""
+
 module_with_methods = """
 class First(object):
     def method(self): ...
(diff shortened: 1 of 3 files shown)


sobolevn:
@novikovfred thanks a lot!

Issue #176, opened 2018-10-01
Feature: support async nesting level


sobolevn:
When calculating offset we are missing some `async` nodes like:
1. `ast.AsyncFor`
2. `ast.AsyncWith`

They should be added here: https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/visitors/ast/complexity/offset.py

We also need to unit test the changes: https://github.com/wemake-services/wemake-python-styleguide/blob/master/tests/test_visitors/test_ast/test_complexity/test_offset_visitor.py

PR #185, opened 2018-10-02
Issue #176. Support check nesting level for ast.AsyncFor and ast.AsyncWith

TyVik:
Closes https://github.com/wemake-services/wemake-python-styleguide/issues/176


wemake_python_styleguide/visitors/ast/complexity/offset.py
```diff
@@ -23,6 +23,8 @@
     'visit_Break',
     'visit_Assign',
     'visit_Expr',
+    'visit_AsyncFor',
```
sobolevn: Sorry, but  it seems that I have forgotten about `visit_Await`. Can you please add it here and write a unit test for `await something()`?

PR #186, opened 2018-10-02
Issue #175. `async` stuff for check variable names.

### LINKED ISSUE #175: Feature: test async structures naming
We do not check any names in `async` members.

What we need to do:

0. Implement missing support for `async` stuff
1. Add unit tests for `async def` for function names and argument names
2. Add unit tests for `async for` and `async` comprehension variable names
3. Add unit tests for `async with` variable names

Tests are located here: https://github.com/wemake-services/wemake-python-styleguide/tree/master/tests/test_visitors/test_ast/test_general/test_wrong_name

We do not need to test `async` code as much as we test the regular one in case it will share the same logic. We can only write one test case for each violation and good path.

TyVik:
Closes #175

### DIFF
--- a/wemake_python_styleguide/constants.py
+++ b/wemake_python_styleguide/constants.py
@@ -8,7 +8,6 @@ It also contains some exceptions that we allow to use in our codebase.
 """
 
 import re
-import sys
 
 # TODO: use consistent `.` for the `#:` comments
 # TODO: use consistent names: `*_BLACKLIST` and `*_WHITELIST`
--- a/wemake_python_styleguide/constants.py
+++ b/wemake_python_styleguide/constants.py
@@ -51,7 +51,7 @@ BAD_MODULE_METADATA_VARIABLES = frozenset((
     '__about__',
 ))
 
-
+#: List of variable names we forbid to use.
 BAD_VARIABLE_NAMES = frozenset((
     # Meaningless words:
     'data',
--- a/wemake_python_styleguide/constants.py
+++ b/wemake_python_styleguide/constants.py
@@ -53,7 +53,7 @@ BAD_MODULE_METADATA_VARIABLES = frozenset((
 ))
 
 
-_BAD_VARIABLE_NAMES = [
+BAD_VARIABLE_NAMES = frozenset((
     # Meaningless words:
     'data',
     'result',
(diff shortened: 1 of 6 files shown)


wemake_python_styleguide/constants.py
```diff
@@ -53,7 +52,7 @@
 ))
 
 
-_BAD_VARIABLE_NAMES = [
+BAD_VARIABLE_NAMES = frozenset((
```
sobolevn: You have missed a docsting here: `#: List of variable names we forbid to use.`

wemake_python_styleguide/visitors/ast/naming.py
```diff
@@ -75,7 +79,10 @@ def visit_Attribute(self, node: ast.Attribute) -> None:
 
         self.generic_visit(node)
 
-    def visit_FunctionDef(self, node: ast.FunctionDef) -> None:
+    def visit_FunctionDef(
+        self,
+        node: Union[ast.FunctionDef, ast.AsyncFunctionDef],
```
sobolevn: We already have a type for it: https://github.com/wemake-services/wemake-python-styleguide/blob/3cedeb3c13ab6b16980a39edf657ab93d4c1f19e/wemake_python_styleguide/visitors/ast/complexity/counts.py#L20

Can you please move it to `types.py` and reuse it?
A comment docstring will also be required.

wemake_python_styleguide/visitors/ast/naming.py
```diff
@@ -79,10 +78,7 @@ def visit_Attribute(self, node: ast.Attribute) -> None:
 
         self.generic_visit(node)
 
-    def visit_FunctionDef(
-        self,
-        node: Union[ast.FunctionDef, ast.AsyncFunctionDef],
-    ) -> None:
+    def visit_FunctionDef(self, node: MethodMembers) -> None:
```
sobolevn: One last thing. This is not quite clean to declare a `FunctionDef` handler that also visits `AsyncFunctionDef`. It violates the design idea of `ast.NodeVisitor`.

What can be done instead? We can create a proxy handler called `visit_any_function` and alias it to support both function types.
Example link: https://github.com/wemake-services/wemake-python-styleguide/blob/5539d0e873db7dc7b841fc0242b23b1fa329af7a/wemake_python_styleguide/visitors/ast/complexity/counts.py#L89-L92

TyVik: My bad, did not notice... Fixed.

PR #203, opened 2018-10-03
Issue #172, feature: count async functions complexity

### LINKED ISSUE #172: Feature: count async functions complexity
As one can see currently we do not check `async` functions for complexity: https://github.com/wemake-services/wemake-python-styleguide/blob/6fb64c00213dcc7e35f1944fd8576220e59ee49c/wemake_python_styleguide/visitors/ast/complexity/function.py#L149

But, we definitely need to! 

What need's to be done:
1. We need to create `visit_AsyncFunctionDef` method in the same visitor
2. It should be equal to `visit_AsyncFunctionDef = visit_FunctionDef`
3. We need to fix all types that are currently used there to support both function
4. We need to add unit tests for `async` functions in each of these files: https://github.com/wemake-services/wemake-python-styleguide/tree/master/tests/test_visitors/test_ast/test_complexity/test_function

That can be simplified by doing something like this (https://github.com/wemake-services/wemake-python-styleguide/blob/master/tests/test_visitors/test_ast/test_complexity/test_function/test_returns.py#L12-L17):

```python
function_with_returns = """
{0} def function():
    if 1 > 2:
        return 1
    return 0
"""
```

And then conditionally formatting this code:

```python
@pytest.mark.parametrize('code', [
    function_without_returns,
    function_with_returns,
])
@pytest.mark.parametrize('mode', [
    'async',  # coroutine 
    '',  # regular function
])
def test_some(code, mode):
     tree = parse_ast_tree(code.format(mode))
```

5. We need to add new `async` functions to the e2e tests here: https://github.com/wemake-services/wemake-python-styleguide/tree/master/tests/fixtures/config

novikovfred:
# I have made things!

- Rename `MethodMembers` to `AnyFunctionDef`
- Count async functions complexity

## Checklist

## Related issues

- Closes #172

### DIFF
--- a/tests/test_visitors/test_ast/test_complexity/test_function/test_elifs.py
+++ b/tests/test_visitors/test_ast/test_complexity/test_function/test_elifs.py
@@ -8,7 +8,7 @@ from wemake_python_styleguide.visitors.ast.complexity.function import (
 )
 
 function_with_elifs = """
-def test_module():
+{0} def test_module():
     if 1 > 2:
         print(1)
     elif 2 > 3:
@@ -20,7 +20,7 @@ def test_module():
 """
 
 function_with_ifs = """
-def test_module():
+{0} def test_module():
     if True:
         print(1)
     if 2 > 3:
@@ -30,13 +30,13 @@ def test_module():
 """
 
 function_with_raw_if = """
-def function():
+{0} def function():
     if 1 == 2:
         print(1)
 """
 
 function_with_if_else = """
-def function(param):
+{0} def function(param):
     if param == 2:
         print(1)
     else:
@@ -44,7 +44,7 @@ def function(param):
 """
 
 function_with_ternary = """
-def with_ternary(some_value):
+{0} def with_ternary(some_value):
     return [some_value] if some_value > 1 else []
 """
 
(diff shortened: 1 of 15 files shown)


tests/test_visitors/test_ast/test_complexity/test_function/test_elifs.py
```diff
@@ -8,7 +8,7 @@
 )
 
 function_with_elifs = """
-def test_module():
+{0} def test_module():
```
sobolevn: It should be `{0}def`. Without an extra space.
The same applies to all other fixtures.

tests/test_visitors/test_ast/test_complexity/test_function/test_elifs.py
```diff
@@ -56,11 +56,15 @@ def with_ternary(some_value):
     function_with_if_else,
     function_with_ternary,
 ])
+@pytest.mark.parametrize('mode', [
+    'async',  # coroutine
```
sobolevn: It should be `'async '` with an extra space.

novikovfred:
@sobolevn Yep, stupid mistakes, sry.

novikovfred:
@sobolevn done

PR #221, opened 2018-10-05
Add check for comparison between two literals

### LINKED ISSUE #218: Forbid only constants in comparison
# Rule request

## Thesis

We need to forbid constants in some specific cases like:

```python
if 9: ...
if 1 > 1: ...
if False: ...
if 9 > 0 and 4 < 2:
if 3 < 5 < 10: ...
```

It is clear that is does not make any sense. And it was made by mistake. 

Related:
- https://help.semmle.com/wiki/display/PYTHON/Comparison+of+constants
- https://help.semmle.com/wiki/display/PYTHON/Constant+in+conditional+expression+or+statement

tipabu:
Closes #218

Note that this does not include checks for literals in conditionals.

# I have made things!

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ We used to have incremental versioning before `0.1.0`.
 ### Features
 
 - Forbids to use `quit` and `exit` functions
+- Forbids the comparison of two literals
 
 ### Misc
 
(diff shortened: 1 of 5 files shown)


sobolevn:
@tipabu please, also add your visitor into the preset in `presets/general.py`.
Otherwise it won't be included into the checker.

PR #231, opened 2018-10-09
Fix flaking test in test_module_names.

lensvol:
Our code always expects Checker class to have options attribute set by flake8. Since we call flake8 by using an external helper, it usually behaves as expected and all is well. But in the case of test_module_names, we are instantiating Checker instance by ourselves without calling parse_options() first, which leads to AttributeError.

It is possible not to see this bug due to pytest skipping some tests, but it is easy to trigger by introducing new unrelated changes into the testing code and change cached coverage, leading to pytest deciding to run this test again.

# I have made things!

## Checklist


lensvol:
Since changes are in the test case itself, I've decided not to write an additional test case for this test case.

PR #233, opened 2018-10-10
Fix `TooManyMethodsViolation` not displaying line number.

lensvol:
When `TooManyMethodsViolation` was triggered, flake8 showed it happening at line number 0. This was due to this violation being SimpleViolation, instead of a more appropriate ASTViolation. It may have occured because other violations in the same packageare module-level ones and therefore it is sufficient for them to be `simple`, without pointing to particular node.

I am not sure how to proceed with tests here, seeing that it may involve breaking current philosophy of using provided `assert_errors` helper, since it checks only class of the violation without looking at the node contents.

# I have made things!

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ to the project during `#hactoberfest`. List of awesome people:
 
 - Fixes that `MultipleIfsInComprehensionViolation` was not enabled
 - Fixes flaky behaviour of test_module_names
+- Fixed `TooManyMethodsViolation` not displaying line number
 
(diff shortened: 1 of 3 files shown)


sobolevn:
Thanks 👍

sobolevn:
@lensvol btw, do you use this linter in production in Yandex?

lensvol:
> @lensvol btw, do you use this linter in production in Yandex?

No idea, sorry :) I've left Yandex 4 months ago, working at another company now.

PR #242, opened 2018-10-12
Added inspection for class-level attributes containing upper-class letters

lensvol:
"noqa: Z214" is needed on WrongNameVisitor since I could not find a way to implement this without adding ClassDef-specific handler to the class, which in turn raised method count over the limit.

I could have maintaned a "last visited node" context inside the visitor and trigger check in standard place, but it would not be explicit enough in my opinion.

Previous pull request was closed because I forgot to run mypy, which in turn required me to add some additional `isinstances` to facilitate type inference.

P.S. There is also a problem with variables inside code comprehensions being counted as locals, which in turn triggers *Z210*. Will fix it in separate PR if I have the time or just create an issue.

# I have made things!

## Checklist

### DIFF
--- a/tests/test_visitors/test_ast/test_naming/test_class_attributes.py
+++ b/tests/test_visitors/test_ast/test_naming/test_class_attributes.py
@@ -44,7 +44,7 @@ def test_wrong_attributes_names(
     assert_errors(visitor, [WrongVariableNameViolation])
 
 
-@pytest.mark.parametrize('short_name', string.ascii_letters)
+@pytest.mark.parametrize('short_name', string.ascii_lowercase)
 @pytest.mark.parametrize('code', [
     static_attribute,
     instance_attribute,
@@ -53,7 +53,7 @@ def test_too_short_attribute_names(
     assert_errors, parse_ast_tree, short_name, code, default_options,
 ):
     """Testing that attribute can not have too short names."""
-    tree = parse_ast_tree(code.format(short_name.lower()))
+    tree = parse_ast_tree(code.format(short_name))
 
     visitor = WrongNameVisitor(default_options, tree=tree)
     visitor.run()
--- a/tests/test_visitors/test_ast/test_naming/test_class_attributes.py
+++ b/tests/test_visitors/test_ast/test_naming/test_class_attributes.py
@@ -7,6 +7,7 @@ import pytest
 from wemake_python_styleguide.violations.naming import (
     PrivateNameViolation,
     TooShortVariableNameViolation,
+    UpperCaseAttributeViolation,
     WrongVariableNameViolation,
 )
 from wemake_python_styleguide.visitors.ast.naming import (
@@ -52,7 +53,7 @@ def test_too_short_attribute_names(
     assert_errors, parse_ast_tree, short_name, code, default_options,
 ):
     """Testing that attribute can not have too short names."""
-    tree = parse_ast_tree(code.format(short_name))
+    tree = parse_ast_tree(code.format(short_name.lower()))
 
     visitor = WrongNameVisitor(default_options, tree=tree)
     visitor.run()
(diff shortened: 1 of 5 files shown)


wemake_python_styleguide/logics/variables.py
```diff
@@ -40,6 +40,28 @@ def is_wrong_variable_name(name: str, to_check: Iterable[str]) -> bool:
     return False
 
 
+def is_upper_case_name(name: str):
+    """
+    Checks that attribute name has no upper-case letters.
+
+    >>> is_upper_case_name('camelCase')
+    True
```
sobolevn: We utilize a single empty line between doctest examples.

wemake_python_styleguide/violations/naming.py
```diff
@@ -476,3 +478,30 @@ class UnderScoredNumberNameViolation(SimpleViolation):
     #: Error message shown to the user.
     error_template = 'Found underscored name pattern "{0}"'
     code = 114
+
+
+class UpperCaseAttributeViolation(ASTViolation):
+    """
+    Forbids to use anything but snake_case for naming attributes on a class.
+
+    Reasoning:
```
sobolevn: Can you please propose a solution for this violation?

wemake_python_styleguide/visitors/ast/naming.py
```diff
@@ -126,6 +128,38 @@ def visit_Name(self, node: ast.Name) -> None:
...
+        """
+        Used to find upper attribute declarations.
+
+        Raises:
+            WrongVariableNameViolation
+            TooShortVariableNameViolation
+            PrivateNameViolation
+
+        """
+        top_level_assigns = [
```
sobolevn: Can you please move all the business logic into a separate method?

sobolevn: I have considered your suggestion, but I prefer to be consistent.

wemake_python_styleguide/visitors/ast/naming.py
```diff
@@ -126,6 +128,38 @@ def visit_Name(self, node: ast.Name) -> None:
...
+            sub_node
+            for sub_node in node.body
+            if isinstance(sub_node, ast.Assign)
+        ]
+
+        upper_case_violations = [
+            target
```
sobolevn: Can you please open a new issue for `counting comprehensions as locals`?

wemake_python_styleguide/logics/variables.py
lensvol: Will fix it.

wemake_python_styleguide/visitors/ast/naming.py
lensvol: So I'll just leave `visit_ClassDef` around call to that separate method?

wemake_python_styleguide/visitors/ast/naming.py
lensvol: I'm going to, didn't have time when I discovered it. :)

wemake_python_styleguide/violations/naming.py
lensvol: Okay. :)

wemake_python_styleguide/visitors/ast/naming.py
lensvol: Here it is: https://github.com/wemake-services/wemake-python-styleguide/issues/247

tests/test_visitors/test_ast/test_naming/test_class_attributes.py
```diff
@@ -52,7 +53,7 @@ def test_too_short_attribute_names(
     assert_errors, parse_ast_tree, short_name, code, default_options,
 ):
     """Testing that attribute can not have too short names."""
-    tree = parse_ast_tree(code.format(short_name))
+    tree = parse_ast_tree(code.format(short_name.lower()))
```
sobolevn: It is better to use https://docs.python.org/2/library/string.html#string.ascii_lowercase here

lensvol: Okay.

PR #243, opened 2018-10-12
Issue 219

### LINKED ISSUE #219: Forbid to compare the identical variables
# Rule request

## Thesis

We need to find cases like this: `if some_value == some_value:`.
We also need to find other operators: `is`, `is not`, `>`, `<`, `!=`, `>=`, `<=`.

## Reasoning

It clearly does not make any sense. It is a bug in the source code.

Related: https://help.semmle.com/wiki/display/PYTHON/Comparison+of+identical+values

SheldonNunes:
# I have made things!

## Checklist

## Related issues
Closes #219

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ to the project during `#hactoberfest`. List of awesome people:
 - Forbids the comparison of two literals
 - Forbids the incorrect order comparison
 - Forbids underscores before numbers in names
+- Forbids comparison of the same variable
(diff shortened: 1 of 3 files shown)


SheldonNunes:
Cool all done. I wonder whether there should be checks for redundant if statements too 
e.g `if True` and `if False`

I can open another issue if you think it is worth pursuing :)

sobolevn:
@SheldonNunes you are ahead of out stuff. 

Can you please submit a new request for this feature?

PR #251, opened 2018-10-13
Issue #215, Forbid try with finally without except

### LINKED ISSUE #215: Forbid `try` with `finally`without `except`
# Rule request

## Thesis

We should forbid using `try` block with anything other than `except`.

## Reasoning

Related: https://help.semmle.com/wiki/display/PYTHON/Should+use+a+%27with%27+statement

Let's remember why do we use `try` with `finally`. To clean thing up when something explodes inside `try` block. That a popular pattern for files, streams, etc.

What else can be used in this case? We can use `with` to make this code readable. Consider this example:

```python
try:
    f = open("filename")
    f.write(...)
finally:
    f.close()
```

Versus context manager:

```python
with open("filename") as f:
    f.write(...)
```

The second version looks cleaner and more readable. 

## Limitations

This rule is limited to cases when there are no `except` blocks. Because when there are except blocks `finally` becomes very useful to execute any logic after both successful and exceptional branches.

novikovfred:
# I have made things!

## Checklist

- Closes #215


CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ to the project during `#hactoberfest`. List of awesome people:
 - Enforce consistent octal, binary, and hex numbers
 - We not check the argument count in Lambda functions
 - Forbid `for` loops with unused `else`
+- Forbid `try` with `finally`without `except`
```
sobolevn: Missing space here

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -615,3 +617,23 @@ class RedundantForElseViolation(ASTViolation):
     #: Error message shown to the user.
     error_template = 'Found `else` in `for` loop with `break`'
     code = 436
+
+
+class RedundantFinallyViolation(ASTViolation):
+    """
+    Forbids to use ``finally`` in ``try`` block without ``except``.
```
sobolevn:
``` 
without ``except`` block.
```

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -615,3 +617,23 @@ class RedundantForElseViolation(ASTViolation):
...
+
+    Reasoning:
+        This rule will reduce complexity and improve readability.
+
+    Solution:
+        Refactor your ``try`` logic.
```
sobolevn: How? We need to write about `with` and context managers in general.

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -615,3 +617,23 @@ class RedundantForElseViolation(ASTViolation):
...
+
```
sobolevn: You are also missing `Example::` section.
Using `open()` is a good use case to cover.

PR #260, opened 2018-10-15
Issue 244

### LINKED ISSUE #244: Forbid redundant if statements
# Rule request
## Thesis

We should prevent if statements that use `True` or `False` without any comparison

e.g. `if True: ...` and `if False: ...`

## Reasoning

These statements don't make sense. They will either cause it to execute everytime or never. This is an error in the source code and should be forbidden

SheldonNunes:
# I have made things!
## Checklist

## Related issues
Closes #244


wemake_python_styleguide/violations/consistency.py
```diff
@@ -489,10 +492,41 @@ class RedundantComparisonViolation(ASTViolation):
...
 
     should_use_text = False
     #: Error message shown to the user.
     error_template = 'Found comparison between same variable'
     code = 312
+
+
+class NamedConstantConditionalViolation(ASTViolation):
+    """
+    Forbids using if statements that use only Named Constants.
```
sobolevn: Should be ``if`` (with backticks)

sobolevn: `named constants` can be lowercase.

wemake_python_styleguide/violations/consistency.py
```diff
@@ -489,10 +492,41 @@ class RedundantComparisonViolation(ASTViolation):
...
+
+    Reasoning:
+        When a Named Constant is used in conditional
```
sobolevn: The same, let it be `named constant`

wemake_python_styleguide/visitors/ast/conditionals.py
```diff
@@ -0,0 +1,27 @@
...
+
+import ast
+
+from wemake_python_styleguide.violations.consistency import (
+    NamedConstantConditionalViolation,
+)
+from wemake_python_styleguide.visitors.base import BaseNodeVisitor
+
+
+class NamedConstantConditionalVisitor(BaseNodeVisitor):
```
sobolevn: Let's move it to `comparisons.py`

tests/test_visitors/test_ast/test_conditionals/test_named_constants.py
```diff
@@ -0,0 +1,66 @@
...
+import pytest
+
+from wemake_python_styleguide.violations.consistency import (
+    NamedConstantConditionalViolation,
+)
+from wemake_python_styleguide.visitors.ast.conditionals import (
+    NamedConstantConditionalVisitor,
+)
+
+if_statement = """
```
sobolevn: We also need to test:
1. `while` statements (`while True` is popular usecase that can be used)
2. ternary expressions
3. `if` inside the comprehensions: `[x for x in [1, 2, 3] if True]` should be forbidden

tests/test_visitors/test_ast/test_conditionals/test_named_constants.py
```diff
@@ -0,0 +1,66 @@
...
+variable = 1
+if {0}: ...
+"""
+
+
+@pytest.mark.parametrize('code', [
+    if_statement,
```
sobolevn: Add new templates here

tests/test_visitors/test_ast/test_conditionals/test_named_constants.py
```diff
@@ -0,0 +1,66 @@
...
+])
+@pytest.mark.parametrize('comparators', [
+    ('variable < 3'),
```
sobolevn: Please, remove extra `()`

tests/test_visitors/test_ast/test_conditionals/test_named_constants.py
```diff
@@ -0,0 +1,66 @@
...
+    assert_errors(visitor, [])
+
+
+@pytest.mark.parametrize('code', [
+    if_statement,
+])
+@pytest.mark.parametrize('comparators', [
+    'True',
+    'False',
+    'None',
```
sobolevn: We should also test:
1. numbers: `if 0:` and `if 4:` should always be error
2. strings: `if '':` and `if 'test':` should always be error
3. arrays and dicts (both empty and with data, including variables like: `[variable]`)

tests/test_visitors/test_ast/test_conditionals/test_named_constants.py
SheldonNunes: Are you sure that there shouldn't be support for case no 1. ?

sobolevn: @SheldonNunes I am really not sure. What do you think?

SheldonNunes: I can see some use cases for `while True` especially around polling. I was thinking wrapping the boolean value in a variable may be what we want to enforce. However I can see a situation where a variable is set to `True` and we have a break clause inside to prevent running a potential large block of code. The variable would provide no real benefit as it wouldn't be modified and could potentially confuse the reader.

sobolevn: Ok, let's leave `while` out of scope for now.

SheldonNunes:
Added changes based on your feedback :)

SheldonNunes:
Ah that is awesome was trying to figure out how I could merge them!
Updated :)

sobolevn:
Thanks a lot! That is awesome!

PR #261, opened 2018-10-15
(#170) - Require space between keyword and opening parenthesis

### LINKED ISSUE #170: Require a space between keywords and parentheses
# Rule request

## Thesis

We should enforce this style:

```python
return (some, tuple, items)
```

Over this style:

```python
return(some, tuple, items)
```

Also fine:

```python
return some, tuple, items
```

## Reasoning

Some people use `return` and `yield` keywords as functions. The same happened to good old `print`.

We should enforce consistent parentheses usage with keywords.
I am not sure about which keywords are affected by this. One of the tasks is to try all keywords and find out which are affected.

valignatev:
## Checklist

## Related issues
- Closes #170

### DIFF
--- a/tests/test_visitors/test_tokenize/test_wrong_keywords/test_keywords_spaces.py
+++ b/tests/test_visitors/test_tokenize/test_wrong_keywords/test_keywords_spaces.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+import sys
+
 import pytest
 
 from wemake_python_styleguide.violations.consistency import (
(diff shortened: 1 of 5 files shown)


tests/test_visitors/test_tokenize/test_wrong_keywords/test_keywords_spaces.py
```diff
@@ -0,0 +1,57 @@
...
+    WrongKeywordTokenVisitor,
+)
+
+
+@pytest.mark.parametrize('code', [
+    'del(a)',
+    'for(a, b) in [(1, 2)]:',
+    'for (a, b) in((1, 2)):',
+    'if a in(1, 2, 3):',
+    'assert(\n'
```
sobolevn: Please, use multiline examples. Create a new variable for it.

tests/test_visitors/test_tokenize/test_wrong_keywords/test_keywords_spaces.py
```diff
@@ -0,0 +1,57 @@
...
+    '    1, 2, 3,\n'
+    ') in b',
+    'def foo():\n'
```
sobolevn: The same goes for this example: it should be a named example.

tests/test_visitors/test_tokenize/test_wrong_keywords/test_keywords_spaces.py
```diff
@@ -0,0 +1,57 @@
...
+    visitor.run()
+
+    assert_errors(visitor, [MissingSpaceBetweenKeywordAndParenViolation])
+
+
+@pytest.mark.parametrize('code', [
+    'del (a, b)',
+    'for (a, b) in (1, 2, 3)',
+    'def foo():\n'
+    '    yield (1, 2, 3)',
```
sobolevn: Please, also add both correct and incorrect examples for:
1. return
2. await
3. import
4. as
5. import
6. except
7. elif
8. else
9. in

sobolevn:
🌮

PR #279, opened 2018-10-20
Solved Issue-254: Forbid `yield` keyword in `__init__` method

Shikhov-P:
# `yield` is no longer allowed in the `__init__`

## Checklist

- [ ] I have added the code to make sure the `yield` generator can't be used in the `__init__` initializer 
- [ ] The following test cases have been covered: `def __init__` with yield, `def __init__` without yield, `async def __init__` with yield,  `async def __init__` without yield.

### DIFF
--- a/wemake_python_styleguide/violations/best_practices.py
+++ b/wemake_python_styleguide/violations/best_practices.py
@@ -40,6 +40,7 @@ Summary
    RedundantForElseViolation
    RedundantFinallyViolation
    ReassigningVariableToItselfViolation
+   YieldInsideInitViolation
 
 Comments
 --------
--- a/wemake_python_styleguide/violations/best_practices.py
+++ b/wemake_python_styleguide/violations/best_practices.py
@@ -733,8 +733,40 @@ class ReassigningVariableToItselfViolation(ASTViolation):
 
     Note:
         Returns Z438 as error code
+
     """
 
     #: Error message shown to the user.
     error_template = 'Found reassigning variable "{0}" to itself'
     code = 438
+
+
+@final
+class YieldInsideInitViolation(ASTViolation):
+    """
+    Forbids to use ``yield`` inside of ``__init__``
+
+    Reasoning:
+        ``__init__`` should be used to initialize new objects.
+        It shouldn't create any new objects by yielding them
+
+    Example::
+         # Correct:
+        class ModuleMembersVisitor(object):
+            def __init__(self, *args, **kwargs):
+                self._public_items_count = 0
+
+        # Wrong:
+        class ModuleMembersVisitor(object):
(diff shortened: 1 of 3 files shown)


wemake_python_styleguide/visitors/ast/classes.py
```diff
@@ -66,9 +71,16 @@ def _check_base_classes_number(self, node: ast.ClassDef) -> None:
         """Check number of base classes."""
         if len(node.bases) > self.options.max_base_classes:
             self.add_violation(
-                TooManyBaseClassesViolation(node, ext=node.name),
+                TooManyBaseClassesViolation(node, text=node.name),
             )
 
+    def _check_yield_inside_init(self, node: AnyFunctionDef) -> None:
+        """Check yield is not in the __init__."""
+        if node.name == '__init__':
```
sobolevn: We have a constant for `__init__`: https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/constants.py#L161

Please, use it.

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -733,8 +734,40 @@ class ReassigningVariableToItselfViolation(ASTViolation):
...
 
     #: Error message shown to the user.
     error_template = 'Found reassigning variable "{0}" to itself'
     code = 438
+
+
+@final
+class YieldInsideInitViolation(ASTViolation):
+    """
+    Forbids to use ``yield`` inside of ``__init__``
```
sobolevn: Should be `Forbids to use ``yield`` inside of ``__init__`` method.`

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -733,8 +734,40 @@ class ReassigningVariableToItselfViolation(ASTViolation):
...
+
+    Reasoning:
+        ``__init__`` should be used to initialize new objects.
+        It shouldn't create any new objects by yielding them
```
sobolevn: Missing dot in the end.

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -733,8 +734,40 @@ class ReassigningVariableToItselfViolation(ASTViolation):
...
+
+    Example::
+         # Correct:
```
sobolevn: Missing new line between `Example::` and `# Correct`.

sobolevn:
@Shikhov-P I have reused your code to make this feature. I'm quite in a hurry, since I really need to release a new version on this weekend.

Thanks for your work! It became a solid based for this new feature. I will list you as a contributor.
Feel free to take any other issues you want to work with.

Issue #282, opened 2018-10-21
async def offset is broken in ast for versions below 3.7rc2


vsmaxim:
# Bug report
Python with versions below 3.7rc2 are counting offset for async def functions wrong, needs to work around this, so in case of this OffsetVisitor is working completely wrong. 

## What's wrong
Offset for async functions is calculated wrong due to [bug](https://bugs.python.org/issue29205) in python versions <3.7rc2. 
```python
assert_errors = <function assert_errors.<locals>.factory at 0x051D4810>
parse_ast_tree = <function parse_ast_tree.<locals>.factory at 0x051D48E8>
code = "\nasync def container():\n    async with open('some') as temp:\n        pass\n"
number_of_errors = 1
options = <function options.<locals>.factory at 0x05251390>

    def factory(visitor: BaseVisitor, errors: Sequence[str]):
>       assert len(errors) == len(visitor.violations)
E       AssertionError: assert 1 == 2
E        +  where 1 = len([<class 'wemake_python_styleguide.violations.complexity.TooDeepNestingViolation'>])
E        +  and   2 = len([<wemake_python_styleguide.violations.complexity.TooDeepNestingViolation object at 0x058C4B10>, <wemake_python_styleguide.violations.complexity.TooDeepNestingViolation object at 0x058C4690>])
E        +    where [<wemake_python_styleguide.violations.complexity.TooDeepNestingViolation object at 0x058C4B10>, <wemake_python_styleguide.violations.complexity.TooDeepNestingViolation object at 0x058C4690>] = <wemake_python_styleguide.visitors.ast.complexity.offset.OffsetVisitor object at 0x0589CA10>.violations
```
Also, testing it in python shell:
```python
>>> f = ast.parse(nested_asyncwith)
>>> f.body[0].col_offset
6
>>> f.body[0]
<_ast.AsyncFunctionDef object at 0x7fe41dbf0e10>
``` 
where
```python
nested_asyncwith = """
async def container():
    async with open('some') as temp:
        pass
"""
```

## How is that should be
We should fix the async def offset calculations.

vsmaxim:
I will workaround this.

sobolevn:
Closed via https://github.com/wemake-services/wemake-python-styleguide/pull/283

PR #283, opened 2018-10-21
Fix visiting pass node to calc offset

vsmaxim:
# I have made things!

## Checklist

## Related issues

Refs #282.

### DIFF
--- a/wemake_python_styleguide/visitors/ast/complexity/offset.py
+++ b/wemake_python_styleguide/visitors/ast/complexity/offset.py
@@ -25,16 +25,20 @@ from wemake_python_styleguide.visitors.decorators import alias
     'visit_Break',
     'visit_Assign',
     'visit_Expr',
+    'visit_Await',
+    'visit_Pass',
+))
+@alias('visit_async_statement', (
     'visit_AsyncFor',
     'visit_AsyncWith',
-    'visit_Await',
+    'visit_AsyncFunctionDef',
 ))
 class OffsetVisitor(BaseNodeVisitor):
     """Checks offset values for several nodes."""
 
-    def _check_offset(self, node: ast.AST) -> None:
-        offset = getattr(node, 'col_offset', None)
-        if offset is not None and offset > self.options.max_offset_blocks * 4:
+    def _check_offset(self, node: ast.AST, error: int = 0) -> None:
+        offset = getattr(node, 'col_offset', 0) - error
+        if offset > self.options.max_offset_blocks * 4:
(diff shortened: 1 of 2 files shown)


wemake_python_styleguide/visitors/ast/complexity/offset.py
```diff
@@ -55,3 +59,21 @@ def visit_line_expression(self, node: ast.AST) -> None:
...
+        for them isn't calculated properly. We can calculate right version
+        of offset with subscripting 6.
+
+        Read more: https://bugs.python.org/issue29205
+
+        Raises:
+            TooDeepNestingViolation
+
+        """
+        error = 6 if node.col_offset % 4 != 0 else 0
```
vsmaxim: Not sure what to do with this magic number as it is temporary solution.

wemake_python_styleguide/visitors/ast/complexity/offset.py
```diff
@@ -25,16 +25,20 @@
...
     'visit_Assign',
     'visit_Expr',
+    'visit_Await',
+    'visit_Pass',
+))
+@alias('visit_async_statement', (
     'visit_AsyncFor',
     'visit_AsyncWith',
-    'visit_Await',
+    'visit_AsyncFunctionDef',
```
sobolevn: Why this line is required? I am not sure that it actually is.
Because you can not nest functions in other functions. 

That's why I am wondering if we really need `async def` to be checked here.

wemake_python_styleguide/visitors/ast/complexity/offset.py
sobolevn: I like that you are trying to fix the bug, but I do not like the proposed solution.
Here's why:

1. We have this bug in our `tree` object that comes from `flake8`
2. We solve this bug only for this specific visitor
3. Other visitors that might rely on this value will still face this bug

What I suggest instead?
We can create a mechanism for patching nodes that are required to be patch with either business logic or bugfixes. So we will have this bug fixed for all visitors in a single place.

But where we should do it? I guess this should be executed here: https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/checker.py#L176-L177

Before any other checks are executed. We can call this technique `Patchers` and also have a pipeline of patchers that is executed before any checkers. 

And we can conditionally call this specific patcher on the affected `python` versions. And ignore it in the future releases.

What does this sound to you?

vsmaxim: Hm, sounds nice. I'll try to implement this fix soon.

PR #310, opened 2018-10-29
minor grammar fix suggestion

izomeraza:
the most strict --> the strictest


sobolevn:
Ok, I will finish it by myself. Thanks!

PR #315, opened 2018-10-30
Forbids use of `_` anonymous variable

### LINKED ISSUE #309: Limit how `_` variable can be used
# Rule request

## Thesis

Today I saw this code:

```python
[redis_db.delete(symbol_lock_key(_, endpoint_name)) for _ in symbols if True or redis_db.get(symbol_lock_key(_, endpoint_name)) == upid]
```

The thing about it is that we are using `_` as a regular variable.
We also need to list this special case in our `naming.rst`.

## Reasoning

`_` is a special case that indicates that this variable is not used.
So, when it is used we need to alert the user.

nixphix:
Closes #309

Added rule for `_` variable name **Z119**

## Checklist

Adding `_` check raised violation for existing code in the repo, need to fix them.


wemake_python_styleguide/logics/naming/logical.py
```diff
@@ -160,3 +160,35 @@ def does_contain_consecutive_underscores(name: str) -> bool:
...
+    False
+
+    >>> is_anonymous_variable('_')
+    True
+
+    """
+    if name == '_':
+        return True
+
+    return False
```
42B:
```python
return name == '_'
```

wemake_python_styleguide/violations/naming.py
```diff
@@ -519,3 +521,36 @@ class ConsecutiveUnderscoresInNameViolation(MaybeASTViolation):
...
+    .. versionadded:: 0.5.0
+
+    Note:
+        Returns Z119 as error code
+
+    """
+
+    #: Error message shown to the user.
+    error_template = 'Found anonymous variable: {0}'
+    code = 119
```
sobolevn: Why is it `119` and not `117`?

wemake_python_styleguide/visitors/ast/naming.py
```diff
@@ -60,11 +61,17 @@ def _check_name(self, node: ast.AST, name: str) -> None:
 
         if logical.does_contain_underscored_number(name):
             self.add_violation(UnderscoredNumberNameViolation(node, text=name))
+
         if logical.does_contain_consecutive_underscores(name):
             self.add_violation(
                 ConsecutiveUnderscoresInNameViolation(node, text=name),
             )
 
+        if logical.is_anonymous_variable(name):
```
sobolevn: That's not how this should work.
We should forbid usage of the variable name, not its declaration. What do I mean by usage:
- `print(_)`
- `4 * _`
- `x = _`

What do I mean by declaration?
- `some, _ = call()`
- `(do_some() for _ in range(5))`

nixphix: I'm going through `green tree snakes` document to implement this rule. Meanwhile, let me know if there is a similar implementation in the repo.

sobolevn: Nope, I do not think so. You need to get the context of `ast.Name`. It should not be `Load` for this variable.

nixphix:
for some reason `test_all_visitors_contained_in_checker` is not passing, does this require explicit configuration of the new Node Visitor ?

sobolevn:
@nixphix add your new visitor to the preset.

sobolevn:
@nixphix thanks a lot! Awesome work!

P.S. Do you have any feedback? How was your developer experience? Docs quality? Review process? How did you like the codebase quality and style? Anything, generally!

nixphix:
> @nixphix thanks a lot! Awesome work!
> 
> P.S. Do you have any feedback? How was your developer experience? Docs quality? Review process? How did you like the codebase quality and style? Anything, generally!

It was great, got to know about AST better.  Using `Poetry` for the first time.

Issue #267, opened 2018-10-16
Restrict unicode names


sobolevn:
# Rule request

## Thesis

Currently we can do something like this:

```python
никита = 'sobolev'
```

We can use unicode to create variables!
We can even do bizarre things like `Ø: Set[str] = set()`.

## Reasoning

This should be forbidden for sanity and readability.

riyasyash:
I can work on this :)

sobolevn:
@riyasyash awesome!

PR #319, opened 2018-11-05
Restrict unicode names #267

geoc0ld:
# I have made things!

## Checklist


wemake_python_styleguide/visitors/ast/naming.py
```diff
@@ -33,6 +35,8 @@
 class _NameValidator(object):
     """Utility class to separate logic from the visitor."""
 
+    _unicode_pattern: ClassVar[Pattern] = re.compile(r'([a-zA-Z_0-9]+)')
```
sobolevn: Why have you chosen this way instead of https://stackoverflow.com/questions/35889505/check-that-a-string-contains-only-ascii-characters/35890514 ?

wemake_python_styleguide/violations/naming.py
```diff
@@ -510,3 +512,28 @@ def __init__(self):
...
+     Restrict unicode names.
+
+    Reasoning:
+        This should be forbidden for sanity and readability.
+
+    Solution:
+        Rename so that they contains only latin letters.
+
+    Example::
+        # Correct:
```
sobolevn: This should be:

```
Example::

         # Correct:
         some_variable = 42

         # Wrong:
         переменная = 42
```

sobolevn: Mind new lines and whitespaces.

tests/test_visitors/test_ast/test_naming/test_unicode_variable.py
```diff
@@ -0,0 +1,62 @@
...
+    assert_errors(visitor, [UnicodeNameViolation])
+
+
+corrrect_fragmet_all = """
+some_variable = 42
+test_variable = some_variable
+"""
+
+
+@pytest.mark.parametrize('code', [
```
sobolevn: You have to refactor your tests. We now use different approach to testing. 
See https://github.com/wemake-services/wemake-python-styleguide/blob/master/tests/test_visitors/test_ast/test_naming/test_naming_rules/test_wrong_names.py

wemake_python_styleguide/visitors/ast/naming.py
geoc0ld: The implementation for tests a "class name" was not changed, because the new approach does not work with class name. Remove "_check_name" from "visit_classDef"?

sobolevn:
@geoc0ld thanks! I will take this PR from here. I need to release a new version today.

PR #339, opened 2018-11-18
Forbid multiple assignment targets for context managers #328

### LINKED ISSUE #328: Forbid multiple assignment targets for context managers
# Rule request

## Thesis

Code example:

```python
with open('') as first, second:
    ...
```

Quote from `pylint`:

```
confusing-with-statement (W0124):
--
  | Following "as" with another context manager looks like a tuple. Emitted when a with statement component returns multiple values and uses name binding with as only for a part of those values, as in with ctx() as a, b. This can be misleading, since it's not clear if the context manager returns a tuple or if the node without a name binding is another context manager.
```

geoc0ld:
# I have made things!

## Checklist

## Related issues

Closes #328

### DIFF
--- a/tests/test_checker/test_noqa.py
+++ b/tests/test_checker/test_noqa.py
@@ -61,6 +61,7 @@ def test_noqa_fixture_disabled(absolute_path, all_violations):
         'Z222': 1,
         'Z223': 1,
         'Z224': 1,
+        'Z225': 0,
 
         'Z300': 1,
         'Z301': 1,
--- a/tests/test_checker/test_noqa.py
+++ b/tests/test_checker/test_noqa.py
@@ -61,7 +61,6 @@ def test_noqa_fixture_disabled(absolute_path, all_violations):
         'Z222': 1,
         'Z223': 1,
         'Z224': 1,
-        'Z225': 0,
 
         'Z300': 1,
         'Z301': 1,
@@ -79,6 +78,7 @@ def test_noqa_fixture_disabled(absolute_path, all_violations):
         'Z313': 1,
         'Z314': 1,
         'Z315': 1,
+        'Z316': 0,
 
         'Z400': 0,
         'Z401': 0,
--- a/tests/test_visitors/test_ast/test_complexity/test_counts/test_context_manager_targets.py
+++ b/tests/test_visitors/test_ast/test_complexity/test_counts/test_context_manager_targets.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+
+import pytest
+
+from wemake_python_styleguide.violations.complexity import (
+    TooManyContextManagerAssignmentsViolation,
+)
+from wemake_python_styleguide.visitors.ast.complexity.counts import (
+    ContextManagerVisitor,
+)
+
+incorrect_count = """
+with open('') as first, second:
+    ...
+"""
+
+
+@pytest.mark.parametrize('code', [
+    incorrect_count,
+])
+def test_context_manager_multiple_targets(
+    assert_errors,
+    parse_ast_tree,
+    code,
+    default_options,
+):
+    """Testing incorrect count context manager assignment."""
+    tree = parse_ast_tree(code)
+
+    visitor = ContextManagerVisitor(default_options, tree=tree)
+    visitor.run()
+
+    assert_errors(visitor, [TooManyContextManagerAssignmentsViolation])
+
+
+correct_count = """
+with open('') as first:
+    ...
+"""
+
+
+@pytest.mark.parametrize('code', [
+    correct_count,
+])
+def test_context_manager_alone_target(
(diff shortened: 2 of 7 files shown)


tests/test_checker/test_noqa.py
```diff
@@ -61,6 +61,7 @@ def test_noqa_fixture_disabled(absolute_path, all_violations):
         'Z222': 1,
         'Z223': 1,
         'Z224': 1,
+        'Z225': 0,
```
sobolevn: It is not really a complexity violation. I would say that this is a consistency issue.

tests/test_visitors/test_ast/test_complexity/test_counts/test_context_manager_targets.py
```diff
@@ -0,0 +1,57 @@
...
+    """Testing incorrect count context manager assignment."""
+    tree = parse_ast_tree(code)
+
+    visitor = ContextManagerVisitor(default_options, tree=tree)
+    visitor.run()
+
+    assert_errors(visitor, [TooManyContextManagerAssignmentsViolation])
+
+
+correct_count = """
```
sobolevn: can you please move this example up? so it will be near `incorrect_count`

tests/test_visitors/test_ast/test_complexity/test_counts/test_context_manager_targets.py
```diff
@@ -0,0 +1,57 @@
...
+    ContextManagerVisitor,
+)
+
+incorrect_count = """
+with open('') as first, second:
+    ...
+"""
+
+
+@pytest.mark.parametrize('code', [
```
sobolevn: There's no need for `@pytest.mark.parametrize`, since there's only one case.

tests/test_visitors/test_ast/test_complexity/test_counts/test_context_manager_targets.py
```diff
@@ -0,0 +1,57 @@
...
+    assert_errors(visitor, [TooManyContextManagerAssignmentsViolation])
+
+
+correct_count = """
+with open('') as first:
+    ...
+"""
+
+
+@pytest.mark.parametrize('code', [
```
sobolevn: There's no need for `@pytest.mark.parametrize`, since there's only one case.

wemake_python_styleguide/violations/complexity.py
```diff
@@ -574,3 +576,31 @@ class TooManyForsInComprehensionViolation(ASTViolation):
...
+        Emitted when a with statement component returns multiple values
+        and uses name binding with as only for a part of those values,
+        as in with ctx() as a, b.This can be misleading, since it's not
+        clear if the context manager returns a tupleor if the node
+        without a name binding is another context manager.
+
+    Example::
+
+        # Wrong:
+            with open('') as first, second:
```
sobolevn: Please, fix the indentation. `# Wrong:` and `with open('')` should have the same indentation.

wemake_python_styleguide/violations/complexity.py
```diff
@@ -574,3 +576,31 @@ class TooManyForsInComprehensionViolation(ASTViolation):
...
+                ...
+
+    .. versionadded:: 0.5.1
```
sobolevn: It should be `0.6.0`

tests/test_visitors/test_ast/test_complexity/test_counts/test_context_manager_targets.py
```diff
@@ -0,0 +1,57 @@
...
+    TooManyContextManagerAssignmentsViolation,
+)
+from wemake_python_styleguide.visitors.ast.complexity.counts import (
+    ContextManagerVisitor,
+)
+
+incorrect_count = """
+with open('') as first, second:
+    ...
+"""
```
sobolevn: We also need to test this case:

```python
with open('first') as first, open('second') as second:
    ...
```

sobolevn: It should be allowed.

sobolevn: @geoc0ld on the other hand, no. This should not be allowed. This is very hard to read.
Please, write a test that this sample raises a violation.

sobolevn: @geoc0ld on the other hand, no. This should not be allowed. This is very hard to read.
Please, write a test that this sample raises a violation.

geoc0ld:
@sobolevn i added a line into CHANGELOG, also added the "0.6.0" section. It is correct, or remove this changes?

sobolevn:
@geoc0ld everything is fine. Thank you!

Issue #506, opened 2019-02-27
Forbid multiline conditions


sobolevn:
# Rule request

## Thesis

We need to forbid code like:

```python
       if isinstance(node.test, ast.UnaryOp) and isinstance(
            node.test.op,
            ast.Not,
        ):
```

It is really hard to read it. And conditions start to be "blurry".
The preferred way to write such conditions is:

```python
if isinstance(node.test, ast.UnaryOp:
    if isinstance(node.test.op, ast.Not):
```

## Reasoning

This way of writing conditions hides the inner complexity this line has.
And it decreases readability of the code.

PR #524, opened 2019-03-23
Issue #506

shok9894:
Forbid multiline conditions


tests/test_checker/test_noqa.py
```diff
@@ -151,6 +151,7 @@ def test_noqa_fixture_disabled(absolute_path, all_violations):
         'Z462': 1,
         'Z463': 1,
         'Z464': 1,
+        'Z465': 1,
```
sobolevn: Should be `2`

Issue #528, opened 2019-03-28
False positive Z309 for keys


sobolevn:
# Bug report

## What's wrong

```python
def test(office, returned_item):
    assert returned_item['id'] == office.id
```

raises `Z309 Found reversed comparison order` violation

## How is that should be

It should not raise a violation. We use variables on both sides.

## System information

I am using `0.8.1` release.

PR #532, opened 2019-03-29
False positive Z309 for keys #528

### LINKED ISSUE #533: False positive for Z121 when using `_` for unused var names
# Bug report

## What's wrong

This code triggers `Z121`:

```python
def logo_and_user():
    user, _ = User.objects.get_or_create(username='sobolevn')
    logo, _ = Logo.objects.get_or_create(username='sobolevn')  # here
```

Log: `test.py:4:11: Z121 Found usage of a variable marked as unused: _`
But it should not, since these variables are unused.

## How is that should be

It should not raise any `Z121` violations here.

## System information

I am using `8.0.1`

SizovIgor:
Closes #533


CHANGELOG.md
```diff
@@ -3,6 +3,19 @@
 We follow Semantic Versions since the `0.1.0` release.
 We used to have incremental versioning before `0.1.0`.
 
+## 0.9.1
```
sobolevn: This should be just one `0.9.0` release

sobolevn:
@SizovIgor Thanks a lot! Bug is fixed. 🔥

Issue #471, opened 2019-01-31
Forbid local variables that are only used in `return` statements


sobolevn:
# Rule request

## Thesis

Sometimes I see code like this:

```python
def function():
     ...
     some_value = ...
     return some_value
```

What the point in creating this intermediate variable? Why do we need it?

## Reasoning

This is done for consistency and more readable source code.

loucadufault:
Can I start working on this ? To clarify the fix would be to remove all these intermediate variables in all function returns?

sobolevn:
@loucadufault you are welcome! 

The fix would be to return a direct value. 

Wrong:

```python
def some_function():
     some_value = 1
     return some_value
```

Correct:

```python
def some_function():
     return 1
```

Chinmaygoyal:
Can I work on this issue?

SizovIgor:
@Chinmaygoyal I am on finish line to complete this issue

sobolevn:
@Chinmaygoyal feel free to take #545

DmytroLitvinov:
Hello @sobolevn 
The immediate value is good in the next case:
```python
def function(a):
     result = None
     if a:
         ...  # some actions
        result = 123

     return result
```
As for me, it is also understandable what from naming of value what we will return after calling of function.

sobolevn:
@DmytroLitvinov yes, this code is valid.

DmytroLitvinov:
https://github.com/sbdchd/flake8-pie

PR #544, opened 2019-04-09
Issue #471

SizovIgor:


tests/test_visitors/test_ast/test_keywords/test_consistency_returning/test_consistency_return_variables.py
```diff
@@ -0,0 +1,81 @@
...
+
+correct_example1 = """
+def some_function():
+    return 1
+"""
+
+correct_example2 = """
+def some_function():
+     some_value = 1
+     name = last_name + first_name
```
sobolevn: This is not correct. `some_value` is not used.

sobolevn: Move it wrong examples.

tests/test_visitors/test_ast/test_keywords/test_consistency_returning/test_consistency_return_variables.py
```diff
@@ -0,0 +1,81 @@
...
+
+wrong_example1 = """
+def function():
+     some_value = 1
+     return some_value
+"""
+
+
+@pytest.mark.parametrize('code', [
+    wrong_example1,
```
sobolevn: Add one more example with multiple return branches. Two wrong `return`s == two errors.

wemake_python_styleguide/visitors/ast/keywords.py
```diff
@@ -73,6 +74,25 @@ def _check_last_return_in_function(self, node: ast.Return) -> None:
         if node is parent.body[-1] and node.value is None:
             self.add_violation(InconsistentReturnViolation(node))
 
+    def _check_variables_used_only_in_return(
+        self,
+        node: AnyFunctionDef,
+    ) -> None:
+        variable_name = ''
+        return_variable_id = ''
+        if len(node.body) > 1:
```
sobolevn: After writing new tests you will find bugs in this code. It should be rewritten.

PR #575, opened 2019-06-07
Fixing docstring copy paste errors

AlexTereshenkov:
Some classes have been copy pasted and then edited, but the headline remained the same. They have been updated to reflect the rule of each particular class.


sobolevn:
Thanks!

PR #617, opened 2019-06-28
Closes #601

### LINKED ISSUE #601: Do not allow to yield too long tuples
# Rule request

## Thesis

Expressions like `yield foo, bar, baz, fling, flung` should not be allowed.

## Reasoning

We should use lists of similar type or wrapper objects instead.
Related: https://github.com/troessner/reek/blob/master/docs/Long-Yield-List.md
Original: https://github.com/wemake-services/wemake-python-styleguide/issues/422

fullset:
# I have made things!

## Checklist

## Related issues

Hi.
@sobolevn Look, please!

### DIFF
--- a/tests/fixtures/noqa.py
+++ b/tests/fixtures/noqa.py
@@ -27,6 +27,20 @@ def __getattr__():  # noqa: Z413
     anti_z444 = 1
 
 
+def foo_func(limit=10):
+    # See:
+    # https://github.com/wemake-services/wemake-python-styleguide/issues/601
+    counter = 0
+    c1_in = 0
+    c2_in = 0
+    c1_out = 0
+    c2_out = 0
+
+    while counter < limit:
+        yield c1_in, c2_in, limit, c1_out, c2_out, counter  # noqa: Z227
+        counter += 1
+
+
 def function_name(
     value: int = 0,  # noqa: Z110
 ):
--- a/tests/fixtures/noqa.py
+++ b/tests/fixtures/noqa.py
@@ -27,18 +27,11 @@ def __getattr__():  # noqa: Z413
     anti_z444 = 1
 
 
-def foo_func(limit=10):
+def foo_func():
     # See:
     # https://github.com/wemake-services/wemake-python-styleguide/issues/601
-    counter = 0
-    c1_in = 0
-    c2_in = 0
-    c1_out = 0
-    c2_out = 0
 
-    while counter < limit:
-        yield c1_in, c2_in, limit, c1_out, c2_out, counter  # noqa: Z227
-        counter += 1
+    yield (1, 2, 3, 4, 5, 6)  # noqa: Z227
 
 
 def function_name(
--- a/tests/test_visitors/test_ast/test_complexity/test_counts/test_yield_length.py
+++ b/tests/test_visitors/test_ast/test_complexity/test_counts/test_yield_length.py
@@ -1,0 +1,60 @@
+# -*- coding: utf-8 -*-
+
+import pytest
+
+from wemake_python_styleguide.visitors.ast.complexity.counts import (
+    YieldTupleVisitor,
+    TooLongYieldTupleViolation,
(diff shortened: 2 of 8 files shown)


tests/test_visitors/test_ast/test_complexity/test_counts/test_yield_length.py
```diff
@@ -0,0 +1,67 @@
...
+    i = 0
+    while True:
+        yield i + 1, i + 2, i + 3
+"""
+
+long_yield = """
+def function_name(foo, bar, baz):
+    i = 0
+    while True:
+        yield i + 1, i + 2, i + 3, foo + 1, bar + 1, baz + 1
```
sobolevn: Let's do a template instead! And insert values into it.
Because `short_yield` and `long_yield` are already equal.

We should also test: `()` (empty tuple), `(1, 2, 3, 4, 5, 6)` (tuple with `(` and `)`)

tests/test_visitors/test_ast/test_complexity/test_counts/test_yield_length.py
```diff
@@ -0,0 +1,67 @@
...
+    code,
+    default_options,
+):
+    """Testing that violations are raised when reaching max value."""
+    tree = parse_ast_tree(code)
+
+    visitor = YieldTupleVisitor(default_options, tree=tree)
+    visitor.run()
+
+    assert_errors(visitor, [TooLongYieldTupleViolation])
```
sobolevn: `assert_error_text` is missing.

wemake_python_styleguide/visitors/ast/complexity/counts.py
```diff
@@ -253,3 +255,36 @@ def visit_Try(self, node: ast.Try) -> None:
...
+@alias('visit_any_function', (
+    'visit_FunctionDef',
+    'visit_AsyncFunctionDef',
+))
+class YieldTupleVisitor(BaseNodeVisitor):
+    """Finds too long ``tuples`` in ``yield`` expressions."""
+
+    def _check_yield_values(self, node: AnyFunctionDef) -> None:
+        for sub_node in ast.walk(node):
+            if isinstance(sub_node, ast.Yield) and sub_node.value:
```
sobolevn: We can directly visit `ast.Yield`, we don't need to visit functions.

fullset:
No problem. I'm going to fix these, but a bit later.

tests/fixtures/noqa.py
```diff
@@ -27,6 +27,20 @@ def __getattr__():  # noqa: Z413
...
+    # https://github.com/wemake-services/wemake-python-styleguide/issues/601
+    counter = 0
+    c1_in = 0
+    c2_in = 0
+    c1_out = 0
+    c2_out = 0
+
+    while counter < limit:
+        yield c1_in, c2_in, limit, c1_out, c2_out, counter  # noqa: Z227
+        counter += 1
```
sobolevn: Do we need this line?

tests/fixtures/noqa.py
```diff
@@ -27,6 +27,20 @@ def __getattr__():  # noqa: Z413
...
+    # See:
+    # https://github.com/wemake-services/wemake-python-styleguide/issues/601
+    counter = 0
+    c1_in = 0
+    c2_in = 0
+    c1_out = 0
+    c2_out = 0
+
+    while counter < limit:
+        yield c1_in, c2_in, limit, c1_out, c2_out, counter  # noqa: Z227
```
sobolevn: Do we need all these variables? Can we `yield` just number? 
It would be shorter but have the same effect.

fullset:
@sobolevn done.

tests/fixtures/noqa.py
fullset: Ok. The simpler the better.

sobolevn:
Awesome job! Thank you.

PR #621, opened 2019-07-01
Issue 616

### LINKED ISSUE #616: Incorrect line number for Z331
# Bug report

## What's wrong

```python
def foreign_jwt_authentication(token: bytes):  # noqa: Z331
    """
    Goes to the remote provider to validate the token.

    Returns the payload if everything is fine.
    Otherwise raises an exception.
    """
    unverified_payload = jwt.decode(  # type: ignore
        token, key=None, verify=False,
    )
    _make_authentication_request(settings.TOKEN_VERIFICATION_URL, token)
    return unverified_payload
```

Error:

```
./server/rest_app/logics/auth.py:41:1: Z331 Found local variable that are only used in `return` statements
```

## How is that should be

Should be reported on the ` return unverified_payload` line.

fullset:
# I have made things!

## Checklist

## Related issues

Fixes #616

### DIFF
--- a/wemake_python_styleguide/visitors/ast/keywords.py
+++ b/wemake_python_styleguide/visitors/ast/keywords.py
@@ -2,7 +2,7 @@
 
 import ast
 from collections import defaultdict
-from typing import ClassVar, Dict, List, Type, Union
+from typing import ClassVar, Dict, List, Tuple, Type, Union
 
 from typing_extensions import final
 
@@ -26,6 +26,10 @@ from wemake_python_styleguide.visitors.base import BaseNodeVisitor
 from wemake_python_styleguide.visitors.decorators import alias
 
 AnyWith = Union[ast.With, ast.AsyncWith]
+NamesAndReturns = Tuple[
+    Dict[str, List[ast.Name]],
+    Dict[str, ast.Return],
+]
 ReturningViolations = Union[
     Type[InconsistentReturnViolation],
     Type[InconsistentYieldViolation],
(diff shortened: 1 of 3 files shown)


wemake_python_styleguide/visitors/ast/keywords.py
```diff
@@ -261,6 +261,18 @@ def _get_return_node_variables(
                     returns[variable_name].append(sub_node.value)
         return returns
 
+    def _get_return_nodes(
```
sobolevn: These two pieces of code are almost identical:

```py
def _get_return_node_variables(
        self,
        node: List[ast.AST],
    ) -> Dict[str, List[ast.Name]]:
        returns: Dict[str, List[ast.Name]] = defaultdict(list)
        for sub_node in node:
            if isinstance(sub_node, ast.Return):
                if isinstance(sub_node.value, ast.Name):
                    variable_name = sub_node.value.id
                    returns[variable_name].append(sub_node.value)
        return returns

    def _get_return_nodes(
        self,
        node: List[ast.AST],
    ) -> Dict[str, ast.Return]:
        return_sub_nodes: Dict[str, ast.Return] = defaultdict()
        for sub_node in node:
            if isinstance(sub_node, ast.Return):
                if isinstance(sub_node.value, ast.Name):
                    variable_name = sub_node.value.id
                    return_sub_nodes[variable_name] = sub_node
        return return_sub_nodes
```

Let's improve it!

sobolevn: We can return tuples of `(return_node, variable_names)`. Any other solution is also fine.

fullset: I tried this way. It works fine. The problem is that annotation like this:
```
def _get_return_node_variables(
    self,
    node: List[ast.AST],
) -> Tuple[Dict[str,  List[ast.Name]], Dict[str, ast.Return]]:
```
leads to violations:
```
./wemake_python_styleguide/visitors/ast/keywords.py:255:10: TAE002 too complex annotation (4 > 3)
./wemake_python_styleguide/visitors/ast/keywords.py:255:10: Z221 Found line with high Jones Complexity: 17
```

So, don't want to rewrite the whole visitor. I'll think more about this problem and will try to solve it.

sobolevn: Just create a type alias somewhere at the top: `NamesAndReturns = Tuple[Dict[str,  List[ast.Name]], Dict[str, ast.Return]]`

fullset: Oh, saw this many times and didn't guess to use..

sobolevn:
🍕 🍺

PR #641, opened 2019-07-10
Adds flake8-executable as a dependency.

### LINKED ISSUE #608: Add flake8-executable
https://github.com/xuhdev/flake8-executable

See how plugins should be added here: https://github.com/wemake-services/wemake-python-styleguide/blob/master/CONTRIBUTING.md#adding-new-flake8-plugins

proofit404:
Closes #608

### DIFF
--- a/tests/fixtures/external_plugins.py
+++ b/tests/fixtures/external_plugins.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env perl
 from sys import *
 import sys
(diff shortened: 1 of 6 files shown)


tests/fixtures/external_plugins.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env perl
```
sobolevn: Nice one 🙂

Issue #658, opened 2019-07-12
Create `TooManyImportedNamesViolation`


sobolevn:
# Rule request

## Thesis

We can write: `from a import b`
Or you can write:

```python
from a import b, c, d, e, f, g, k, l, m, n, o
```

We need to count individual imported names.
Where `import a` counts as 1 imported name.
And `from a import b, c` counts as 2

## Reasoning

Injecting too many names inside your context is bad.

sobolevn:
I am not sure about the default value. 
This rule 100% should be configurable. But, the value should be guessed.

PR #665, opened 2019-07-14
Closes #662

### LINKED ISSUE #662: Forbid to use `raise StopIteration()` inside generators
# Rule request

## Thesis

We need to check async and regular functions that they contain `yield`, `yield x` or `yield from` to count them as generators.

Next, we need to find all `raise` nodes and check:
- if the context is a generator
- if the exception is `StopIteration` or `StopIteration()` or `StopIteration(some)`
- raise a violation if so

## Reasoning

One should use `return value` instead.

bellum:
# I have made things!

## Checklist

## Related issues
Closes #662 

## Changes out of scope of the related issue
- I have refactored processing `ast.Raise` classes to get exception names as new function `wemake_python_styleguide.logic.nodes.get_exception_name` and except using it inside my new code I have updated it inside `wemake_python_styleguide.visitors.ast.keywords.WrongRaiseVisitor`. Let me know in case this should be inside separate commit.
- I have created new function `wemake_python_styleguide.logic.nodes.get_subnodes_by_type` to get subnodes of given node by given ast type.

### DIFF
--- a/tests/test_visitors/test_ast/test_functions/test_stop_iteration.py
+++ b/tests/test_visitors/test_ast/test_functions/test_stop_iteration.py
@@ -9,50 +9,44 @@ from wemake_python_styleguide.visitors.ast.functions import (
     FunctionDefinitionVisitor,
 )
 
-stop_iteration_generator_method = """
+stop_iteration_method = """
 class CheckStopIteration():
     def check_stop_iteration(self):
-        yield
-        raise {0}
+        {0}
+        raise {1}
 """
 
-stop_iteration_generator_function = """
+stop_iteration_function = """
 def check_stop_iteration():
-    yield
-    raise {0}
-"""
-
-
-stop_iteration_bare_method = """
-class CheckStopIteration():
-    def check_stop_iteration(self):
-        raise {0}
-"""
-
-stop_iteration_bare_function = """
-def check_stop_iteration():
-    raise {0}
+    {0}
+    raise {1}
 """
 
 
 @pytest.mark.parametrize('code', [
-    stop_iteration_generator_method,
-    stop_iteration_generator_function,
+    stop_iteration_method,
+    stop_iteration_function,
+])
+@pytest.mark.parametrize('statement', [
+    'yield',
+    'yield some_parameter',
 ])
 @pytest.mark.parametrize('exception', [
     'StopIteration',
     'StopIteration()',
+    'StopIteration(1)',
 ])
 def test_stop_iteration_inside_generators(
     assert_errors,
     parse_ast_tree,
     code,
+    statement,
     exception,
     default_options,
     mode,
 ):
     """Testing that `raise StopIteration` is restricted inside generators."""
-    tree = parse_ast_tree(mode(code.format(exception)))
+    tree = parse_ast_tree(mode(code.format(statement, exception)))
 
     visitor = FunctionDefinitionVisitor(default_options, tree=tree)
     visitor.run()
@@ -61,23 +55,57 @@ def test_stop_iteration_inside_generators(
 
 
 @pytest.mark.parametrize('code', [
-    stop_iteration_bare_method,
-    stop_iteration_bare_function,
+    stop_iteration_method,
+    stop_iteration_function,
+])
+@pytest.mark.parametrize('statement', [
+    'yield from generator()',
 ])
 @pytest.mark.parametrize('exception', [
     'StopIteration',
     'StopIteration()',
+    'StopIteration(1)',
+])
+def test_stop_iteration_in_generators_yield_from(
+    assert_errors,
+    parse_ast_tree,
+    code,
+    statement,
+    exception,
+    default_options,
+):
+    """Testing that `raise StopIteration` is restricted inside generators."""
+    tree = parse_ast_tree(code.format(statement, exception))
+
+    visitor = FunctionDefinitionVisitor(default_options, tree=tree)
+    visitor.run()
+
+    assert_errors(visitor, [StopIterationInsideGeneratorViolation])
+
+
+@pytest.mark.parametrize('code', [
+    stop_iteration_method,
+    stop_iteration_function,
+])
+@pytest.mark.parametrize('statement', [
+    'print("not a generator")',
+])
+@pytest.mark.parametrize('exception', [
+    'StopIteration',
+    'StopIteration()',
+    'StopIteration(1)',
 ])
 def test_stop_iteration_inside_bare_functions(
     assert_errors,
     parse_ast_tree,
     code,
+    statement,
     exception,
     default_options,
     mode,
 ):
     """Testing that `raise StopIteration` is allowed inside bare functions."""
-    tree = parse_ast_tree(mode(code.format(exception)))
+    tree = parse_ast_tree(mode(code.format(statement, exception)))
 
     visitor = FunctionDefinitionVisitor(default_options, tree=tree)
     visitor.run()
@@ -86,23 +114,29 @@ def test_stop_iteration_inside_bare_functions(
 
 
 @pytest.mark.parametrize('code', [
-    stop_iteration_generator_method,
-    stop_iteration_generator_function,
+    stop_iteration_method,
+    stop_iteration_function,
+])
+@pytest.mark.parametrize('statement', [
+    'yield',
+    'yield some_parameter',
 ])
 @pytest.mark.parametrize('exception', [
     'RuntimeError',
     'RuntimeError()',
+    'RuntimeError(1)',
 ])
 def test_other_exceptions_inside_generators(
     assert_errors,
     parse_ast_tree,
     code,
+    statement,
     exception,
     default_options,
     mode,
 ):
     """Testing that `raise` of other exceptions is allowed inside generators."""
-    tree = parse_ast_tree(mode(code.format(exception)))
+    tree = parse_ast_tree(mode(code.format(statement, exception)))
 
     visitor = FunctionDefinitionVisitor(default_options, tree=tree)
     visitor.run()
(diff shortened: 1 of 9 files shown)


tests/test_visitors/test_ast/test_functions/test_stop_iteration.py
```diff
@@ -0,0 +1,129 @@
...
+    StopIterationInsideGeneratorViolation,
+)
+from wemake_python_styleguide.visitors.ast.functions import (
+    FunctionDefinitionVisitor,
+)
+
+stop_iteration_generator_method = """
+class CheckStopIteration():
+    def check_stop_iteration(self):
+        yield
```
sobolevn: Can you please change the template generation here. I see this process as:

```python
stop_iteration_generator_function = """
def check_stop_iteration():
    {0}
    raise {1}
"""
```

Things that can be inserted as `{0}`:
- `yield`
- `yield some_parameter`
- `yield from generator()`
- `print('not a generator')` - in this case we do not raise a violation at all

Moreover, we have `mode` fixture to convert `def -> async def`, so there's no need to have explicit `async def` functions. Only functions and methods.
See an example: https://github.com/wemake-services/wemake-python-styleguide/blob/master/tests/test_visitors/test_ast/test_functions/test_complex_default_values.py#L32

 

```

File: tests/test_visitors/test_ast/test_functions/test_stop_iteration.py [hunk 2]
```diff
@@ -0,0 +1,129 @@
...
+"""
+
+
+@pytest.mark.parametrize('code', [
+    stop_iteration_generator_method,
+    stop_iteration_generator_function,
+])
+@pytest.mark.parametrize('exception', [
+    'StopIteration',
+    'StopIteration()',
```
sobolevn: Please, also test `StopIteration(1)`

tests/test_visitors/test_ast/test_functions/test_stop_iteration.py
```diff
@@ -0,0 +1,129 @@
...
+    assert_errors(visitor, [])
+
+
+@pytest.mark.parametrize('code', [
+    stop_iteration_generator_method,
+    stop_iteration_generator_function,
+])
+@pytest.mark.parametrize('exception', [
+    'RuntimeError',
+    'RuntimeError()',
```
sobolevn: Please, also test `RuntimeError(1)`

wemake_python_styleguide/logic/nodes.py
```diff
@@ -49,3 +49,29 @@ def get_parent(node: ast.AST) -> Optional[ast.AST]:
 def get_context(node: ast.AST) -> Optional[ast.AST]:
     """Returns the context or ``None`` if node has no context."""
     return getattr(node, 'wps_context', None)
+
+
+SUBNODE_TYPE = TypeVar('SUBNODE_TYPE')
```
sobolevn: We use `PascalCase` for types and generics.

wemake_python_styleguide/logic/nodes.py
```diff
@@ -49,3 +49,29 @@ def get_parent(node: ast.AST) -> Optional[ast.AST]:
...
+
+
+def get_subnodes_by_type(
+    node: ast.AST,
+    subnodes_type: Type[SUBNODE_TYPE],
```
sobolevn: Should be: `Union[Type[SUBNODE_TYPE], AnyNodes]`
See: https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/types.py#L61

wemake_python_styleguide/visitors/ast/functions.py
```diff
@@ -186,17 +192,40 @@ def _check_argument_default_values(self, node: AnyFunctionDef) -> None:
             if not isinstance(real_arg, self._allowed_default_value_types):
                 self.add_violation(ComplexDefaultValueViolation(node))
 
+    def _check_generator(self, node: AnyFunctionDef) -> None:
```
sobolevn: Let's move the generator detection logic to `logic/`, it might be reused soon.
And call it as `is_generator(node: AnyFunctionDef) -> bool` function.

bellum:
@sobolevn Thanks for your feedback. Gonna resolve them today. How would it better to do?
1. to make new commit with review fixes and squash two commits after your approval
2. to ammend this one commit with review fixes

wemake_python_styleguide/logic/nodes.py
bellum: In case I specify it like this then Iterator should have got `Union[Type[SUBNODE_TYPE], Type[ast.AST]]` type too which leads to inability to use this generic type on side of invocation of the method. That's why I have decided to restrict it to one type because in my case I wanted to get subnodes with type `ast.Raise` and supply them to function which receives `ast.Raise`. After this change mypy claims that `wemake_python_styleguide.logic.nodes.get_exception_name` expects `ast.Raise` but I supply `Union[Type[SUBNODE_TYPE], Type[ast.AST]]`. Am I doing something wrong?

Declaration:
```python
def get_subnodes_by_type(
    node: ast.AST,
    subnodes_type: Union[Type[SUBNODE_TYPE], AnyNodes],
) -> Iterator[Union[Type[SUBNODE_TYPE], Type[ast.AST]]]:
    """Returns the list of subnodes of given node with given subnode type."""
```

tests/test_visitors/test_ast/test_functions/test_stop_iteration.py
bellum: There is one problem here with sync and async generators as async generators do not support `yield from`.

bellum: So it's required to test `yield from` separately for sync generators only without using `mode` fixture. Would it be fine?

sobolevn:
Just commit and push, github can squash commit via UI automatically when PR is merged.

sobolevn:
@bellum LGTM. Let's consider this one as done, but I will merge it after `0.11` is released.
I am going to mess up the existing violation code and categories. Since there are already quite a lot of them in the `best_practices.py`

sobolevn:
I hope I will release `0.11` later this week.

bellum:
@sobolevn Ok, great. No problem. Thanks. Do I understand right that it's better not to rush at fixing other rule related issues due to your intention to change a lot there? If yes then could you direct me which other issues would be better to consider?

wemake_python_styleguide/logic/nodes.py
sobolevn: Got it 👍

sobolevn:
There are three options:
1. Push new features, and I will merge them after `0.11`: no manual fixes from your side will be required once PR is approved
2. Solve bugs with the dependencies: https://github.com/wemake-services/wemake-python-styleguide/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+label%3Adependencies+label%3Abug
3. Create a codacy/codeclimate/jupyter integration: https://github.com/wemake-services/wemake-python-styleguide/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+label%3Adependencies+label%3Afeature

What suits you best?

bellum:
@sobolevn I would choose first option to work on new features. Would it be fine to take this one #658?

By the way I have got the idea of one more issue. The thing which I thought about during working on this issue is that it's required to forbid `StopAsyncIteration` raising inside async generators too.

sobolevn:
@bellum sounds great! 

Considering `StopAsyncIteration`, I am not using `asyncio`, so I am not familiar with it.
I have found https://www.python.org/dev/peps/pep-0492/

What do you suggest to do instead of `raise StopAsyncIteration()`?

bellum:
@sobolevn It should not be used explicitly inside async generators too.

```python
import asyncio

async def g():
    yield
    raise StopAsyncIteration

async def _async_main():
    async for i in g():
        print(i)


def _main():
    loop = asyncio.get_event_loop()
    try:
        loop.run_until_complete(_async_main())
    finally:
        loop.close()


if __name__ == '__main__':
    _main()
```

The snippet above results with next error. So `StopAsyncIteration` is wrapped `RuntimeError` the same as `StopIteration` exception. Instead it's required to use return because for async iterators `StopAsyncIteration` is the same as `StopIteration` for sync iterators whereas `StopIteration` for async iterators is the same as `await statement` because this is how it was evolved (starting from `yield`). 

```
None
  File "./scripts/async.py", line 9, in g
    raise StopAsyncIteration
StopAsyncIteration

The above exception was the direct cause of the following exception:

  File "./scripts/async.py", line 26, in <module>
    _main()
  File "./scripts/async.py", line 20, in _main
    loop.run_until_complete(_async_main())
  File "/usr/lib/python3.6/asyncio/base_events.py", line 473, in run_until_complete
    return future.result()
  File "./scripts/async.py", line 13, in _async_main
    async for i in g():
RuntimeError: async generator raised StopAsyncIteration
```

sobolevn:
Wow, 

```python
async def g():
    yield
    return None
```

Raises:

```
» python ./scripts/parse.py ex.py
  File "./scripts/parse.py", line 5
    return 1
    ^
SyntaxError: 'return' with value in async generator
```

But!

```python
async def g():
    yield
    return
```

Works!

```
» python ./scripts/parse.py ex.py
None
```

:facepalm:

sobolevn:
Ok, let do several things:
1. Forbid `StopAsyncIteration` inside both sync and async iterators
2. Forbid `return value` inside async generators (new separate violation)

bellum:
@sobolevn 
Didn't know that it's forbidden to use `return value` in async generators! 
I agree with you.

sobolevn:
I fixed all conflicts and merged this PR. Thanks a lot @bellum 👍

PR #669, opened 2019-07-14
Closes #658

### LINKED ISSUE #658: Create `TooManyImportedNamesViolation`

bellum:
# I have made things!

## Checklist

## Related issues

Closes #658

## Implementation details
Decided to extend existing `wemake_python_styleguide.visitors.ast.complexity.counts.ImportMembersVisitor` instead of creating new one. But that would be easy to change this on request.

@sobolevn review please when you have time


wemake_python_styleguide/options/defaults.py
```diff
@@ -57,6 +57,9 @@
 #: Maximum number of imports in a single module.
 MAX_IMPORTS: Final = 12
 
+#: Maximum number of imported names in a single module.
+MAX_IMPORTED_NAMES: Final = 50
```
sobolevn: Can you please write a comment about how this value was created? Like here https://github.com/wemake-services/wemake-python-styleguide/blob/dc96ff3dea3c5b759cd93bbb681702673b7b607f/wemake_python_styleguide/options/defaults.py#L55

wemake_python_styleguide/options/defaults.py
```diff
@@ -58,7 +58,7 @@
 MAX_IMPORTS: Final = 12
 
 #: Maximum number of imported names in a single module.
-MAX_IMPORTED_NAMES: Final = 50
+MAX_IMPORTED_NAMES: Final = 50  # this value was "guessed"
```
sobolevn: Nice one, haha 😆

sobolevn:
@bellum thanks a lot! Merged.

PR #678, opened 2019-07-18
Fix for issue 649.

### LINKED ISSUE #649: Forbid useless `return` and `return None` at the end of the function
# Rule request

## Thesis

Some functions return values: `return value`, some have only early returns `if ...: return`
Some `return None` as a value, some just can have it as the last statement.

This is wrong:

```python
def test():
     print('test')
     return

def test():
     print('test')
     return None

def test(some):
     if some:
          return
     print('test')
     return
```

These are correct:

```python
def test():
     print('test')
```

```python
def test(some):
     if some:
          return
     print('test')
```

```python
def test(some):
     if some:
          return some
     print('test')
     return None
```

The rule is related to #378 
The rule is:
1. Check if `return` is the last statement
2. Check if using pure `return`, then raise a violation
3. If using `return None` check if that is the only return in the function, if so raise a violation

## Reasoning

This is done for consistency. It should also improve the readability and typing.

NanduTej:
# I have made things!

## Checklist

- [ ] I have double checked that there are no unrelated changes in this pull request (old patches, accidental config files, etc)

### DIFF
--- a/wemake_python_styleguide/logic/returns.py
+++ b/wemake_python_styleguide/logic/returns.py
@@ -1,14 +1,18 @@
+# -*- coding: utf-8 -*-
+
 import ast
 from collections import defaultdict
 from typing import Dict, List, Tuple
 
 NamesAndReturns = Tuple[
     Dict[str, List[ast.Name]],
-    Dict[str, ast.Return]
+    Dict[str, ast.Return],
 ]
 
 
 def get_assign_node_variables(node: List[ast.AST]) -> List[str]:
+    """Gets the List of assign ids of the nodes."""
+
     assign = []
     for sub_node in node:
         if isinstance(sub_node, ast.Assign):
@@ -21,6 +25,8 @@ def get_assign_node_variables(node: List[ast.AST]) -> List[str]:
 
 
 def get_name_nodes_variable(node: List[ast.AST]) -> Dict[str, List[ast.Name]]:
+    """Gets the List of names of the nodes."""
+
     names: Dict[str, List[ast.Name]] = defaultdict(list)
     for sub_node in node:
         if isinstance(sub_node, ast.Name):
@@ -34,6 +40,8 @@ def get_name_nodes_variable(node: List[ast.AST]) -> Dict[str, List[ast.Name]]:
 
 
 def get_return_node_variables(node: List[ast.AST]) -> NamesAndReturns:
+    """Gets the return values and the return sub nodes in the code."""
+
     returns: Dict[str, List[ast.Name]] = defaultdict(list)
     return_sub_nodes: Dict[str, ast.Return] = defaultdict()
     for sub_node in node:
--- a/wemake_python_styleguide/logic/returns.py
+++ b/wemake_python_styleguide/logic/returns.py
@@ -1,0 +1,45 @@
+import ast
+from collections import defaultdict
+from typing import Dict, List, Tuple
+
+NamesAndReturns = Tuple[
+    Dict[str, List[ast.Name]],
+    Dict[str, ast.Return]
+]
+
+
+def get_assign_node_variables(node: List[ast.AST]) -> List[str]:
+    assign = []
+    for sub_node in node:
+        if isinstance(sub_node, ast.Assign):
+            if isinstance(sub_node.targets[0], ast.Name):
+                assign.append(sub_node.targets[0].id)
+        if isinstance(sub_node, ast.AnnAssign):
+            if isinstance(sub_node.target, ast.Name):
+                assign.append(sub_node.target.id)
+    return assign
+
+
+def get_name_nodes_variable(node: List[ast.AST]) -> Dict[str, List[ast.Name]]:
+    names: Dict[str, List[ast.Name]] = defaultdict(list)
+    for sub_node in node:
+        if isinstance(sub_node, ast.Name):
+            if isinstance(sub_node.ctx, ast.Load):
+                names[sub_node.id].append(sub_node)
+        if isinstance(sub_node, ast.AugAssign):
+            if isinstance(sub_node.target, ast.Name):
+                variable_name = sub_node.target.id
+                names[variable_name].append(sub_node.target)
+    return names
+
+
+def get_return_node_variables(node: List[ast.AST]) -> NamesAndReturns:
+    returns: Dict[str, List[ast.Name]] = defaultdict(list)
+    return_sub_nodes: Dict[str, ast.Return] = defaultdict()
+    for sub_node in node:
+        if isinstance(sub_node, ast.Return):
+            if isinstance(sub_node.value, ast.Name):
+                variable_name = sub_node.value.id
+                returns[variable_name].append(sub_node.value)
+                return_sub_nodes[variable_name] = sub_node
+    return returns, return_sub_nodes
--- a/wemake_python_styleguide/logic/returns.py
+++ b/wemake_python_styleguide/logic/returns.py
@@ -12,7 +12,6 @@ NamesAndReturns = Tuple[
 
 def get_assign_node_variables(node: List[ast.AST]) -> List[str]:
     """Gets the List of assign ids of the nodes."""
-
     assign = []
     for sub_node in node:
         if isinstance(sub_node, ast.Assign):
@@ -26,7 +25,6 @@ def get_assign_node_variables(node: List[ast.AST]) -> List[str]:
 
 def get_name_nodes_variable(node: List[ast.AST]) -> Dict[str, List[ast.Name]]:
     """Gets the List of names of the nodes."""
-
     names: Dict[str, List[ast.Name]] = defaultdict(list)
     for sub_node in node:
         if isinstance(sub_node, ast.Name):
@@ -41,7 +39,6 @@ def get_name_nodes_variable(node: List[ast.AST]) -> Dict[str, List[ast.Name]]:
 
 def get_return_node_variables(node: List[ast.AST]) -> NamesAndReturns:
     """Gets the return values and the return sub nodes in the code."""
-
     returns: Dict[str, List[ast.Name]] = defaultdict(list)
     return_sub_nodes: Dict[str, ast.Return] = defaultdict()
     for sub_node in node:
--- a/wemake_python_styleguide/visitors/ast/keywords.py
+++ b/wemake_python_styleguide/visitors/ast/keywords.py
@@ -1,8 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import ast
-from collections import defaultdict
-from typing import ClassVar, Dict, List, Tuple, Type, Union
+from typing import ClassVar, List, Type, Union
 
 from typing_extensions import final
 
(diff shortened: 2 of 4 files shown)


NanduTej:
Hey @sobolevn! I have made the changes and I think they work as they are supposed to. Can you check and let me know?

sobolevn:
Looking good!
Can you please rebase you changes? And I will merge them into `0.12`

NanduTej:
@sobolevn I did the rebase.

wemake_python_styleguide/visitors/ast/keywords.py
```diff
@@ -301,6 +302,21 @@ def _check_for_violations(self, names, return_sub_nodes, returns) -> None:
                     ),
                 )
 
+    def _check_return_at_the_end(self, node):
```
sobolevn: This function should be typed.
Like so: `_check_return_at_the_end(self, node: AnyFunctionDef) -> None:`

NanduTej: @sobolevn added the change.

sobolevn:
@NanduTej please, fix style 🙂 

You can run `make test` locally to get the same checks CI has.

wemake_python_styleguide/visitors/ast/keywords.py
```diff
@@ -311,11 +311,11 @@ def _check_return_at_the_end(self, node) -> None:
                 self.add_violation(
                     InconsistentReturnVariableViolation(last),
                 )
-            elif isinstance(last.value, ast.NameConstant) and (last.value.value is None):
-                self.add_violation(
-                    InconsistentReturnVariableViolation(last),
-                )
-
+            elif isinstance(last.value, ast.NameConstant):
+                if (last.value.value is None):
```
sobolevn: No need to `( ... )` here.

NanduTej: Yeah, just saw that :sweat_smile: .

sobolevn:
There are now issues visible to CI in the projects right now: https://travis-ci.org/wemake-services/wemake-python-styleguide/builds

But, new rules - reveal new problems! That's why we build new rules 💪 
There are several things to fix here in your PR.

## First issue

```python
    def _check_first_element(
        self,
        node: ast.AST,
        statement: ast.AST,
        extra_lines: int,
    ) -> Optional[bool]:
        if statement.lineno == node.lineno and not extra_lines:
            return False
        return None
```

This code should not raise any violations. It is correct. See the original issue (3rd example):
Снимок экрана 2019-07-18 в 15 28 10

This code from `./wemake_python_styleguide/logic/naming/name_nodes.py` is also correct:

```python
def get_assigned_name(node: ast.AST) -> Optional[str]:
    """
    Returns variable names for node that is just assigned.

    Returns ``None`` for nodes that are used in a different manner.
    """
    if isinstance(node, ast.Name) and isinstance(node.ctx, ast.Store):
        return node.id

    if isinstance(node, ast.Attribute) and isinstance(node.ctx, ast.Store):
        return node.attr

    if isinstance(node, ast.ExceptHandler):
        return getattr(node, 'name', None)

    return None  # this is a meaningful value!
```

**Please**, make sure that all the example from the original task works.
Add them in tests if they are missing right now.

## Second issue

```
./wemake_python_styleguide/visitors/ast/keywords.py
  215:1    WPS214 Found too many methods: 8
  @final
  ^
```

It means that your new method makes the class too complex. Here's how to fix it:
Move some logic from the class. How to define that this is pure logic and not methods? Look at `_get_return_node_variables` and `_get_name_nodes_variable` and `_get_assign_node_variables` - they don't use `self` at all (we might actually build a new rule to check that). It means that they can be extracted to `logic/returns.py` file. Then you will have extra capacity to add your new method without this violation.

Thanks you for your work! I really appreciate it 👍 
Let's make an awesome product together! 🎉

NanduTej:
@sobolevn sorry for spamming so many commits, my flake8 does not give me the full messages as the build was giving me, so I pushed all the changes I could see and tested them. I am still confused as to what the issue is with the 2 'return None'. I have added all the tests mentioned in the issue and all of them are passing.

sobolevn:
@NanduTej the easiest way to fix it is:

1. create `ex.py` with the code that raises a violation
2. run it with `scripts/parse.py` to see the ast, it might give you a hint on what is going on
3. debug your visitor on why it is happening: use `print(node, vars(node))` or `pdb` 
4. add this code sample to the tests, so we can be sure that the problem is fixed, more tests = better product

sobolevn:
If you setup does not match the one from CI, try the clean build:

```bash
deactivate
rm -rf dist pip-wheel-metadata wemake_python_styleguide.egg-info .venv
poetry install
poetry run make test
```

NanduTej:
> If you setup does not match the one from CI, try the clean build:
> 
> ```shell
> deactivate
> rm -rf dist pip-wheel-metadata wemake_python_styleguide.egg-info .venv
> poetry install
> poetry run make test
> ```

Thanks! This worked.

sobolevn:
@NanduTej how's it going? Do you need any help from my side? 🙂

NanduTej:
@sobolevn Sorry, didn't finish the work on this. I am out for my convocation ceremony😄. Will get back and work on it in 3 days.

sobolevn:
Had to google what "convocation ceremony" is 😆 
Congratulations! 🎉 

Feel free to get back to it when you have the time.

NanduTej:
Hey @sobolevn! I tried to debug but could not get anywhere. Can you help me out further? I did create the ex.py file and pasted the 2 pieces of code and ran the parse.py program. I did not find anything out of order in the output.

sobolevn:
@NanduTej what exactly is your problem?

NanduTej:
On running flake8 ., There is an error popping up.

__________________________________________________________________
./wemake_python_styleguide/logic/naming/name_nodes.py

  29:5     WPS331 Found local variable that are only used in `return` statements
  return None
  ^

./wemake_python_styleguide/visitors/ast/statements.py

  219:9    WPS331 Found local variable that are only used in `return` statements
  return None
  ^
_________________________________________________________________

I am unable to debug this.

wemake_python_styleguide/visitors/ast/keywords.py
```diff
@@ -280,12 +240,13 @@ def _check_variables_for_return(self, node: AnyFunctionDef) -> None:
                 ast.walk(node),
             ),
         )
-        assign = self._get_assign_node_variables(nodes)
-        names = self._get_name_nodes_variable(nodes)
-        returns, return_sub_nodes = self._get_return_node_variables(nodes)
+        assign = get_assign_node_variables(nodes)
+        names = get_name_nodes_variable(nodes)
+        returns, return_sub_nodes = get_return_node_variables(nodes)
```
sobolevn: You can insert `breakpoint()` (if you are using 3.7, otherwise see https://realpython.com/python-debugging-pdb/) and debug this function.

sobolevn: You can also use `ipbd`:
1. `pip install ipdb`
2. `export PYTHONBREAKPOINT=ipdb.set_trace`
3. Set `breakpoint()` where you need it
4. Run `flake8` on file that raises an unwanted violation

NanduTej:
Hey @sobolevn! Can you explain this to me?
https://github.com/wemake-services/wemake-python-styleguide/blob/b6b273828a8aae20eb1272f8841a4cb904f6def0/tests/test_checker/test_noqa.py#L87

sobolevn:
@NanduTej this is a integration tests for several things:
1. that violation is active and enabled
2. that violation is raised for the bad code
3. that line number where violation is raised is correct
4. that `noqa` works
Docs: https://github.com/wemake-services/wemake-python-styleguide/blob/master/CONTRIBUTING.md#before-submitting

NanduTej:
Hey @sobolevn! I am sorry I am taking so much time with this. Can you help me with understanding the flow of how flake8 works. Maybe that will give me a clue as to where to look for the error.

sobolevn:
@NanduTej can you please show me the piece of code that does not work for you? Maybe I can help you from there. Since, I am pretty sure that `flake8` has nothing to do with this task.

(In case you still need this information, it is right in our docs: https://wemake-python-stylegui.de/en/latest/pages/api/index.html#overview and here's how all classes work: https://wemake-python-stylegui.de/en/latest/pages/api/index.html#api-reference)

NanduTej:
On running flake 8, I run into this error. 
./wemake_python_styleguide/logic/naming/name_nodes.py
______________________________________________________________________________
  29:5     WPS331 Found local variable that are only used in `return` statements
  return None
  ^

./wemake_python_styleguide/visitors/ast/statements.py

  219:9    WPS331 Found local variable that are only used in `return` statements
  return None
  ^

______________________________________________________________________________
The code is executing fine with all my tests. When I push my code, this error will make build fail which is coming when flake8 command is being run.

sobolevn:
So, then add these cases to your tests:

```python
correct_example_N = """
def get_assigned_name(node: ast.AST) -> Optional[str]:
    if isinstance(node, ast.Name) and isinstance(node.ctx, ast.Store):
        return node.id

    if isinstance(node, ast.Attribute) and isinstance(node.ctx, ast.Store):
        return node.attr

    if isinstance(node, ast.ExceptHandler):
        return getattr(node, 'name', None)

    return None
"""
```

And:

```python
correct_example_N1 = """
class WrongAttributeVisitor(object):
     def _is_super_called(self, node: ast.Call) -> bool:
        if isinstance(node.func, ast.Name):
            if node.func.id == 'super':
                return True
        return False
"""
```

And then debug what is going on: I wrote a debugging section of documentation just for this case. 
https://wemake-python-stylegui.de/en/latest/pages/api/debugging.html

P.S. Don't forget to rebase your code.

wemake_python_styleguide/logic/returns.py
```diff
@@ -0,0 +1,50 @@
...
+    return names
+
+
+def get_return_node_variables(node: List[ast.AST]) -> NamesAndReturns:
+    """Gets the return values and the return sub nodes in the code."""
+    returns: Dict[str, List[ast.Name]] = defaultdict(list)
+    return_sub_nodes: Dict[str, ast.Return] = defaultdict()
+    for sub_node in node:
+        if isinstance(sub_node, ast.Return):
+            if isinstance(sub_node.value, ast.Name):
```
NanduTej: For the correct_example_N we are not collecting the previous returns where I expected them to be collected in return_sub_nodes and so we are running into issues here. The correct_example_N1 is passing the tests. How should we collect those returns here? Example the returns of type ast.Attribute, and should we be collecting them here?
@sobolevn

sobolevn: Sorry, I have missed your question. Can you try to use `ast.walk(subnode)` there? I guess that happens because they are not direct children of the give node, but are lower down the tree.

sobolevn: You can try some other task if you are stuck with one. And pass this one to me.

NanduTej: Sure

PR #687, opened 2019-07-20
Closes #684

### LINKED ISSUE #684: Forbid magic methods to be async
# Rule request

## Thesis

Only several magic method can be `async`. Other method should be just functions.

## Reasoning

Having `async __lt__(self, ...) -> ...` is incorrect. It cannot be used like so.
The same applies to almost all other method including `__init__`, `__new__`, `__bool__`, etc.
Only `__aenter__`, `__aexit__` and possibly some other can be `async`.

We must have an explicit list of whitelisted async magic methods.

bellum:
# I have made things!

## Checklist

## Related issues

Closes #684

### DIFF
--- a/wemake_python_styleguide/violations/oop.py
+++ b/wemake_python_styleguide/violations/oop.py
@@ -408,8 +408,8 @@ class AsyncMagicMethodViolation(ASTViolation):
     """
     Forbids to make some magic methods async.
 
-    We forbid to make __init__, __eq__, __lt__, etc async.
-    We allow to make __anext__, __aenter__, __aexit__ async.
+    We forbid to make ``__init__``, ``__eq__``, ``__lt__``, etc async.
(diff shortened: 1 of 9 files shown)


wemake_python_styleguide/violations/oop.py
```diff
@@ -399,3 +401,45 @@ class DirectMagicAttributeAccessViolation(ASTViolation):
...
     code = 609
     previous_codes = {462}
+
+
+@final
+class AsyncMagicMethodViolation(ASTViolation):
+    """
+    Forbids to make some magic methods async.
+
+    We forbid to make __init__, __eq__, __lt__, etc async.
```
sobolevn: Can you please add double backticks (```) around values? Otherwise `__` will turn them into bold text.

bellum: Of course. Done.

sobolevn:
Thanks a lot! 🚀

PR #707, opened 2019-07-23
Just correcting misspell in a couple words.

ssilaev:


kxepal:
There should be another linter for that around.

ssilaev:
> There should be another linter for that around.

Actually, I would not seeing it if don't go a dig into the last commits.

kxepal:
Sorry, I forgot the most important part of my message: :wink:

PR #748, opened 2019-08-06
#737 Forbids too deep access level

### LINKED ISSUE #737: Forbid too deep subscript and attribute usages
# Rule request

## Thesis

Code like:

```python
print(my_matrix[0][0][0][0][0][0][0][1])
```

And code like:

```python
self.attr.inner.wrapper.field.value

self.attr.inner.wrapper.method.call()
```

When counting access level we count:
1. `ast.Subscript`
2. `ast.Attribute`

Things we do not count:
1. `ast.Index` complexity
2. `ast.Call` (since there are a lot of `call`-based APIs like `django` ORM: `.filter().exclude().annotate().values().first()` which we do not want to break)

That's how it works:

```python
obj.attr.call().other[0].field.type.boom
# obj has access level of one: `.attr`
# `call()` has access level of 5: `.other`, `[0]`, `.field`, `.type`, `.boom`
```

Should be forbidden. We need to make this complexity rule configurable.
Default value can be `4`, chains longer than `4` should raise a violation.

## Reasoning

It indicates a bad design and overcomplicated data without proper API.

artemiy312:
# I have made things!

## Checklist

## Related issues

Closes #737


tests/fixtures/noqa.py
```diff
@@ -146,6 +146,7 @@ class ManyParents(First, Second, Third, Exception):  # noqa: WPS215
     assert test_function(5)
     assert test_function(6)
 
+some.other[0].field.type.boom  # noqa: WPS219,WPS428
```
sobolevn: This should be an assignment, so it won't raise `MeaninglessExpressionViolation`

tests/test_visitors/test_ast/test_complexity/test_access/test_access.py
```diff
@@ -0,0 +1,61 @@
+# -*- coding: utf-8 -*-
+
+import pytest
+
+from wemake_python_styleguide.visitors.ast.complexity.access import (
+    AccessVisitor,
+    TooDeepAccessViolation,
```
sobolevn: Please, import violations directly. And not from `visitors`

tests/test_visitors/test_ast/test_complexity/test_access/test_access.py
```diff
@@ -0,0 +1,61 @@
...
+    assert_error_text,
+    parse_ast_tree,
+    code,
+    options,
+    mode,
+):
+    """Testing that violations are raised when reaching too deep access."""
+    tree = parse_ast_tree(mode(code))
+
+    option_values = options(max_access_level=2)
```
sobolevn: Let's test border values: let it be `3`

tests/test_visitors/test_ast/test_complexity/test_access/test_access.py
```diff
@@ -0,0 +1,61 @@
...
+
+from wemake_python_styleguide.visitors.ast.complexity.access import (
+    AccessVisitor,
+    TooDeepAccessViolation,
+)
+
+subscript_access = 'my_matrix[0][0][0][0]'
+attribute_access = 'self.attr.inner.wrapper.value'
+mixed_access = 'self.attr[0].wrapper[0]'
+mixed_with_calls_access = 'self.attr[0]().wrapper[0][0].bar().foo[0]()'
```
sobolevn: We also need a case with `call` but too long.

wemake_python_styleguide/logic/prop_access.py
```diff
@@ -33,3 +35,20 @@ def parts(node: Union[ast.Attribute, ast.Subscript]) -> Iterable[ast.AST]:
...
+    Returns consecutive attribute and subscript accesses.
+
+    We get all parts from it except ``obj: ast.Name``.
+
+    .. code:: python
+
+      obj.attr.other[0].field
+
+    """
+    for part in parts(node):  # pragma: no cover, can't complete the loop
```
sobolevn: No, sorry. `# pragma: no cover` is not acceptable here. It means that:
1. You either have insufficient test cases
2. Or you have wrong logic that never executes

This one needs to be fixed.

wemake_python_styleguide/violations/complexity.py
```diff
@@ -572,6 +574,50 @@ class TooManyAssertsViolation(ASTViolation):
...
+        Having too deep access level indicates a bad design
+        and overcomplicated data without proper API.
+
+    Solution:
+        Split the expression into variables, functions or classes.
+        Refactor the API for your data layout.
+
+    Example::
+
+        # correct: access level — 4
```
sobolevn: `correct` -> `Correct`

wemake_python_styleguide/violations/complexity.py
```diff
@@ -572,6 +574,50 @@ class TooManyAssertsViolation(ASTViolation):
...
+        self.attr.inner.wrapper[1]
+
+        # wrong: access level — 5
```
sobolevn: `wrong` -> `Wrong`

wemake_python_styleguide/violations/complexity.py
```diff
@@ -572,6 +574,50 @@ class TooManyAssertsViolation(ASTViolation):
...
 
 
+@final
+class TooDeepAccessViolation(ASTViolation):
+    """
+    Forbids to have consecutive expressions with too deep access level.
+
+    We consider these expressions as accesses:
+
+        ``ast.Subscript``
```
sobolevn: Should be a list:

```
- ``ast.Subscript``
- ``ast.Attribute``
```

Please, also mention that ``ast.Call` is treated differently (and why so).

artemiy312:
@sobolevn I have fixed minor issues. Please, take a look again

tests/test_logic/test_prop_access.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+
+import ast
+
+from wemake_python_styleguide.logic import prop_access
+
+
+def test_accesses():
```
sobolevn: Sorry, but this code still does not look right. I will explain in one of the next comments.

wemake_python_styleguide/visitors/ast/complexity/access.py
```diff
@@ -0,0 +1,60 @@
...
+
+        """
+        self._check_consecutive_access_number(node)
+        self.generic_visit(node)
+
+    def _check_consecutive_access_number(self, node: AnyAccess) -> None:
+        if node in self._visited_accesses:
+            return
+
+        access_chain = set(accesses(node))
```
sobolevn: Now I understand what you are trying to achieve. You need to filter out `ast.Call` nodes from `(ast.Attribute, ast.Subscript)`. You can use in-place `filter(..., parts(node))` call and drop `accesses` function at all.

tests/test_logic/test_prop_access.py
sobolevn: Done! Please, remove this test and `accesses` function.

artemiy312:
@sobolevn Done! I have dropped `access` function

wemake_python_styleguide/visitors/ast/complexity/access.py
```diff
@@ -0,0 +1,67 @@
...
+        self.generic_visit(node)
+
+    def _is_any_access(self, node: ast.AST) -> bool:
+        return isinstance(node, (ast.Attribute, ast.Subscript))
+
+    def _check_consecutive_access_number(self, node: AnyAccess) -> None:
+        if node in self._visited_accesses:
+            return
+
+        consecutive_access = cast(Set[AnyAccess], set(takewhile(
```
sobolevn: Why do we need a `cast` here? Is it possible to have: `consecutive_access: Set[AnyAccess] = set(takewhile(..))` ?

tests/test_visitors/test_ast/test_complexity/test_access/test_access.py
```diff
@@ -0,0 +1,68 @@
...
+)
+
+# boundary expressions
+subscript_access = 'my_matrix[0][0][0][0]'
+attribute_access = 'self.attr.inner.wrapper.value'
+mixed_access = 'self.attr[0].wrapper[0]'
+mixed_with_calls_access = 'self.attr[0]().wrapper[0][0].bar().foo[0]()'
+
+# correct expressions
+call_chain = 'manager.filter().exclude().annotate().values().first()'
```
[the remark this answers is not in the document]
artemiy312: Sure

wemake_python_styleguide/visitors/ast/complexity/access.py
artemiy312: `mypy` complains on this `consecutive_access: Set[AnyAccess] = set(takewhile(..))`:
```
wemake_python_styleguide/visitors/ast/complexity/access.py:59: error: Argument 2 to "takewhile" has incompatible type "Iterable[AST]"; expected "Iterable[Union[Attribute, Subscript]]"
```

`parts` returns `Iterator[ast.AST]`, so I have to [downcast to subtype](https://mypy.readthedocs.io/en/latest/casts.html) explicitly or use `Set[ast.AST]` instead. But here I should use specific type, not generic `ast.AST`

sobolevn: How do you think: can we improve `paths` types?

artemiy312: `paths` returns `ast.expr`, because the function depends on child node attributes: `Attribute.value`, `Subscript.value` and `Call.func`, which have `ast.expr` types.

I don't know how to improve it.

artemiy312:
>How do you like our tool and the process?
Do you have any feedback?

It's great! Tests and linters led me to solve the issue, showed me what I have to do to fit the project. Docs provided basic info about domain and how to become a contributor. 

But it was not without hurdles. The first thing I did was a violation subclass, which broke some tests. I was happy about it, but not all failed tests showed helpful messages to figure out what should I fix. Therefore, I had to read test sources, make guesses, run tests again and again :)
I believe failed tests should propose a way to fix them if it is possible.

message examples that can improve tests:
- Sort `autoclass` directives docs by **violation codes**

```python
    def test_violation_autoclass_order(all_module_violations):
        """Used to force violations order inside the `autoclass` directives."""

>           assert sorted_by_code == sorted_by_autoclass
E           AssertionError: assert ['JonesScoreV...olation', ...] == ['JonesScoreVi...olation', ...]
E             Left contains one more item: 'ExampleViolation'
E             Use -v to get the full diff
```

- Add unlisted WPS230 violation to tests/fixtures/noqa.py.
```python
    def test_noqa_fixture_disabled(absolute_path, all_violations):
        """End-to-End test to check that all violations are present."""

>           assert key in errors, 'Unlisted #noqa violation'
E           AssertionError: Unlisted #noqa violation
E           assert 'WPS230' in mappingproxy({'WPS100': 0, 'WPS101': 0, 'WPS102': 0, 'WPS110': 3, 'WPS111': 1, 'WPS112': 1, 'WPS113': 1, 'WPS114': 1, ..., 'WPS604': 2, 'WPS605': 1, 'WPS606': 1, 'WPS607': 1, 'WPS608': 1, 'WPS609': 1, 'WPS610': 1, 'WPS611': 1, 'WPS612': 1})
``` 

It is not a complete list, but it show the point. 

@sobolevn Thanks for the brilliant project!

sobolevn:
@artemiy312 I agree, that's a problem indeed. Can you please create a new issue out of it? Let's improve this part.

PR #804, opened 2019-09-08
Ignore overload from typing

ea-volkov:
# I have made things!

## Checklist

### DIFF
--- a/tests/test_visitors/test_ast/test_blocks/test_overlap/test_function_block.py
+++ b/tests/test_visitors/test_ast/test_blocks/test_overlap/test_function_block.py
@@ -199,19 +199,9 @@ overload_template = """
 
 @pytest.mark.parametrize('import_overload', [
     """
-from typing import overload
 @overload
     """,
     """
-from typing import overload as ovrld
-@ovrld
-""",
-    """
-import typing as tp
-@tp.overload
-    """,
-    """
-import typing
 @typing.overload
     """,
 ])
@@ -237,30 +227,13 @@ def test_function_overload(
 
 @pytest.mark.parametrize('decorator_tempate', [
     """
-import typing
 @typing.func
     """,
     """
-from module import func as fc
-@fc
+@module.overload
     """,
     """
-from typing import func
-@func
-    """,
-    """
-from module import overload
-@overload
-    """,
-    """
-def decorate(func):
-    return None
 @decorate
-    """,
-    """
-def overload(func):
-    return None
-@overload
     """,
 ])
 def test_no_function_overload(
--- a/tests/test_visitors/test_ast/test_blocks/test_overlap/test_function_block.py
+++ b/tests/test_visitors/test_ast/test_blocks/test_overlap/test_function_block.py
@@ -198,12 +198,8 @@ overload_template = """
 
 
 @pytest.mark.parametrize('import_overload', [
-    """
-@overload
-    """,
-    """
-@typing.overload
-    """,
+    '@overload',
+    '@typing.overload',
 ])
 def test_function_overload(
     assert_errors,
@@ -224,15 +220,9 @@ def test_function_overload(
 
 
 @pytest.mark.parametrize('decorator_template', [
-    """
-@typing.func
-    """,
-    """
-@module.overload
-    """,
-    """
-@decorate
-    """,
+    '@typing.func',
+    '@module.overload',
+    '@decorate',
 ])
 def test_no_function_overload(
     assert_errors,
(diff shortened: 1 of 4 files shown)


ea-volkov:
@sobolevn, please, review my changes. Did I understand the task correctly?
There is an error in the statement of the problem - WPS440 instead WPS430.

tests/test_visitors/test_ast/test_blocks/test_overlap/test_function_block.py
```diff
@@ -183,3 +183,101 @@ def test_function_block_correct(
...
+overload_template = """
+{0}
+{1}
+{1}
+"""
+
+
+@pytest.mark.parametrize('import_overload', [
+    """
+from typing import overload
```
sobolevn: we dont' need an import here. just the decorator

wemake_python_styleguide/visitors/ast/blocks.py
```diff
@@ -73,6 +73,8 @@ class BlockVariableVisitor(base.BaseNodeVisitor):
 
     """
 
+    _ast_func = (ast.FunctionDef, ast.AsyncFunctionDef)
```
sobolevn: we already have this defined: https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/compat/aliases.py#L19

wemake_python_styleguide/visitors/ast/blocks.py
```diff
@@ -151,6 +153,25 @@ def visit_locals(self, node: Union[AnyAssign, ast.arg]) -> None:
 
     # Utils:
 
+    def _is_function_overload(self, node: ast.AST):
+        if isinstance(node, self._ast_func):
+            _scope = BlockScope(node)
+            for decorator in node.decorator_list:
+                if (
```
sobolevn: This solution is overly complicated. We do not need to check where the import is from. We don't care about the type. The only thing we care about is the name.

So, this can be changed to the following: 

```python
if node_to_string(decorator) in {'overload', 'typing.overload'}: ...
```

wemake_python_styleguide/visitors/ast/blocks.py
```diff
@@ -73,7 +78,7 @@ class BlockVariableVisitor(base.BaseNodeVisitor):
 
     """
 
-    _ast_func = (ast.FunctionDef, ast.AsyncFunctionDef)
+    _overload_exceptions = {'overload', 'typing.overload'}
```
sobolevn: Needs to be a `frozenset`

wemake_python_styleguide/visitors/ast/blocks.py
```diff
@@ -73,6 +78,8 @@ class BlockVariableVisitor(base.BaseNodeVisitor):
 
     """
 
+    _overload_exceptions = frozenset({'overload', 'typing.overload'})
```
sobolevn:
```suggestion
    _overload_exceptions = frozenset(('overload', 'typing.overload'))
```

ea-volkov:
@sobolevn may I use WPS214 for passing lint stage?

wemake_python_styleguide/visitors/ast/blocks.py
```diff
@@ -151,6 +158,13 @@ def visit_locals(self, node: Union[AnyAssign, ast.arg]) -> None:
 
     # Utils:
 
+    def _is_function_overload(self, node: ast.AST):
```
sobolevn: This can actually be a function, not a method. `self._overload_exceptions` can be used as a regular set.

tests/test_visitors/test_ast/test_blocks/test_overlap/test_function_block.py
```diff
@@ -183,3 +183,70 @@ def test_function_block_correct(
...
+overload_template = """
+{0}
+{1}
+{1}
+"""
+
+
+@pytest.mark.parametrize('import_overload', [
+    """
+@overload
```
sobolevn: `'@overload'` and `'@typing.overload'`

tests/test_visitors/test_ast/test_blocks/test_overlap/test_function_block.py
```diff
@@ -183,3 +183,70 @@ def test_function_block_correct(
...
+
+    visitor = BlockVariableVisitor(default_options, tree=tree)
+    visitor.run()
+
+    assert_errors(visitor, [])
+
+
+@pytest.mark.parametrize('decorator_template', [
+    """
+@typing.func
```
sobolevn: The same here: for some reason `"""` is used for a single line string

PR #812, opened 2019-09-15
Ignore overlap for names imported from other modules

ea-volkov:
# I have made things!

## Checklist

### DIFF
--- a/wemake_python_styleguide/logic/scopes.py
+++ b/wemake_python_styleguide/logic/scopes.py
@@ -8,6 +8,10 @@ from typing_extensions import final
 
 from wemake_python_styleguide.compat.aliases import FunctionNodes
 from wemake_python_styleguide.logic.naming import access, name_nodes
+from wemake_python_styleguide.logic.naming.name_nodes import (
+    flat_variable_names,
+    get_variables_from_node,
+)
 from wemake_python_styleguide.logic.nodes import get_context
 from wemake_python_styleguide.logic.source import node_to_string
 from wemake_python_styleguide.types import ContextNodes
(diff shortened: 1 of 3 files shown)


wemake_python_styleguide/visitors/ast/blocks.py
```diff
@@ -161,11 +161,11 @@ def _scope(
     ) -> None:
         scope = BlockScope(node)
         shadow = scope.shadowing(names, is_local=is_local)
-
         if shadow:
-            self.add_violation(
-                BlockAndLocalOverlapViolation(node, text=', '.join(shadow)),
-            )
+            if not is_imported_var_assigned(node):
```
sobolevn: Sorry, I was not clear enough in my task description. As a result you took a wrong path. Let me explain the task idea once again.

Imagine that we have a block variable `my_var`. We don't care about how exactly it is defined. We only care that it is the block one.

We also have a local variable `my_var` that shadows the block one. Normally, we would raise a violation. But, in case if local `my_var` definition uses the block one - it is okay.

Example:

```python
for my_var in some_attrs():
    my_var = my_var.lower()  # ok
```

ea-volkov: @sobolevn do I disable this rule for cases with any block variables?

sobolevn: Sorry, I don't get your question. You can use Russian in case it is easier for you.

ea-volkov: Переопределение в любом блоке должно быть разрешено для случаев, когда новое значение использует старое?

sobolevn: Yeap 👍

wemake_python_styleguide/logic/scopes.py
```diff
@@ -145,46 +149,37 @@ def is_function_overload(node: ast.AST) -> bool:
...
-            else el
-            for el in node.elts
-        ]
-        return any(
-            (_is_var_imported_from(assigned) for assigned in assigned_vars),
-        )
-    return False
+def _get_variables_from_complex_node(node: ast.AST) -> Set[str]:
+    names = set(get_variables_from_node(node))
+    if isinstance(node, ast.Attribute):
```
sobolevn: We should check for just `ast.Name` here. Are there any other cases that are not covered with `ast.Name`?

PR #844, opened 2019-10-06
Fix ImplicitElifViolation false positives on try/except and for/else.

### LINKED ISSUE #835: `if` after `else` in `try` raises false positive WPS513
# Bug report

## What's wrong

Code:

```python
        try:
            return_value = ast.literal_eval(node.body)
        except ValueError:
            # The value was not literal, it's ok.
            return
        else:
            if return_value in self._primitive_values:
                self.add_violation(ImplicitPrimitiveViolation(node))
```

Output:

```
329:13   WPS513 Found implicit `elif` condition
  if return_value in self._primitive_values:
```

## How is that should be

It should not raise this violation. It is not a part of `if` / `else` case.

lensvol:
So this implements additional heuristics to prevent `ImplicitElifViolation` triggering on the following cases:

```python
try:
    ...
except ValueError:
    ...
else:
    ...
```

and

```python
for a in b:
    ...
else:
    ...
```

There is one line covered by `pragma: no cover`, as it can only trigger only when running this tool on obviously broken code and crafting a specific test case for this unlikely eventuality proved burdensome.

## Checklist

## Related issues

Closes #835

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ We used to have incremental versioning before `0.1.0`.
 ### Features
 
 - Forbids using `Literal[None]` in function annotations
+- Fixes 
 
 
 ## 0.13.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,10 @@ We used to have incremental versioning before `0.1.0`.
 ### Features
 
 - Forbids using `Literal[None]` in function annotations
-- Fixes 
+
+### Bugfixes
+
+- Fixes **ImplicitElifViolation** false positives on a specific edge cases.
 
 
 ## 0.13.0
--- a/tests/test_visitors/test_tokenize/test_conditions/test_implicit_elif.py
+++ b/tests/test_visitors/test_tokenize/test_conditions/test_implicit_elif.py
@@ -46,6 +46,30 @@ else:
         ...
 """
 
+# False positives
+
+for_else = """
+for a in b:
+   ...
+else:
+   if some:
+       ...
+"""
+
+try_except_else = """
+try:
+   ...
+except:
+   ...
+else:
+   if some:
+       ...
+"""
+
+embedded_else = """
+... if ... else ...
+"""
+
 
 @pytest.mark.parametrize('code', [
     elif_cases,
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ We used to have incremental versioning before `0.1.0`.
 ### Features
 
 - Forbids using `Literal[None]` in function annotations
+- Fixes
```
sobolevn: Looks unfinished 🙂 And it should be located in `### Bugfixes` section

tests/test_visitors/test_tokenize/test_conditions/test_implicit_elif.py
```diff
@@ -46,6 +46,30 @@
         ...
 """
 
+# False positives
```
sobolevn: Can you please provide a link to the original issue here?

wemake_python_styleguide/visitors/tokenize/conditions.py
```diff
@@ -57,11 +57,36 @@ def visit_name(self, token: tokenize.TokenInfo) -> None:
...
+        for token in reversed(self.file_tokens[:start_index - 1]):
+            if token.type != tokenize.NAME:
+                continue
+
+            # Here we rely upon an intuition that in Python else have to be
+            # on the same level (same indentation) as parent statement.
+            if token.start[1] == previous_token.start[1]:
+                return token.string in {'if', 'elif'}
+
+        return False  # pragma: no cover
```
sobolevn: Why do we need a `no cover` here? That's a bad sign.

wemake_python_styleguide/visitors/tokenize/conditions.py
```diff
@@ -57,11 +57,36 @@ def visit_name(self, token: tokenize.TokenInfo) -> None:
         """
         self._check_implicit_elif(token)
 
+    def _check_else_belongs_to_if(self, start_index: int) -> bool:
```
sobolevn: We use `_check` prefix for methods that work with nodes. Let's rename it to `_does_belong_to_if`

wemake_python_styleguide/visitors/tokenize/conditions.py
lensvol: I've mentioned that in the description of PR. Short version: it is very hard to craft a specific piece of code to trigger that line, and it will be incorrect Python code at the same time.

If insist, I can do that.

CHANGELOG.md
lensvol: Oops, my bad, untimely Ctrl-Z. :D Will fix.

tests/test_visitors/test_tokenize/test_conditions/test_implicit_elif.py
lensvol: Ok!

wemake_python_styleguide/visitors/tokenize/conditions.py
sobolevn: Can we rewrite the logic then?

lensvol: Sure. :)

PR #846, opened 2019-10-06
Add rule about vague import names.

### LINKED ISSUE #814: Forbid `from-import` for meaningless names
# Rule request

## Thesis

Some packages designed to be imported only as module, and functions inside have name that completes module name.

Bad:

```python
from attr import s
from json import load, dumps
from tomlkit import parse
from schemathesis import from_file
```

## Reasoning

When you see `datetime(...)` in the code, you know that it's from `datetime`. When you see `BaseView` in a Django project, you know that it's from Django framework. When you see `loads`, it can be anything: yaml, toml, json, custom config.

## Proposed solution

Consider the next blacklist:

+ `load`, `loads`
+ `dump`, `dumps`
+ `parse`
+ PyYAML specific: `safe_load`, `safe_dump`, `load_all`, `dump_all`, `safe_load_all`, `safe_dump_all`
+ one-letter names
+ Anything that starts from `from_` or `to_`

lensvol:
Implemented this using the original proposal from #814, opting for the broader scope of the check - all the imported method names are checked against blacklist, not only those relating to `json` or `yaml`.

I feel that this may be a lapse of judgment on my part, so feel free to suggest restricting blacklist on per-module basis.

## Checklist

## Related issues

Closes #814

### DIFF
--- a/tests/test_visitors/test_ast/test_imports/test_vague_imports.py
+++ b/tests/test_visitors/test_ast/test_imports/test_vague_imports.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+
+import pytest
+
+from wemake_python_styleguide.violations.consistency import (
+    VagueImportViolation,
+)
+from wemake_python_styleguide.visitors.ast.imports import WrongImportVisitor
+
+import_template = 'from mod import {0}'
+
+blacklisted_method_names = [
+    'dump',
+    'dumps',
+    'load',
+    'loads',
+    'parse',
+    'safe_dump',
+    'safe_dump_all',
+    'load_all',
+    'dump_all',
+    'safe_load_all',
+    'safe_dump_all',
+]
+
+
+@pytest.mark.parametrize(
+    'method_name',
+    blacklisted_method_names + ['a', 'from_something', 'to_something'],
+)
+def test_special_cases_vague_import(
+    assert_errors,
+    parse_ast_tree,
+    method_name,
+    default_options,
+):
+    """Testing that imports with the same aliases are restricted."""
+    tree = parse_ast_tree(import_template.format(method_name))
+
+    visitor = WrongImportVisitor(default_options, tree=tree)
+    visitor.run()
+
+    assert_errors(visitor, [VagueImportViolation])
--- a/tests/test_visitors/test_ast/test_imports/test_vague_imports.py
+++ b/tests/test_visitors/test_ast/test_imports/test_vague_imports.py
@@ -2,6 +2,7 @@
 
 import pytest
 
+from wemake_python_styleguide import constants
 from wemake_python_styleguide.violations.consistency import (
     VagueImportViolation,
 )
@@ -9,26 +10,14 @@ from wemake_python_styleguide.visitors.ast.imports import WrongImportVisitor
 
 import_template = 'from mod import {0}'
 
-blacklisted_method_names = [
-    'dump',
-    'dumps',
-    'load',
-    'loads',
-    'parse',
-    'safe_dump',
-    'safe_dump_all',
-    'load_all',
-    'dump_all',
-    'safe_load_all',
-    'safe_dump_all',
-]
+alias_import_template = 'from mod import something as {0}'
 
 
 @pytest.mark.parametrize(
     'method_name',
-    blacklisted_method_names + ['a', 'from_something', 'to_something'],
+    list(constants.VAGUE_IMPORTS_BLACKLIST) + ['a', 'from_thing', 'to_thing'],
 )
-def test_special_cases_vague_import(
+def test_vague_method_name_import(
     assert_errors,
     parse_ast_tree,
     method_name,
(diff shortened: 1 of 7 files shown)


wemake_python_styleguide/violations/consistency.py
```diff
@@ -1748,3 +1750,37 @@ class OperationSignNegationViolation(ASTViolation):
...
     code = 346
+
+
+@final
+class VagueImportViolation(ASTViolation):
+    """
+    Forbids imports that outside of the module may cause confusion.
+
+    Reasoning:
+        When you see datetime.* in the code, you know that it's from datetime.
```
sobolevn: We use double backquotes on code like `datetime.*`

wemake_python_styleguide/violations/consistency.py
```diff
@@ -1748,3 +1750,37 @@ class OperationSignNegationViolation(ASTViolation):
...
+        json.loads(content)
+
+        # Wrong:
+        from json import loads
+
+        ...
+
+        loads(content)
+
+    .. versionadded:: 0.1.0
```
sobolevn: Version added is `0.13.0`

wemake_python_styleguide/visitors/ast/imports.py
```diff
@@ -30,6 +31,20 @@
 class _ImportsValidator(object):
     """Utility class to separate logic from the visitor."""
 
+    _vague_imports_blacklist = {
```
sobolevn: We use `frozenset` and `ClassVar` type

tests/test_visitors/test_ast/test_imports/test_vague_imports.py
```diff
@@ -0,0 +1,43 @@
...
+    'safe_dump',
+    'safe_dump_all',
+    'load_all',
+    'dump_all',
+    'safe_load_all',
+    'safe_dump_all',
+]
+
+
+@pytest.mark.parametrize(
```
sobolevn: Can you please also test multiple imports like: `from some import parse, dumps`
And named imports like: `from some import dumps as json_dumps`

sobolevn:
> Implemented this using the original proposal from #814, opting for the broader scope of the check - all the imported method names are checked against blacklist, not only those relating to json or yaml.

Totally agree.

sobolevn:
By the way, maybe it is better to store the blacklist in `constants.py` as we do it for other similar rules? It would be also nice to link to this constant from the violation. See: https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/violations/oop.py#L64

tests/test_visitors/test_ast/test_imports/test_vague_imports.py
lensvol: Actually, it is an interesting question: should named imports of the vague names trigger a violation? IMO they should not, as long as alias conform to the same rules.

wemake_python_styleguide/visitors/ast/imports.py
lensvol: `frozenset` makes sense, but how does `ClassVar` play into this?

tests/test_visitors/test_ast/test_imports/test_vague_imports.py
sobolevn: > they should not

I agree 👍

orsinium: Yes, `from json import loads as json_loads` is silly, but is ok for this check :+1:

wemake_python_styleguide/visitors/ast/imports.py
```diff
@@ -30,6 +31,20 @@
...
+        'loads',
+        'dump',
+        'dumps',
+        'parse',
+        'safe_load',
+        'safe_dump',
+        'load_all',
+        'dump_all',
+        'safe_load_all',
+        'safe_dump_all',
```
orsinium: If you would do more changes in this PR, add, please, also `read` and `write`. Thank you :heart:

orsinium:
Thank you for your PR. Great work!

wemake_python_styleguide/visitors/ast/imports.py
sobolevn: I agree! 👍

wemake_python_styleguide/visitors/ast/imports.py
sobolevn: That's how we separate class-level variables from instance-level variables: `_vague_imports_blacklist: ClassVar[FrozenSet[str]]`. But, it does not really matter, because it should be moved to `constants.py` as described in one of my comments.

lensvol:
Requested changes implemented.

sobolevn:
That's so cool! 🎉

sobolevn:
Thanks a lot 👍

orsinium:
Beautiful :scream_cat:

PR #869, opened 2019-10-11
Added test to ensure parity between previous codes and versionchanged

### LINKED ISSUE #853: Test that violations with `previous_code` field have `.. versionchanged` directive
Some violations do change their codes once in a while. When this happens we do two things:
1. We set `previous_codes` class attribute with the value of the previous codes inside (might contain multiple values), so it would be easier for users to search the docs: https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/violations/oop.py#L91
2. We set `.. versionchanged::` doc attribute inside the docstring: https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/violations/oop.py#L85

But, we can forget to do the  the second step.
We need to test it!

For all violations that have `previous_codes` attribute we need to be sure that the docstring contains the same or greater amount of `.. versionchanged::` marker

Here are the tests for violation docs: https://github.com/wemake-services/wemake-python-styleguide/blob/master/tests/test_violations/test_docs.py

Ben-Drebing:
# I have made things!

## Checklist

## Related issues
- Closes #853


tests/test_violations/test_docs.py
```diff
@@ -24,6 +24,15 @@ def test_violation_name(all_violations):
         assert class_name.endswith('Violation'), class_name
 
 
+def test_violations_with_previous_codes_have_versionchanged(all_violations):
+    """Ensures that all violations with the previous codes attribute
+    also have the same or greater number of ..versionchanged markers"""
+    for violation in all_violations:
+        if hasattr(violation, 'previous_codes'):
```
sobolevn: We don't use `hasattr` in our code. We use `getattr`. In this case it would be:

```python
previous_codes = getattr(violation, 'previous_codes', None)
if previous_codes is not None:
    ...
```

tests/test_violations/test_docs.py
```diff
@@ -24,6 +24,15 @@ def test_violation_name(all_violations):
         assert class_name.endswith('Violation'), class_name
 
 
+def test_violations_with_previous_codes_have_versionchanged(all_violations):
```
sobolevn: Looks like this name is too long 😞

sobolevn:
Oh, and please add a changelog entry.

sobolevn:
@Ben-Drebing friendly ping.

sobolevn:
@Ben-Drebing sadly, but it is still failing. I will help you with the formatting.

sobolevn:
Thanks a lot!

Issue #788, opened 2019-08-23
Forbid incorrectly swapped variables


sobolevn:
# Rule request

## Thesis

We need to forbid code like:

```python
a = b
b = a
```

And:

```python
temp = a
a = b
b = temp
```

The first one indicates a failed attempt to swap variables.
The second one should be rewritten to: `a, b = b, a`

https://rust-lang.github.io/rust-clippy/master/index.html#almost_swapped

milg15:
I might give it a try later

sobolevn:
Please, feel free to send a PR in case you will have the time! 👍

sobolevn:
Closed by https://github.com/wemake-services/wemake-python-styleguide/pull/877

PR #877, opened 2019-10-12
Forbid incorrectly swapped variables #788

PrVrSs:
# I have made things!

## Checklist

### DIFF
--- a/wemake_python_styleguide/logic/collections.py
+++ b/wemake_python_styleguide/logic/collections.py
@@ -2,7 +2,20 @@
 
 import ast
 from functools import partial
-from typing import Any, Iterable, List, Sequence, Tuple, Type
+from typing import (
+    Iterable,
+    List,
+    Optional,
+    Sequence,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+    cast,
+)
+
+_NodeType = TypeVar('_NodeType')
+_DefaultType = TypeVar('_DefaultType')
 
 
 def normalize_dict_elements(node: ast.Dict) -> Sequence[ast.AST]:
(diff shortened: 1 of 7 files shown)


wemake_python_styleguide/violations/best_practices.py
```diff
@@ -1851,3 +1853,37 @@ class ApproximateConstantViolation(ASTViolation):
...
     code = 447
     error_template = 'Found approximate constant: {0}'
+
+
+@final
+class AlmostSwappedViolation(ASTViolation):
+    """
+    Forbids unpythonic swap variables.
+
+    We check for a = b; b = a sequences.
```
sobolevn: We use double backquotes to highlight the code. Please, add it to `a = b; b = a`

wemake_python_styleguide/visitors/ast/statements.py
```diff
@@ -149,9 +154,56 @@ def visit_statement_with_body(self, node: StatementWithBody) -> None:
...
         self._check_useless_node(node, node.body)
         self.generic_visit(node)
 
+    def _check_swapped_variables(
+        self,
+        body: Sequence[ast.stmt],
+    ) -> None:
+        for assigns in sequence_of_node((ast.Assign,), body):
+            self._almost_swapped(
+                cast(Sequence[ast.Assign], assigns),
```
sobolevn: Do we need a `cast` here? Why can't we return the correct type from `sequence_of_node`?

wemake_python_styleguide/visitors/ast/statements.py
```diff
@@ -149,9 +154,56 @@ def visit_statement_with_body(self, node: StatementWithBody) -> None:
...
+            )
+
+    def _almost_swapped(self, assigns: Sequence[ast.Assign]) -> None:
```
sobolevn: This method looks very complex, our linter is also marking it like so. Can you please refactor it a bit?

wemake_python_styleguide/violations/best_practices.py
PrVrSs: done

wemake_python_styleguide/visitors/ast/statements.py
PrVrSs: done

wemake_python_styleguide/visitors/ast/statements.py
PrVrSs: probably done

wemake_python_styleguide/logic/collections.py
```diff
@@ -33,30 +46,33 @@ def normalize_dict_elements(node: ast.Dict) -> Sequence[ast.AST]:
 
 
 def sequence_of_node(
-    node_type: Tuple[Type[ast.stmt]],
+    node_types: Tuple[Type[_NodeType]],
```
sobolevn: It should be `Tuple[Type[_NodeType], ...]`

PrVrSs: done

PrVrSs:
теперь правда, всегда будет по-разному: `a, b = b, a` `|`  `b, a = a, b`. Не могу сказать насколько это критично

sobolevn:
@PrVrSs can we just drop this formatting part? It does not bring much value (because we can use `show-source = true`). Will it make the source code easier? What do you think?

sobolevn:
Thanks a lot for your work. I really appreciate that. 👍

PR #883, opened 2019-10-13
Forbids redundant subscripts

### LINKED ISSUE #880: Forbid `[0:x]` and `[1:None]`, prefer `[:x]` and `[1:]`
# Rule request

## Thesis

We need to enforce consistency in several cases:
- `[0:x]` which copies all elements from the first one to `x`. We need to use `[:x]` instead. Because it is way more readable.
- `[1:None]` which copies all elements from the second to the very last one. We need to use `[1:]` instead. It is also more readable.

So, the rule is: we must omit the starting `0` and the leading `None`. Other values are fine.
Step can also be specified:  `[:0:2]` and `[1::2]`

## Reasoning

We do it for consistency reasons.

cronos2:
# I have made things!

## Checklist

## Related issues

Fixes #880


tests/test_visitors/test_ast/test_subscripts/test_redundant_subscripts.py
```diff
@@ -0,0 +1,62 @@
...
+from wemake_python_styleguide.violations.consistency import (
+    RedundantSubscriptViolation,
+)
+from wemake_python_styleguide.visitors.ast.subscripts import SubscriptVisitor
+
+usage_template = 'constant{0}'
+
+
+@pytest.mark.parametrize('expression', [
+    '[0:7]',
```
sobolevn: We need to test `[0:7:1]`

tests/test_visitors/test_ast/test_subscripts/test_redundant_subscripts.py
```diff
@@ -0,0 +1,62 @@
...
+    '[None:7]',
+    '[3:None]',
```
sobolevn: `[3:None:2]`

wemake_python_styleguide/violations/consistency.py
```diff
@@ -1784,3 +1786,52 @@ class VagueImportViolation(ASTViolation):
...
+class RedundantSubscriptViolation(ASTViolation):
+    """
+    Forbids the use of redundant components in a subscript's slice.
+
+    Reasoning:
+        We do it for consistency reasons.
+
+    Example::
+
+        # Correct:
```
sobolevn: Just two examples would be fine:

```
array[:10]
array[1:]
```

And the same for `Wrong` section:

```
array[0:10]
array[1:None]
```

wemake_python_styleguide/visitors/ast/subscripts.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+
+        if not (lower_ok and upper_ok and step_ok):
+            self.add_violation(
+                consistency.RedundantSubscriptViolation(
+                    node, text=str(node),
+                ),
+            )
+
+    def _is_None(self, component_value: ast.expr) -> bool:
```
sobolevn: `_is_none`

wemake_python_styleguide/violations/consistency.py
```diff
@@ -1799,23 +1799,11 @@ class RedundantSubscriptViolation(ASTViolation):
     Example::
```
sobolevn: Can you please substitute the whole example section with:

```
    Example::

        # Correct:
        array[:7]
        array[3:]

        # Wrong:
        x[0:7]
        x[3:None]
```

tests/fixtures/noqa.py
```diff
@@ -582,3 +582,7 @@ def nested_annotation_return_func() -> Union[str, Union[int, float]]: # noqa: WP
 from json import loads  # noqa: WPS347
 from some_module import a  # noqa: WPS347
 from text import from_file  # noqa: WPS347
+
+constant[0:7]  # noqa: WPS348,WPS428
```
sobolevn: Let's replace it with `print(constant[0:7])  # noqa: WPS348`

Why? We don't need to test all kind of things here. We only test that `noqa` works fine. So, one example is enough.

wemake_python_styleguide/violations/consistency.py
cronos2: Oops, that was my bad. Will do.

tests/fixtures/noqa.py
cronos2: You mean just testing that `# noqa` works for the `[0:7]` case? Or testing the three cases but add `print` to avoid WPS428?

sobolevn: Yes, we need to test that a single `noqa` works for this violation. We also try to avoid unrelated violations. So, a single `print(constant[0:7]) # noqa: WPS348` is enough.

cronos2:
Does Travis also show "Pending" to you? The build was actually successful but it seems GitHub has not picked that up.

wemake_python_styleguide/violations/consistency.py
sobolevn: Looks like the example is still not fix. I'll help you with this one.

sobolevn:
Thanks a lot, @cronos2! That's a very useful rule.

PR #914, opened 2019-10-17
Fix #902

### LINKED ISSUE #902: Allow `super()` as overused expression
We need to add a new predicate to allow `super()` calls here: https://github.com/wemake-services/wemake-python-styleguide/blob/39a3caeb9883c3112d878a88038e8b7331389265/wemake_python_styleguide/visitors/ast/complexity/overuses.py#L114

It is obviously allowed.

Then we need to add a unit test here: https://github.com/wemake-services/wemake-python-styleguide/blob/ed1afeb3366525f0b3f1adc346a189a3c0f3b923/tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_expressions.py

That's it. It is beginner friendly.

hhsdev:
# I have made things!
Added a special case in the rule forbidding overused expressions (`WPS204`) to ignore `super()` calls.

## Checklist

## Related issues
- Closes #902

### DIFF
--- a/tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_expressions.py
+++ b/tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_expressions.py
@@ -184,9 +184,10 @@ def test_class_expression_use(
     method_context1,
     method_context2,
     method_context3,
+    module_context,
 ])
 @pytest.mark.parametrize('expression', ignored_expressions)
-def test_func_ignored_expressions(
+def test_ignored_expressions(
     assert_errors,
     parse_ast_tree,
     options,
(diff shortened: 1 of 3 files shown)


tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_expressions.py
```diff
@@ -170,3 +174,52 @@ def test_class_expression_use(
...
+
+    option_values = options(max_function_expressions=1)
+    visitor = ExpressionOveruseVisitor(option_values, tree=tree)
+    visitor.run()
+
+    assert_errors(visitor, [])
+
+
+@pytest.mark.parametrize('code', [
+    module_context,
```
sobolevn: Why this is a separate case? Why not using `test_func_ignored_expressions`?

wemake_python_styleguide/visitors/ast/complexity/overuses.py
```diff
@@ -186,3 +187,10 @@ def _post_visit(self) -> None:
 
 def _is_class_context(node: ast.AST) -> bool:
     return isinstance(nodes.get_context(node), ast.ClassDef)
+
+
+def _is_super_call(node: ast.AST) -> bool:
+    if isinstance(node, ast.Call) and isinstance(node.func, ast.Name):
+        if node.func.id == 'super':
```
sobolevn: `return node.func.id == 'super'`

tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_expressions.py
hhsdev: Ah, I just copied the tests from above and they had different functions for the two cases. Digging through the code, I now understand that it is unnecessary. Will fix it.

sobolevn:
Thanks a lot!

Issue #908, opened 2019-10-16
`super()` with other methods and properties


sobolevn:
# Rule request

## Thesis

An interesting idea I got this evening. I usually see code like this:

```python
class Child(Parent):
    def some(self):
        original = super().some()
        ...
```

And it is perfectly fine. That's why we have `super()` in the first place. But, I also saw this code today:

```python
class Child(Parent):
    def some(self):
        other = super().other()
        ...
```

Wait, what? Why `Child.some()` uses `super().other()` and not `self.other()`?
I think that it is important to enforce the following rule: one can only use `super()` method that matches the following context. Like: `super().some()` and `super().some` in `Child.some()`, and `super().prop` and `super().prop()` in `Child.prop` (if we are dealing with properties).

## Reasoning

It enforces better OOP design. And better responsibility distribution.

cmppoon:
I want to help with this one. Can I take this?

sobolevn:
Sure, thanks a lot! @cmppoon please, feel free to ask any questions you have.

sobolevn:
Closed by #922

PR #922, opened 2019-10-18
Forbids `super()` with other methods and properties #908

cmppoon:
# I have made things!

## Checklist

## Related issues
- Ref #902 

I'm not sure if this is the correct approach. If something needs to be changed, please let me know.

### DIFF
--- a/tests/fixtures/noqa.py
+++ b/tests/fixtures/noqa.py
@@ -594,14 +594,6 @@ print(constant[0:7])  # noqa: WPS349
 
 var_a = var_a + var_b  # noqa: WPS350
 
-class ParentClass(object):
-    @property
-    def some_var(self):
-        return self._some_var
-
-    def some_method(self):
-        return ''
-
 class ChildClass(ParentClass):
     @property
     def some_var(self):
--- a/tests/fixtures/noqa.py
+++ b/tests/fixtures/noqa.py
@@ -593,3 +593,19 @@ swap_b = swap_a  # noqa: WPS448
 print(constant[0:7])  # noqa: WPS349
 
 var_a = var_a + var_b  # noqa: WPS350
+
+class ParentClass(object):
+    @property
+    def some_var(self):
+        return self._some_var
+
+    def some_method(self):
+        return ''
+
+class ChildClass(ParentClass):
+    @property
+    def some_var(self):
+        return super().other_var # noqa: WPS613
+
+    def some_method(self):
+        super().some_other_method() # noqa: WPS613
--- a/tests/test_visitors/test_ast/test_functions/test_super_call.py
+++ b/tests/test_visitors/test_ast/test_functions/test_super_call.py
@@ -2,7 +2,10 @@
 
 import pytest
 
-from wemake_python_styleguide.violations.oop import WrongSuperCallViolation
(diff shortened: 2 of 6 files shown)


tests/fixtures/noqa.py
```diff
@@ -593,3 +593,19 @@ def nested_annotation_return_func() -> Union[str, Union[int, float]]: # noqa: WP
 print(constant[0:7])  # noqa: WPS349
 
 var_a = var_a + var_b  # noqa: WPS350
+
+class ParentClass(object):
```
sobolevn: Can you please remove this class? It works fine, and in `test_noqa` we only test that `# noqa` comments work. So, there's no need in correct examples. We only care about incorrect ones.

Which is already present.

tests/test_visitors/test_ast/test_functions/test_super_call.py
```diff
@@ -56,9 +73,24 @@ def some_method(self, arg1):
...
+class Example(Parent):
+    def some_method(self):
+        super().other_method()
+"""
+
+super_call_with_wrong_property = """
+class Example(Parent):
+    @property
+    def x(self):
+        return super().y
```
sobolevn: Can we please also test:

- `super().prop.nested`
- `super().prop.method()`
- `super().prop['key']`

We also need to cover all the similar cases for both:
- `super()`
- `super(Class, self)`

I suggest to parametrise your examples like so:

```
super_call_with_method = """
class Example(Parent):
    def some_method(self):
        super({0}).{1}
"""
```

And in tests use two fixtures: `''` and `Class, self` for the first (`{0}`) indexed param. 
And create a second fixture that will insert different properties:

- `.prop`
- `.method()`
- `.prop.nested`
- etc

This way we will cover everything.

sobolevn: The same applies for both correct and wrong examples.

wemake_python_styleguide/violations/oop.py
```diff
@@ -536,3 +538,39 @@ def method(self, argument):
...
+
+
+@final
+class WrongSuperCallContextViolation(ASTViolation):
+    """
+    Forbids to use ``super()`` with incorrect methods or properties access.
+
+    Reasoning:
+        Can only use ``super()`` method that matches the following context.
+        super().some() and super().some in Child.some(),
```
sobolevn: We use double backquotes to highlight text.

wemake_python_styleguide/violations/oop.py
```diff
@@ -536,3 +538,39 @@ def method(self, argument):
...
+
+    Solution:
+        Use ``super()`` methods and properties with the correct context.
+
+    Example::
+
+        # Correct:
+        class Child(Parent):
+            def some_method(self):
+            original = super().some_method()
```
sobolevn: Indentation is incorrect. Please, also remove `...`

wemake_python_styleguide/violations/oop.py
```diff
@@ -536,3 +538,39 @@ def method(self, argument):
...
+            ...
+
+        # Wrong:
+        class Child(Parent):
+            def some_method(self):
+            other = super().other_method()
```
sobolevn: The same here.

PR #874, opened 2019-10-11
issue-792: Forbid Union[T, None]

novikovfred:
# I have made things!

## Checklist

- [x] I have double checked that there are no unrelated changes in this pull request (old patches, accidental config files, etc)
- [x] I have created at least one test case for the changes I have made
- [x] I have updated the documentation for the changes I have made
- [x] I have added my changes to the `CHANGELOG.md`

## Related issues

- Closes #792

PR #924, opened 2019-10-19
Issue 875

### LINKED ISSUE #875: Forbid `Optional[Union[A, B]]`, use `Union[A, B, None]` instead
# Rule request

## Thesis

These two types are equal: `Optional[Union[A, B]] == Union[A, B, None]`

For consistency reasons we need to enforce the second form over the first one.
Related to #792

ko3luhbka:
# I have made things!

## Checklist

## Related issues

- Closes #875

I think there's no need to create new visitor class for this violation and the new logic fits well into 
existing ``SemanticAnnotationVisitor`` class.
Also I'm not shure if I chose correct violation code because of the active pull request #874 which uses 'WPS703'.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ We used to have incremental versioning before `0.1.0`.
 - Forbids to use redundant subscripts (e.g., `[0:7]` or `[3:None]`)
 - Allow `super()` as a valid overused expression
 - `WPS350` enforce using augmented assign pattern
+- Forbids to use `Optional[Union` in function annotations 
 
 ### Bugfixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ We used to have incremental versioning before `0.1.0`.
 - Forbids to use redundant subscripts (e.g., `[0:7]` or `[3:None]`)
 - Allow `super()` as a valid overused expression
 - `WPS350` enforce using augmented assign pattern
-- Forbids to use `Optional[Union` in function annotations 
+- Forbids to use `Optional[Union]` in function annotations 
 
 ### Bugfixes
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ We used to have incremental versioning before `0.1.0`.
 - Forbids to use redundant subscripts (e.g., `[0:7]` or `[3:None]`)
 - Allow `super()` as a valid overused expression
 - `WPS350` enforce using augmented assign pattern
+- Forbids to use `Optional[Union` in function annotations
```
sobolevn: The example is cut, please fix it

tests/fixtures/noqa.py
```diff
@@ -573,14 +573,21 @@ def literal_none_return_func() -> Literal[None]:  # noqa: WPS701
...
-def nested_annotation_return_func() -> Union[str, Union[int, float]]: # noqa: WPS702
+def nested_annotation_return_func() -> Union[str, Union[int, float]]:  # noqa: WPS702
     """Union[str, Union[int, float]]"""
 
 
+def optional_union_annotation_func(arg: Optional[Union[str, int]]):  # noqa: WPS704
+    """Optional[Union[str, int]]"""
+
+
+def optional_union_annotation_return_func() -> Optional[Union[str, int]]:  # noqa: WPS704
```
sobolevn: One example is enough.

ko3luhbka: @sobolevn why we don't want to check return annotations here? Is it because they are checked in ``test_flat_types.py``? And it's actually checked here for ``Literal[None]`` annotation.

sobolevn: Yes, you are correct. We only test here that `noqa: ` comments work. That's it.
Other checks are in `test_flat_types.py`

You can delete this incorrect example with two `Literal` as well. I have just missed it during the review process. Now it causes confusion

sobolevn:
Ok, I will actually fix it myself. I need a deep look at this problem.

Thanks!

PR #926, opened 2019-10-19
Forbid unnecessary literals #780

### LINKED ISSUE #780: Unnecessary literals
# Rule request

## Thesis

Bad:

```python
int()
float()
str()
bytes()
bool()
complex()
```

Good:

```python
0
0.0
''
b''
False
0j
```

[flake8-comprehensions](https://github.com/adamchainz/flake8-comprehensions) does the same, but only for dict, list, and tuple.

## Reasoning

> There should be one-- and preferably only one --obvious way to do it.

PrVrSs:
# I have made things!

## Checklist

## Related issues

Closes #780

### DIFF
--- a/wemake_python_styleguide/violations/consistency.py
+++ b/wemake_python_styleguide/violations/consistency.py
@@ -76,6 +76,7 @@ Summary
    AdditionAssignmentOnListViolation
    RedundantSubscriptViolation
    AugmentedAssignPatternViolation
+   UnnecessaryLiteralsViolation
 
 Consistency checks
 ------------------
@@ -131,6 +132,7 @@ Consistency checks
 .. autoclass:: AdditionAssignmentOnListViolation
 .. autoclass:: RedundantSubscriptViolation
 .. autoclass:: AugmentedAssignPatternViolation
+.. autoclass:: UnnecessaryLiteralsViolation
 
 """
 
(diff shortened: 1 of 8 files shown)


wemake_python_styleguide/violations/consistency.py
```diff
@@ -1866,3 +1868,37 @@ class AugmentedAssignPatternViolation(ASTViolation):
...
+    """
+    Forbids the use of unnecessary literals in your code.
+
+    Reasoning:
+        There should be one-- and preferably only one --obvious way to do it.
+
+    Example::
+
+        # Correct:
+        0
```
sobolevn: Let's make it `default = 0`

sobolevn: And remove other values

wemake_python_styleguide/violations/consistency.py
```diff
@@ -1866,3 +1868,37 @@ class AugmentedAssignPatternViolation(ASTViolation):
...
+        0.0
+        ''
+        b''
+        False
+        0j
+
+        # Wrong:
+        int()
```
sobolevn: Let's make it `default = int()` and remove other values.

wemake_python_styleguide/violations/consistency.py
```diff
@@ -1866,3 +1868,37 @@ class AugmentedAssignPatternViolation(ASTViolation):
...
     error_template = 'Found usable augmented assign pattern'
     code = 350
+
+
+@final
+class UnnecessaryLiteralsViolation(ASTViolation):
+    """
+    Forbids the use of unnecessary literals in your code.
+
+    Reasoning:
```
sobolevn: The reasoning is technically correct, but is not complete.
I guess that it should be formulated like so: `We discourage using primitive calls to get default type values. There are better ways to get these values.`

We are also missing a solution here.
The solution should be: `Use direct default values of the given type`

Issue #927, opened 2019-10-20
False positive for `WPS413`


sobolevn:
# Bug report

## What's wrong

This code produces incorrect results:

```python
class _Callable(object):
    # Callable[[CenterProtocol[T]], T]
    def __getattr__(self, key: str) -> Callable[[object], T1]:
        return operator.attrgetter(key)
```

Output:

```
  32:5     WPS413 Found bad magic module function: __getattr__
  def __getattr__(self, key: str) -> Callable[[object], T1]:
```

## How is that should be

It should work correctly: it is not a module level ` __getattr__`, it is a class-level method.

## System information

I am using `0.12.5`

PR #929, opened 2019-10-20
Fixed false positive in WPS413

cmppoon:
# I have made things!

## Checklist

## Related issues

- Refs #927 

I found incorrect blacklist __getattr__ in MAGIC_MODULE_NAMES_BLACKLIST instead of __getaddr__, and I updated noqa.py accordingly.

### DIFF
--- a/tests/test_visitors/test_ast/test_modules/test_magic_module_functions.py
+++ b/tests/test_visitors/test_ast/test_modules/test_magic_module_functions.py
@@ -10,21 +10,32 @@ from wemake_python_styleguide.visitors.ast.modules import (
     MagicModuleFunctionsVisitor,
 )
 
-regular_function = """
+module_level_method = """
 def {0}(name):
     ...
 """
 
+class_level_method = """
+class Example(object):
+    def {0}(name):
+        ...
+"""
 
+
+@pytest.mark.parametrize('code', [
+    module_level_method,
+    class_level_method,
+])
 @pytest.mark.parametrize('function_names', MAGIC_MODULE_NAMES_BLACKLIST)
-def test_wrong_magic_used(
+def test_right_magic_used(
     assert_errors,
+    code,
     parse_ast_tree,
     options,
     function_names,
 ):
     """Testing magic methods with logic is restricted."""
-    tree = parse_ast_tree(regular_function.format(function_names))
+    tree = parse_ast_tree(code.format(function_names))
     option_values = options(i_control_code=False)
 
     visitor = MagicModuleFunctionsVisitor(option_values, tree=tree)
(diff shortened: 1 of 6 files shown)


wemake_python_styleguide/constants.py
```diff
@@ -281,7 +281,7 @@
 
 #: List of bad magic module functions.
 MAGIC_MODULE_NAMES_BLACKLIST: Final = frozenset((
-    '__getattr__',
+    '__getaddr__',
```
sobolevn: Sorry, I don't understand what are you doing.

Maybe I can help you with some explanations?

cmppoon: I read the comment of ``BadMagicModuleFunctionViolation`` and it says 

> Forbids to use ``__getaddr__`` and ``__dir__`` module magic methods. 

So I looked into the implementation and found that it checks if ``node.name`` is in ``constants.MAGIC_MODULE_NAMES_BLACKLIST``, 

But in ``MAGIC_MODULE_NAMES_BLACKLIST``, ``__getattr__`` was in the blacklist, not ``__getaddr__``. So, I thought that this was the cause of the problem.  Maybe I misunderstood the task.

It would be great if you can provide some explanations about how to solve this problem.

sobolevn: Oh, sorry. `__getaddr__` is a typo. We only work with `__getattr__`. Can you please submit a fix for it?

The problem I have described in this particular issue is that we don't correctly check where `__getattr__` is located. Here: https://github.com/wemake-services/wemake-python-styleguide/blob/d01851a23366aecf4f581f0f7032ba3511d85bdd/wemake_python_styleguide/visitors/ast/modules.py#L87

We need to check that `if isinstance(get_context(node), ast.Module):` and write a new regression test here: https://github.com/wemake-services/wemake-python-styleguide/blob/d01851a23366aecf4f581f0f7032ba3511d85bdd/tests/test_visitors/test_ast/test_modules/test_magic_module_functions.py

tests/test_visitors/test_ast/test_modules/test_magic_module_functions.py
```diff
@@ -10,21 +10,32 @@
...
+"""
+
 
+@pytest.mark.parametrize('code', [
+    module_level_method,
+    class_level_method,
+])
 @pytest.mark.parametrize('function_names', MAGIC_MODULE_NAMES_BLACKLIST)
-def test_wrong_magic_used(
+def test_right_magic_used(
```
cmppoon: I found another typo here (it should be test_right_magic_used since it tests correct cases)

tests/test_visitors/test_ast/test_modules/test_magic_module_functions.py
```diff
@@ -33,15 +44,19 @@ def test_wrong_magic_used(
     assert_errors(visitor, [])
 
 
+@pytest.mark.parametrize('code', [
+    module_level_method,
+])
 @pytest.mark.parametrize('function_names', MAGIC_MODULE_NAMES_BLACKLIST)
-def test_right_magic_used(
+def test_wrong_magic_used(
```
cmppoon: Same here.

PR #930, opened 2019-10-21
Dissalow wrong annotations in assignments

### LINKED ISSUE #925: Type annotation check doesn't work for type aliases
# Bug report

## What's wrong
Type annotation check doesn't work if type aliases are used. For example, this code
```
type_alias = Literal[None]

def my_func(arg: type_alias):
    pass
```
passes flake8 checks without raising `WPS701` error, while this code
```
def my_func(arg: Literal[None]):
    pass
```
makes flake8 complain: "WPS701 Found useless `Literal[None]` typing annotation".
## How is that should be
Type annotation errors should be raised whether type aliases are used or not.

## System information
wemake-python-styleguide: 0.12.5

AlwxSin:
# I have made things!

## Checklist

## Related issues
closes #925

As a side effect, fixed several wrong multiline annotations.


wemake_python_styleguide/violations/annotations.py
```diff
@@ -62,7 +64,7 @@ def func(empty: Literal[None]):
     """
 
     code = 700
-    error_template = 'Found useless `Literal[None]` typing annotation'
```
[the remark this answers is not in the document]
sobolevn: You can remove the word `typing` 🙂

wemake_python_styleguide/violations/annotations.py
```diff
@@ -87,6 +89,7 @@ class NestedAnnotationsViolation(ASTViolation):
         # Wrong:
         Literal[Literal[Literal[1, 2, 3], "foo"], 5, None]
         Union[Union[int, str], float]
+        NONE_ALIAS = Union[Union[int, str]
```
sobolevn: I guess the existing example is good enough. It is about types, not their usage.

wemake_python_styleguide/violations/annotations.py
```diff
@@ -115,6 +118,7 @@ class UnionNestedInOptionalViolation(ASTViolation):
 
         # Wrong:
         Optional[Union[int, str]]
+        NONE_ALIAS = Optional[Union[int, str]]
```
sobolevn: The same here.

wemake_python_styleguide/visitors/ast/annotations.py
```diff
@@ -31,6 +31,11 @@ class _GenericAnnotationVisitor(BaseNodeVisitor):
         'typing_extensions.',
     ))
 
+    def visit_Assign(self, node: ast.Assign) -> None:
+        """Checks assignment patterns."""
+        self._check_annotation(node)
```
sobolevn: Does it cover `x = y = Literal[None]` and `x, y = Literal[None], Literal[None]`?
We need to test it.

AlwxSin: It should. I'll add more tests.

wemake_python_styleguide/violations/annotations.py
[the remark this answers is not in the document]
AlwxSin: After a night I think that I need to exclude strings from checking completely. Or this code will be invalid too
```python
def get_example_of_wrong_annotation:
    wrong1 = 'Literal[None]'
    wrong2 = Optional[Union[str, float]]
    return (wrong1, wrong2)
```

wemake_python_styleguide/visitors/ast/annotations.py
AlwxSin: Added more tests.

wemake_python_styleguide/violations/annotations.py
AlwxSin: Removed.

wemake_python_styleguide/violations/annotations.py
AlwxSin: Removed.

wemake_python_styleguide/violations/annotations.py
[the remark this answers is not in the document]
AlwxSin: So, now I don't check if it is string in assignment.

tests/test_visitors/test_ast/test_annotations/test_annotation_metadata/conftest.py
```diff
@@ -7,6 +7,22 @@ def function(arg: {0}):
...
+function_arg_alias_template = """
+Type_Alias = {0}
+
+def function(arg: Type_Alias):
+    ...
+"""
+
+function_arg_multiple_alias_template = """
+a, Another_Alias = {0}, str
+b = c = {0}
```
sobolevn: Sorry, this does not look correct to me. 

We need to make three separate new templates for this case:
1. `a = {0}`
2. `b = c = {0}`
3. `d, e = {0}, {0}`

And nothing else, function is redundant here. 

I am also not sure about how `d, e = {0}, {0}` should behave. I guess, that it should raise two violations. However, it is not critical and one is good enough.
See which implementation is easier.

Thanks a lot for your time and effort! 👍

AlwxSin: Will do.

But with `d, e = Literal['a'], Literal['b']` I have a false positive `NestedAnnotationsViolation`. Not sure how to defeat it.

AlwxSin: Well, regex will always help 😁
Fixed.

sobolevn:
Thanks a lot!

Issue #892, opened 2019-10-14
Forbid incorrect exception order


sobolevn:
# Rule request

## Thesis

Currently we are allowed to write this code:

```python
    try:
        ...
    except Exception:
        ...
    except ValueError:
        ...
```

What happens here? `except ValueError` will never be reached, because all exception will be covered by `except Exception`.

But, we have a limitation here. Since, we don't know and cannot know any other exception types except the builtin ones, we only check builtin exceptions for now.

## Reasoning

This is a best practice, because this can lead to different bugs.

Original idea: https://github.com/thg-consulting/inspectortiger/blob/6b9736b22528493cd41d0bf92f9fdcdd6c7cd129/inspectortiger/plugins/misc.py#L98

Amyth07:
Can I take this?

sobolevn:
Yes, sure, @Amyth07! Thanks a lot

cmppoon:
Hi, 

Can you assign this to me?

sobolevn:
Done! Thanks a lot!

PR #935, opened 2019-10-23
Forbid incorrect exception order Issue#892

cmppoon:
# I have made things!

## Checklist

## Related issues

- Ref #892 

I encountered a problem where I can't have 8 methods in one class, so I refactored code a little bit. 

Not sure if more test cases should be added to cover more use cases though. If that's the case, please let me know. Thanks a lot.


tests/test_visitors/test_ast/test_exceptions/test_exception_order.py
```diff
@@ -0,0 +1,63 @@
...
+    visitor.run()
+
+    assert_errors(visitor, [])
+
+
+@pytest.mark.parametrize('code', [
+    exception_template,
+])
+@pytest.mark.parametrize('statements', [
+    ('Exception', 'ValueError'),
```
sobolevn: Let's add more tests: 
- `('Exception', 'KeyError')`, 
- `('LookupError', 'IndexError')`, 
- `('BaseException', 'Exception')`

Let's also add some extra templates:

```python
try:
    ...
except (MyCustomError, {0}):
    ...
except (MyCustomError, {1}):
    ...
```

```python
try:
    ...
except CustomError:
    ...
except {0}:
    ...
except OtherCustomError:
    ...
except {1}:
    ...
```

tests/test_visitors/test_ast/test_exceptions/test_exception_order.py
```diff
@@ -0,0 +1,63 @@
...
+except {1}:
+    ...
+"""
+
+
+@pytest.mark.parametrize('code', [
+    exception_template,
+])
+@pytest.mark.parametrize('statements', [
+    ('ValueError', 'Exception'),
```
sobolevn: Let's also add two more cases:

- `('Exception', 'MyValueError'),`
- `('MyCustomException', 'MyValueError'),`

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -1855,3 +1857,41 @@ class StringConstantRedefinedViolation(ASTViolation):
...
+    Forbids the use incorrect order of ``except``.
+
+    Reasoning:
+        Using ``except Exception`` in incorrect order is error-prone.
+        It will make other ``except`` unreachable.
+
+    Solution:
+        Use ``except Exception`` after other ``except`` blocks.
+
+    Example::
```
sobolevn: Let's add `See also:` with https://docs.python.org/3/library/exceptions.html#exception-hierarchy

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -1855,3 +1857,41 @@ class StringConstantRedefinedViolation(ASTViolation):
...
+class IncorrectExceptOrderViolation(ASTViolation):
+    """
+    Forbids the use incorrect order of ``except``.
+
+    Reasoning:
+        Using ``except Exception`` in incorrect order is error-prone.
+        It will make other ``except`` unreachable.
+
+    Solution:
+        Use ``except Exception`` after other ``except`` blocks.
```
sobolevn: The solution should be: `Use correct order of exceptions.`.

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -1855,3 +1857,41 @@ class StringConstantRedefinedViolation(ASTViolation):
...
     code = 447
+
+
+@final
+class IncorrectExceptOrderViolation(ASTViolation):
+    """
+    Forbids the use incorrect order of ``except``.
+
+    Reasoning:
+        Using ``except Exception`` in incorrect order is error-prone.
```
sobolevn: Reasoning should be: `Using incorrect order of exceptions is error-prone, since you end up with some unreachable exception clauses.`

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -1855,3 +1857,41 @@ class StringConstantRedefinedViolation(ASTViolation):
...
+        It will make other ``except`` unreachable.
+
+    Solution:
+        Use ``except Exception`` after other ``except`` blocks.
+
+    Example::
+
```
sobolevn: We also need to add a note here that we only care about builtin exceptions

cmppoon:
Hi,

I suppose that I have to traverse the exception tree to get all built-in exceptions. I looked into the original idea repo and found the code that does the job. 

https://github.com/thg-consulting/inspectortiger/blob/6b9736b22528493cd41d0bf92f9fdcdd6c7cd129/inspectortiger/plugins/misc.py#L66 

Can I do exactly the same thing?

I've been a bit busy lately. Hopefully, I can finish this up in the next couple days . 😃

sobolevn:
Yes, sure! Just provide a link back to the original project.

wemake_python_styleguide/visitors/ast/exceptions.py
```diff
@@ -47,16 +49,45 @@ def _find_returing_nodes(
...
+            base.__name__
+            for base in getmro(exc)
+            if (
+                issubclass(base, BaseException) and
+                base.__name__ != exc.__name__
+            )
+        )
+        _traverse_exception(exc, builtin_exceptions)
+
+    return builtin_exceptions.copy()
```
sobolevn: Why do we need to `.copy` it?

wemake_python_styleguide/visitors/ast/exceptions.py
```diff
@@ -47,16 +49,45 @@ def _find_returing_nodes(
...
 
 
+def _traverse_exception(
+    cls,
+    builtin_exceptions=None,
+) -> Dict[str, Tuple[str]]:
+    """
+    Orinigal code: see below.
+
+    https://github.com/thg-consulting/inspectortiger/blob/
```
sobolevn: Can you please use bit.ly to shorten this link?

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -1855,3 +1857,46 @@ class StringConstantRedefinedViolation(ASTViolation):
...
+        # Wrong:
+        try:
+            ...
+        except Exception:
+            ...
+        except ValueError:
+            ...
+
+    See also:
+        https://docs.python.org/3/library/exceptions.html#exception-hierarchy
```
sobolevn: Can you please use bit.ly to shorten this link?

wemake_python_styleguide/visitors/ast/exceptions.py
```diff
@@ -129,22 +130,19 @@ def _check_return_path(self, node: ast.Try) -> None:
         if else_has and try_has:
             self.add_violation(TryExceptMultipleReturnPathViolation(node))
 
-    def _check_useless_except(self, node: ast.ExceptHandler) -> None:
-        if len(node.body) != 1:
-            return
+    def _check_exception_order(self, node: ast.Try) -> None:
+        built_in_exceptions = _traverse_exception(BaseException)
+        exceptions_list: List[str] = exceptions.get_all_exception_names(node)
```
sobolevn: Why do we need this annotation?

wemake_python_styleguide/visitors/ast/exceptions.py
```diff
@@ -47,16 +49,45 @@ def _find_returing_nodes(
     return try_has, except_has, else_has, finally_has
 
 
+def _traverse_exception(
```
sobolevn: This should be moved to `logic/`

wemake_python_styleguide/visitors/ast/exceptions.py
cmppoon: Apparently, we don't need it. I'm gonna fix it now.

wemake_python_styleguide/visitors/ast/exceptions.py
cmppoon: Same, seems like we don't need it.

wemake_python_styleguide/visitors/ast/exceptions.py
```diff
@@ -51,12 +52,12 @@ def _find_returing_nodes(
 class WrongTryExceptVisitor(BaseNodeVisitor):
     """Responsible for examining ``try`` and friends."""
 
-    _base_exception: ClassVar[str] = 'BaseException'
     _bad_returning_nodes: ClassVar[AnyNodes] = (
         ast.Return,
         ast.Raise,
         ast.Break,
     )
+    _except_exception: ClassVar[str] = 'Exception'
```
sobolevn: Why do we need this line?

wemake_python_styleguide/visitors/ast/exceptions.py
```diff
@@ -66,53 +67,23 @@ def visit_Try(self, node: ast.Try) -> None:
...
-            if isinstance(exc_handler.type, ast.Name):
-                exceptions.append(source.node_to_string(exc_handler.type))
-            elif isinstance(exc_handler.type, ast.Tuple):
-                exceptions.extend([
-                    source.node_to_string(node)
-                    for node in exc_handler.type.elts
-                ])
-
-        for exc_name, count in Counter(exceptions).items():
+        exceptions_list: List[str] = exceptions.get_all_exception_names(node)
```
sobolevn: Why do we need an annotation here? Is the `exceptions.get_all_exception_names` return type is not enough?

cmppoon: I think I just forgot to change/remove these two lines. Thanks for pointing that out. I will submit a fix now.

sobolevn:
Thank you so much! Awesome new rule!

PR #941, opened 2019-10-26
#938: fix WPS324 false-positive for nested functions

### LINKED ISSUE #938: WPS324 false-positive with nested functions
# Bug report

## What's wrong

This code produces false positive `WPS324`:

```python
def test():
    def decorator():
        return
    return decorator
```

## How is that should be

These two returns are completely different cases. We should check `.wps_context` of the node.
See: 
- https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/visitors/ast/keywords.py#L133
- https://github.com/wemake-services/wemake-python-styleguide/blob/b43e738c4527219f03c2eec420fe2e20cfd4c8cc/wemake_python_styleguide/logic/nodes.py#L29

ulturt:
# I have made things!

I have fixed InconsistentReturnViolation false positive for nested  functions

## Checklist

## Related issues

- Closes #938

### DIFF
--- a/tests/test_visitors/test_ast/test_keywords/test_consistency_returning/test_consistency_return.py
+++ b/tests/test_visitors/test_ast/test_keywords/test_consistency_returning/test_consistency_return.py
@@ -97,6 +97,13 @@ def function():
     return
 """
 
+wrong_example4 = """
+def function():
+    def decorator():
+        return
+    return decorator
+"""
+
 double_wrong_return1 = """
 def function():
     if some:
@@ -117,6 +124,7 @@ def function():
     wrong_example1,
     wrong_example2,
     wrong_example3,
+    wrong_example4,
 ])
 def test_wrong_return_statement(
     assert_errors,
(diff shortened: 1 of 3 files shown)


tests/test_visitors/test_ast/test_keywords/test_consistency_returning/test_consistency_return.py
```diff
@@ -97,6 +97,13 @@ def function():
     return
 """
 
+wrong_example4 = """
```
sobolevn: Thanks! Can you please also make the similar fixture for `yield`?

wemake_python_styleguide/visitors/ast/keywords.py
```diff
@@ -121,6 +121,8 @@ def _iterate_returning_values(
         returns: List[ast.Return] = []
         has_values = False
         for sub_node in ast.walk(node):
+            if get_context(sub_node) != node:
```
sobolevn: This code is now too complex, a bit of refactoring is required.

ulturt: yes, fixed

PR #955, opened 2019-10-29
Use `i_control_code` option in WPS113 (#951)

### LINKED ISSUE #951: Make `WPS113` respect `i_control_code` setting
Currently `WPS113` raises a violation on

```python
# __init__.py
from classes.typeclass import typeclass as typeclass
```

Note this part: `typeclass as typeclass`
Why do we use it? Because we use `mypy` with `implicit_reexport = False` setting on.
So, when we define public interface for a module we need to use explicit reexport in a form of `from classes.typeclass import typeclass as typeclass`

So, that's why we need to tweak `WPS113` not to highlight these errors when we make libraries. 
In other words, when `i_control_code` is set to `True`, we need to ignore `WPS113` at all.

ko3luhbka:
# I have made things!

## Checklist

## Related issues

- Closes #951

### DIFF
--- a/tests/test_checker/test_noqa.py
+++ b/tests/test_checker/test_noqa.py
@@ -33,7 +33,7 @@ SHOULD_BE_RAISED = types.MappingProxyType({
     'WPS110': 3,
     'WPS111': 1,
     'WPS112': 1,
-    'WPS113': 1,
+    'WPS113': 0,
     'WPS114': 1,
     'WPS115': 1,
     'WPS116': 1,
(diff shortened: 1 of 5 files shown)


tests/test_checker/test_noqa.py
```diff
@@ -33,7 +33,7 @@
     'WPS110': 3,
     'WPS111': 1,
     'WPS112': 1,
-    'WPS113': 1,
+    'WPS113': 0,
```
sobolevn: So, what happened here? Why this was changed?

ko3luhbka: Since now `WPS113` is raised only when `i_control_code = False`, the error doesn't occur during `test_noqa_fixture_disabled` test run. Maybe we should have separate list of violations which only occur when `i_control_code` = False? Like WPS411, for example.
Or maybe I don't understand or missed something?

sobolevn: `i_control_code = False` list would be awesome to have! Will you please open a new issue for that?

ko3luhbka: Done!
#957

PR #967, opened 2019-11-02
Replace scripts/parse.py witn an external tool.

lensvol:
# I have made things!

This should allow people to decipher the output of an `ast` module with more ease.

## Checklist


lensvol:
Sure, I completely missed that one. 🤦‍♂️

CONTRIBUTING.md
```diff
@@ -103,10 +103,10 @@ This step is mandatory during the CI.
 
 We also have several helpers to make your development work easier:
 
-- `./scripts/parse.py` is used to visualize `ast` nodes in other python modules,
-  usage: `python ./scripts/parse.py my_module.py`
-- `./scripts/tokens.py` is used to visualize tokens in other python modules,
-  usage: `python ./scripts/tokens.py my_module.py`
+- `astboom` is used to visualize `ast` nodes in other python modules,
+  usage: `astboom < my_module.py`
```
sobolevn: Btw, it is not related to this PR at all, I am just wondering.
Why do these two tools differ in usage? `<` vs `filename`

lensvol: Well, it does not really make sense to run `astboom` on files since even a simple Python expression produces sprawling AST visualization.

The intended mode of usage is something like `astboom 'print("Hello, world!")` with reading from STDIN as a fallback option. But this tool will continue to evolve in the nearest future and there is a chance I'll revisit that decision sooner or later. :)

PR #982, opened 2019-11-05
Add separate tests for violations controlled by 'i_control_code' option (#957)

### LINKED ISSUE #957: Add separate test for violations controlled by `i_control_code` option
Some violations are ignored when `i_control_code` options is set to `True` (like WPS113). Thus, we have to disable them in `SHOULD_BE_RAISED` list because e2e tests run with `i_control_code = True`. So these violations are actually not completely tested.
Why not to have a separate list of violations which only raise when `i_control_code = False`? And run separate test(s) with `--i_control_code=False`.

ko3luhbka:
# I have made things!

## Checklist

## Related issues

- Closes #957 

I had to split `tests/conftest.py` into two modules because WPS202 was raised after I've added `all_controlled_violations` fixture in `conftest.py`.

### DIFF
--- a/tests/test_checker/test_noqa.py
+++ b/tests/test_checker/test_noqa.py
@@ -228,6 +228,7 @@ SHOULD_BE_RAISED = types.MappingProxyType({
     'WPS702': 1,
 })
 
+# Violations which may be tweaked by `i_control_code` option
 SHOULD_BE_RAISED_NO_CONTROL = types.MappingProxyType({
     'WPS113': 1,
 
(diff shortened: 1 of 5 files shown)


tests/test_checker/test_noqa.py
```diff
@@ -228,6 +228,13 @@
     'WPS702': 1,
 })
 
+SHOULD_BE_RAISED_NO_CONTROL = types.MappingProxyType({
```
sobolevn: Inline comment here would be awesome.

ko3luhbka:
Not shure about correct place for describing `SHOULD_BE_RAISED_NO_CONTROL`.

sobolevn:
Thanks a lot!

PR #992, opened 2019-11-08
Remove false positive overlap error on simple annotations without assign

### LINKED ISSUE #987: False positive BlockAndLocalOverlapViolation with type annotation
# Bug report

## What's wrong
With given code
```python
import asyncpg as apg


async def db_operation(db: apg.pool.Pool):
    """Operation with db."""
    connection: apg.Connection
    async with db.acquire() as connection:
        return await connection.fetch(
            """
            SELECT * FROM table
            """
        )
```
`BlockAndLocalOverlapViolation` raises with 
```
     7:   5 WPS440 Found block variables overlap: connection [wemake-python-styleguide]
  async with db.acquire() as connection:
  ^
```

Without line ```connection: apg.Connection``` everything is ok.

## How is that should be
This is valid python annotation and `BlockAndLocalOverlapViolation` violation should not be raised here.

AlwxSin:
closes #987

# I have made things!

## Checklist

## Related issues
#987

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ We used to have incremental versioning before `0.1.0`.
 - Fixes that `--i-dont-control-code` was not present in command line options
 - Fixes `BlockVariableVisitor` false positives on a properties
 - Fixes that `//` was not recognised as a math operation
+- Fixes false positive `BlockAndLocalOverlapViolation` on annotations without value assign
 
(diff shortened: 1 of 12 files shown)


AlwxSin:
I had to change the `assign_statement` fixture for many tests.

sobolevn:
Thanks a lot!

PR #1038, opened 2019-11-29
Fix false positive for WPS504 (#1036)

### LINKED ISSUE #1036: False positive for WPS504
# Bug report

## What's wrong

This code produces a false positive violation:

```python
def test():
    delta = due - since
    if not delta.days:
        raise ValidationError(FormatErrorMessages.empty_duration)
    elif delta.days > max_days:
        raise ValidationError(FormatErrorMessages.too_long_duration)
    return due
```

Output:

```
361   51:5     WPS504 Found negated condition
362   if not delta.days:
363   ^
```

## How is that should be

We don't have an `else` clause here. I cannot switch this condition.
So, when `else` is missing, we should ignore this violation.

## System information

This actually happens on `0.12.5`

```
» flake8 --version
3.7.9 (flake8-annotations-complexity: 0.0.2, flake8-bandit: 2.1.2, flake8-broken-line: 0.1.1, flake8-bugbear: 19.8.0, flake8-comprehensions: <cached_property.cached_property object at 0x103968438>, flake8-debugger: 3.2.1, flake8-django: 0.0.3, flake8-docstrings: 1.5.0, pydocstyle: 4.0.1, flake8-eradicate: 0.2.3, flake8-executable: 2.0.3, flake8-print: 3.1.4, flake8-pytest: 1.3, flake8-string-format: 0.2.3, flake8_builtins: 1.4.1, flake8_coding: 1.3.2, flake8_commas: 2.0.0, flake8_isort: 2.3, flake8_pep3101: 1.2.1, flake8_quotes: 2.1.1, logging-format: 0.6.0, mccabe: 0.6.1, naming: 0.8.2, pycodestyle: 2.5.0, pyflakes: 2.1.1, radon: 2.4.0, rst-docstrings: 0.0.11, wemake-python-styleguide: 0.12.5) CPython 3.7.3 on Darwin
```

GreyZmeem:
## Checklist

## Related issues

- Closes #1036


sobolevn:
@GreyZmeem just for the record why did you close it? 🙂 
Is there anything I can help you with?

GreyZmeem:
@sobolevn CI check failed for my PR because of `WPS214 Found too many methods: 8` violation in `IfStatementVisitor` class (I've added new method for checking `if` node has `else` expression). And... I'm not sure how I can refactor it :) Moving this code into `_check_negated_conditions` method either violates complexity check or looks ugly. 

I thought I could to move this code to some kind of `utils` or `helpers` module, but I couldn’t find one (maybe I overlooked it).
Therefore, so far I have decided to close the PR, because I’m not sure what to do.

https://github.com/wemake-services/wemake-python-styleguide/blob/5570c73de4dd393b7303daec233f2b6005889e1e/wemake_python_styleguide/visitors/ast/conditions.py#L61

wemake_python_styleguide/visitors/ast/conditions.py
```diff
@@ -154,6 +154,16 @@ def _check_useless_len(self, node: AnyIf) -> None:
             if given_function_called(node.test, {'len'}):
                 self.add_violation(UselessLenCompareViolation(node))
 
+    def _has_else_expr(self, node: ast.If) -> bool:
```
sobolevn: Let's move it to https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/logic/ifs.py

sobolevn:
@GreyZmeem reopened. This was an easy way to resolve your problem 👍

tests/test_visitors/test_ast/test_conditions/test_negated_conditions.py
```diff
@@ -12,9 +12,18 @@
...
 """
 
+complex_elif_conditions = """
+if {0}:
+    ...
+elif ...:
+    ...
+"""
+
 complex_conditions = """
```
sobolevn: Can you please leave the test case:

```python
if {0}:
    ...
else:
    ...
```

And create a new one:

```python
if {0}:
    ...
elif ...:
    ...
else:
    ...
```

GreyZmeem: @sobolevn I've parameterized the tests with the template variable, since the behavior should not differ in this case from whether there is an `elif` block or not.

PR #1065, opened 2019-12-09
Warn about descriptors not in methods

### LINKED ISSUE #1033: Warn about descriptors not in methods
# Rule request

## Thesis

Descriptors do magic only as methods. Let's warn about using them for regular functions:

```python
@classmethod
def f():
    ...

@staticmethod
def f():
    ...

@property
def f():
    ...
```

## Reasoning

In most cases, it's a typo, and that function was intended to be a method.

zteoh:
Address issue #1033
Warn if descriptions (@staticmethod, @classmethod, @property) is used for regular functions not in a class.

# I have made things!

## Checklist

## Related issues

Closes #1033

### DIFF
--- a/tests/test_checker/test_noqa.py
+++ b/tests/test_checker/test_noqa.py
@@ -219,7 +219,6 @@ SHOULD_BE_RAISED = types.MappingProxyType({
     'WPS447': 1,
     'WPS448': 1,
     'WPS449': 1,
-<<<<<<< HEAD
     'WPS450': 1,
     'WPS451': 0,  # defined in version specific table.
     'WPS452': 1,  # also defined in version specific table.
@@ -227,8 +226,6 @@ SHOULD_BE_RAISED = types.MappingProxyType({
     'WPS454': 1,
     'WPS455': 1,
     'WPS456': 1,
-=======
->>>>>>> make requested changes
 
     'WPS500': 1,
     'WPS501': 1,
--- a/tests/test_checker/test_noqa.py
+++ b/tests/test_checker/test_noqa.py
@@ -219,6 +219,7 @@ SHOULD_BE_RAISED = types.MappingProxyType({
     'WPS447': 1,
     'WPS448': 1,
     'WPS449': 1,
+<<<<<<< HEAD
     'WPS450': 1,
     'WPS451': 0,  # defined in version specific table.
     'WPS452': 1,  # also defined in version specific table.
@@ -226,6 +227,8 @@ SHOULD_BE_RAISED = types.MappingProxyType({
     'WPS454': 1,
     'WPS455': 1,
     'WPS456': 1,
+=======
+>>>>>>> make requested changes
 
     'WPS500': 1,
     'WPS501': 1,
@@ -274,6 +277,7 @@ SHOULD_BE_RAISED = types.MappingProxyType({
     'WPS611': 1,
     'WPS612': 1,
     'WPS613': 1,
+    'WPS614': 1,
 })
 
 # Violations which may be tweaked by `i_control_code` option:
--- a/wemake_python_styleguide/violations/best_practices.py
+++ b/wemake_python_styleguide/violations/best_practices.py
@@ -1974,6 +1974,7 @@ class FloatKeyViolation(ASTViolation):
 @final
 class ProtectedModuleMemberViolation(ASTViolation):
     """
+<<<<<<< HEAD
     Forbid importing protected objects from modules.
 
     Related to :class:`~ProtectedModuleViolation`.
@@ -1986,6 +1987,21 @@ class ProtectedModuleMemberViolation(ASTViolation):
     Solution:
         Do not import protected objects from modules.
         Respect the encapsulation.
+=======
+    Forbids descriptors in regular functions.
+
+    Forbids using @staticmethod, @classmethod and @property for functions not
+    in class
+
+    Reasoning:
+        Descriptors like @staticmethod, @classmethod and @property do magic
+        only as methods. We would want to warn users if the descriptors are
+        used on regular functions
+
+    Solution:
+        Do not use @staticmethod, @classmethod and @property on regular
+        functions or wrap the functions into a Class
+>>>>>>> fix styling and linting errors
 
     Example::
 
@@ -2000,7 +2016,14 @@ class ProtectedModuleMemberViolation(ASTViolation):
 
     """
 
+<<<<<<< HEAD
     error_template = 'Found protected object import: {0}'
+=======
+    error_template = """
+        Method decorators (eg. classmethod, staticmethod and property)
+        should not be used for functions
+        """
+>>>>>>> fix styling and linting errors
     code = 450
 
 
--- a/wemake_python_styleguide/violations/best_practices.py
+++ b/wemake_python_styleguide/violations/best_practices.py
@@ -73,7 +73,6 @@ Summary
    BaseExceptionRaiseViolation
    NonTrivialExceptViolation
    FloatingNanViolation
-   MethodDecoratorUsedForFunctionViolation
 
 Best practices
 --------------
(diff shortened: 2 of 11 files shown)


tests/test_checker/test_noqa.py
```diff
@@ -185,6 +185,7 @@
     'WPS447': 1,
     'WPS448': 1,
     'WPS449': 1,
+    'WPS450': 0,
```
sobolevn: This violation should be present in `tests/fixtures/noqa.py`
And this should equal to `1`

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -69,6 +69,7 @@
    StringConstantRedefinedViolation
    IncorrectExceptOrderViolation
    FloatKeyViolation
+   MethodDecoratorUsedForFunctionViolation
```
sobolevn: This should be located in `oop.py`, I think that it is the most semantical place.

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -1941,3 +1943,43 @@ class FloatKeyViolation(ASTViolation):
...
     error_template = 'Found float used as a key'
     code = 449
+
+
+@final
+class MethodDecoratorUsedForFunctionViolation(ASTViolation):
+    """
+    Forbids descriptors in regular functions.
+
+    Forbids using @staticmethod, @classmethod and @property for functions not
```
sobolevn: We use double \` chars to escape code, so it would render as `@staticmethod`

sobolevn: Please, mark all `@staticmethod`,  `@classmethod`, and `@property` like so

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -1941,3 +1943,43 @@ class FloatKeyViolation(ASTViolation):
...
+    in class
+
+    Reasoning:
+        Descriptors like @staticmethod, @classmethod and @property do magic
+        only as methods. We would want to warn users if the descriptors are
+        used on regular functions
```
sobolevn: Missing leading `.` in the end of the sentence

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -1941,3 +1943,43 @@ class FloatKeyViolation(ASTViolation):
...
+
+    Solution:
+        Do not use @staticmethod, @classmethod and @property on regular
+        functions or wrap the functions into a Class
+
+    Example::
+
+        # Correct:
+        Class TestClass(Object):
+            @staticmethod
```
sobolevn: Let's use `@property` here as an example. Because we don't like `@staticmethod`s in this project 🙂

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -1941,3 +1943,43 @@ class FloatKeyViolation(ASTViolation):
...
+
+        # Wrong:
+        @staticmethod
+        def myFunction():
+
+    .. versionadded:: 0.15.0
+
+    """
+
+    error_template = """
```
sobolevn: `error_template = 'Method decorators should not be used for functions'`

wemake_python_styleguide/visitors/ast/complexity/counts.py
```diff
@@ -72,6 +77,31 @@ def _check_decorators_count(self, node: ModuleMembers) -> None:
                 ),
             )
 
+    def _check_method_decorators(self, node: ModuleMembers) -> None:
+        parent_is_class = isinstance(get_parent(node), ast.ClassDef)
```
sobolevn: You can early return here:

```python
if isinstance(get_parent(node), ast.ClassDef):
     return   # classes can contain descriptors
```

wemake_python_styleguide/visitors/ast/complexity/counts.py
```diff
@@ -72,6 +77,31 @@ def _check_decorators_count(self, node: ModuleMembers) -> None:
...
+
+        # Check whether filtered decorators are method decorators
+        is_method_decorator = []
+        for decorator in node.decorator_list:
+            if isinstance(decorator, ast.Name):
+                is_method_decorator.append(
+                    decorator.id in {'classmethod', 'staticmethod', 'property'},
```
sobolevn: `{'classmethod', 'staticmethod', 'property'}` needs to be moved into a constant on a class level:

```python
_descriptor_decorators = frozenset((
    'classmethod', 
    'staticmethod', 
    'property',
))
```

wemake_python_styleguide/visitors/ast/complexity/counts.py
```diff
@@ -72,6 +77,31 @@ def _check_decorators_count(self, node: ModuleMembers) -> None:
                 ),
             )
 
+    def _check_method_decorators(self, node: ModuleMembers) -> None:
+        parent_is_class = isinstance(get_parent(node), ast.ClassDef)
+
+        # Check whether filtered decorators are method decorators
+        is_method_decorator = []
```
sobolevn: When something is named as `is_*` we expect it to be `bool`, not a list.
Let's call this variable `method_decorators`

wemake_python_styleguide/visitors/ast/complexity/counts.py
```diff
@@ -72,6 +77,31 @@ def _check_decorators_count(self, node: ModuleMembers) -> None:
...
+        for decorator in node.decorator_list:
+            if isinstance(decorator, ast.Name):
+                is_method_decorator.append(
+                    decorator.id in {'classmethod', 'staticmethod', 'property'},
+                )
+
+        # Check for presence of method deecorators
+        has_method_decorators = (
+            reduce(
+                lambda boolean1, boolean2: boolean1 or boolean2,
```
sobolevn: You can use `operator.or_` here. With `import operator`

wemake_python_styleguide/visitors/ast/complexity/counts.py
```diff
@@ -72,6 +77,31 @@ def _check_decorators_count(self, node: ModuleMembers) -> None:
...
+                is_method_decorator,
+                0,
```
sobolevn: It should be `False`, not `0`

zteoh: Thank you so much for all the feedback and comments! I agree with all of them and really appreciate it! I have a question with changing `0` to `False` because it seems to be giving me this error when i run `flake8 .`:
```
./wemake_python_styleguide/visitors/ast/complexity/counts.py

  96:17    WPS425 Found boolean non-keyword argument: False
  False,
  ^
```
Should I still change it to `False` even though it would mean the `flake8 .` would not pass?

Thanks!

zteoh:
Hi @sobolevn !

Thank you so much for your swift response and the input! I have made all the changes requested. Currently, it is failing a flake8 test and the error message is:

```
./wemake_python_styleguide/visitors/ast/complexity/counts.py

  106:17   WPS425 Found boolean non-keyword argument: False
  False,
  ^
```
Do let me know if you know any way to solve this problem. I am trying to have `False` as the default for the `reduce` function the case where the list `method_decorators` is empty.

Thanks so much!

wemake_python_styleguide/visitors/ast/complexity/counts.py
sobolevn: Try `initializer=False`

zteoh: I'm getting this error when I changed it to
```
# Check for presence of method deecorators
has_method_decorators = (
    reduce(
        lambda boolean1, boolean2: operator.or_(boolean1, boolean2),
        method_decorators,
        initializer=False,
    )
)
```

```
TypeError: reduce() takes no keyword arguments
```

zteoh: I managed to figure out that storing `false` in a variable would not raise the error. What do you think of this:

```
default_param = False
has_method_decorators = (
    reduce(
        lambda boolean1, boolean2: operator.or_(boolean1, boolean2),
        method_decorators,
        default_param,
    )
)
```

sobolevn: I guess, `# noqa` is just better in this case.

sobolevn: Thanks a lot! (Oh, I really dislike this builtin functions which do not allow kw-args!)

PR #1066, opened 2019-12-09
Added new rule for checking for infinite while loops. Added test call…

### LINKED ISSUE #1007: Warn about possibly infinite `while`
# Rule request

Check that `while True` block has `raise` or `return`.

**UPD**: and `break` too. 

## Thesis

It's OK to not have them for CLI entry point, but in other cases, it can be an error. So, let's ask users explicitly `noqa` such cases to be sure that they really know what they're doing.

ragohin:
…ed test_infinite_while_loops.py

# I have made things!

## Checklist

## Related issues

Closes #1007


wemake_python_styleguide/visitors/ast/loops.py
```diff
@@ -203,6 +206,25 @@ def _check_multiline_loop(self, node: _AnyLoop) -> None:
                 self.add_violation(MultilineLoopViolation(node))
                 break
 
+    def _has_break_raise_or_return(self, node: _AnyLoop) -> bool:
```
sobolevn: This can be moved to `logic/` dir

wemake_python_styleguide/visitors/ast/loops.py
```diff
@@ -203,6 +206,25 @@ def _check_multiline_loop(self, node: _AnyLoop) -> None:
...
+            if not isinstance(subnode, ast.While):
+                if isinstance(subnode, ast.Return):
+                    return True
+                if isinstance(subnode, ast.Raise):
+                    return True
+                if isinstance(subnode, ast.Break):
+                    return True
+        return False
+
+    def _check_infinite_while_loop(self, node: _AnyLoop) -> None:
```
sobolevn: This can be moved to a another or a new visitor, that is `ast.While` specific.

ragohin: Shouldn't this be inside of this class though, since it says at line 129 "Checks ''for'' and ''while'' loops" ? It seems like _check_multiline_loop also is ast.While specific.

wemake_python_styleguide/visitors/ast/loops.py
ragohin: Done!

wemake_python_styleguide/visitors/ast/loops.py
ragohin: I think what I have implemented as of now might work, but let me know!

wemake_python_styleguide/logic/nodes.py
```diff
@@ -1,9 +1,12 @@
...
 
 import ast
-from typing import Optional
+from typing import Optional, Union
 
-from wemake_python_styleguide.types import ContextNodes
+from wemake_python_styleguide.compat.aliases import ForNodes
+from wemake_python_styleguide.types import AnyFor, ContextNodes
+
+_AnyLoop = Union[AnyFor, ast.While]
```
sobolevn: Should go to `types.py`

wemake_python_styleguide/logic/nodes.py
```diff
@@ -29,3 +32,36 @@ def get_parent(node: ast.AST) -> Optional[ast.AST]:
 def get_context(node: ast.AST) -> Optional[ContextNodes]:
     """Returns the context or ``None`` if node has no context."""
     return getattr(node, 'wps_context', None)
+
+
+def does_loop_contain_node(  # TODO: move, reuse in annotations.py
```
sobolevn: Please, remove `TODO`

wemake_python_styleguide/logic/nodes.py
```diff
@@ -29,3 +32,36 @@ def get_parent(node: ast.AST) -> Optional[ast.AST]:
...
+    loop: Optional[_AnyLoop],
+    to_check: ast.Break,
+) -> bool:
+    """Helper function for has_break method."""
```
sobolevn: Please, change the description to something more meaningful.

wemake_python_styleguide/logic/nodes.py
sobolevn: Do we actually use this method from outside? If not, it should be protected. `_does_loop_contain_node`

wemake_python_styleguide/visitors/ast/loops.py
```diff
@@ -208,22 +178,21 @@ def _check_multiline_loop(self, node: _AnyLoop) -> None:
 
     def _has_break_raise_or_return(self, node: _AnyLoop) -> bool:
```
sobolevn: This method can still be moved into `logic/`

wemake_python_styleguide/visitors/ast/loops.py
```diff
@@ -338,6 +307,6 @@ def _check_implicit_items(self, node: ast.For) -> None:
 
     def _is_assigned_target(self, node: ast.Subscript) -> bool:
         parent = nodes.get_parent(node)
-        if not isinstance(parent, (*AssignNodes, ast.AugAssign)):
```
sobolevn: Are you sure about this change? Why do you have to remove `ast.AugAssign`?

wemake_python_styleguide/visitors/ast/loops.py
ragohin: Sorry, that function should not have been in this commit in that location. It has been moved to logic/.

wemake_python_styleguide/logic/nodes.py
ragohin: Done!

wemake_python_styleguide/logic/nodes.py
ragohin: Done!

wemake_python_styleguide/logic/nodes.py
ragohin: Done!

wemake_python_styleguide/logic/nodes.py
ragohin: Done!

wemake_python_styleguide/visitors/ast/loops.py
ragohin: Sorry, I don't remember changing that, I must've pulled an outdated version of the code. Fixed.

ragohin:
I added one more small change so that "while True:" loops are defined with print("forever") inside. Hopefully everything is correct now!

tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
```diff
@@ -0,0 +1,89 @@
...
+
+import pytest
+
+from wemake_python_styleguide.violations.best_practices import (
+    InfiniteWhileLoopViolation,
+)
+from wemake_python_styleguide.visitors.ast.loops import WrongLoopVisitor
+
+incorrect_loop1 = """
+while True:
```
sobolevn: Looks like there are a lot of repetitive patterns in these samples.
Let's parametrize them!

Like so:

```python
incorrect_loop_template = """
while True:
    {0}
"""
```

And later we can use this way of testing:
```python
@pytest.mark.parametrize('keyword', [
    'break',
    'return',
    'return some',
    'raise Some',
    'raise Some()',
    'raise',
]
def test_whatever(keyword):
     code = template.format(keyword)
```

We also need to test a nested keyword: 

```python
while True:
    if some:
        {0}
```

ragohin: Alright I think I fixed everything! Let me know what you think!

tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
```diff
@@ -7,58 +7,77 @@
...
     while True:
         print('forever')
-        break
+        {0}
+"""
+
+incorrect_loop_template_nonfunction3 = """
+while True:
+    if some:
+        print('forever')
```
sobolevn: This is not parametrized

ragohin:
I think there is an issue with the configuration on the laptop I pushed my code with most recently.

I originally was working on my Mac, which just crashed, so I switched to wsl on my Windows computer. I think the code will build properly if I submit on my Mac, but I can't use it right now :/

Is there any work around with this because I'm not sure I can satisfy the Travis CI build with my current laptop?

tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
ragohin: How can I parametrize this in a way where {0} accepts any command except for (raise, return, and break)?

ragohin: Basically my most recent pull request has (raise, return, break, etc) parametrized, but not "print('forever')" which should just be replaced by any command that is not (raise, return, break, etc).

Issue #1103, opened 2020-01-11
move some constants to optional configuration options


skarzi:
hi :wave: 

I am using `wemake-python-styleguide` in several types of python's projects and some of them are built using framework e.g scrapy. Such frameworks and sometimes also domain language force developers to use some names that are forbidden by few `wemake-python-styleguide` violations, e.g [WPS110](https://wemake-python-stylegui.de/en/latest/pages/usage/violations/naming.html#wemake_python_styleguide.violations.naming.WrongVariableNameViolation). 
It would be really great to make `setup.cfg`'s options from some constants, for instance `VARIABLE_NAMES_BLACKLIST` to easily support above mentioned cases.
Please, let me know what is your opinion about this. 

Thanks for such great project!
Kind Regards,
Lukasz

sobolevn:
Hi, @skarzi! Thanks a lot for your feedback.

I currently do not want to make this configurable, because it will allow to mostly disable this check. And I have spent years on figuring out on what to put there.

Can you please be more specific? What names are conflicting with scrapy?

skarzi:
that's true, however I think we can assume that developers know what they are doing.
This option will be used really rarely, but can be useful in some cases.

In `scrapy` it's popular to use name `item` (for instances of [`scrapy.item.Item`](https://docs.scrapy.org/en/latest/topics/items.html#scrapy.item.Item) class)  and few methods that developers sometimes override take parameter with such name.

sobolevn:
How about `allowed-domain-names=['item', 'handle']`? This will remove several names from the list. The name will show that it should only be changed when the domain tells you so.

And in contrast we can add `forbidden-domain-names` that will be added to the list. Like sometimes you cannot name your variable as `index`, you have to use `zip` or something like this. I am not sure about this part, though.

skarzi:
sounds great!
Can I prepare PR with this features?

sobolevn:
Sure! Thanks!

Issue #1019, opened 2019-11-24
Forbid to mix imports types


sobolevn:
# Rule request

## Thesis

When we write code with lot's of imports it is easy to forget to watch for already existing imports. So, we can end up with a situation like this:

```python
from a import b
from a.b import c
```

This is error-prone! Because you can write `b.c` and `c` which is the same thing!
Your code will start taking damage from it.

uhbif19:
Do I understand correctly, that this is about forbidding only the imports leading to the same thing represented in two different ways? Is your import example the only one possible, or there can be more complex cases exist?

sobolevn:
> Do I understand correctly, that this is about forbidding only the imports leading to the same thing represented in two different ways?

Yes

> Is your import example the only one possible, or there can be more complex cases exist?

I cannot think about any other cases. It might be worth to check cases like:

```python
from a import b
from a.b.c import d
```

But, I guess it is a different thing.

kxepal:
With hypothesis common pattern is
```
import hypothesis
from hypothesis import strategies as st
```

While you can do `hypothesis.strategies` with a single import, it's too long to write all the time. Does this case violates proposed rule?

sobolevn:
@kxepal that's a great case! aliases should be allowed.

skarzi:
Can I prepare PR for this issue?

sobolevn:
Yes, please!

PR #1106, opened 2020-01-13
add new domain names options

skarzi:
# I have made things!

## Checklist

## Related issue

Closes: #1103 

Introduce 2 new options:
    
+ `--allowed-domain-names` - remove given names from variable names'
   blacklist
+ `--forbidden-domain-names` - extends variable names' blacklist with
   given names
    
The options listed above are used to create variable names' blacklist
which is leveraged to detect `WrongVariableNameViolation` violations.

### DIFF
--- a/wemake_python_styleguide/logic/naming/blacklists.py
+++ b/wemake_python_styleguide/logic/naming/blacklists.py
@@ -1,15 +1,19 @@
 # -*- coding: utf-8 -*-
 
-from typing import Set
+from typing import FrozenSet
 
 from wemake_python_styleguide.constants import VARIABLE_NAMES_BLACKLIST
 from wemake_python_styleguide.types import ConfigurationOptions
 
 
-def variable_names_blacklist_from(options: ConfigurationOptions) -> Set[str]:
+def variable_names_blacklist_from(
+    options: ConfigurationOptions,
+) -> FrozenSet[str]:
     """Creates variable names blacklist from options and constants."""
     variable_names_blacklist = {
         *VARIABLE_NAMES_BLACKLIST,
         *options.forbidden_domain_names,
     }
-    return variable_names_blacklist - set(options.allowed_domain_names)
+    return frozenset(
+        variable_names_blacklist - set(options.allowed_domain_names),
+    )
--- a/wemake_python_styleguide/logic/naming/blacklists.py
+++ b/wemake_python_styleguide/logic/naming/blacklists.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+
+from typing import Set
+
+from wemake_python_styleguide.constants import VARIABLE_NAMES_BLACKLIST
+from wemake_python_styleguide.types import ConfigurationOptions
+
+
+def variable_names_blacklist_from(options: ConfigurationOptions) -> Set[str]:
+    """Creates variable names blacklist from options and constants."""
+    variable_names_blacklist = {
+        *VARIABLE_NAMES_BLACKLIST,
+        *options.forbidden_domain_names,
+    }
+    return variable_names_blacklist - set(options.allowed_domain_names)
--- a/wemake_python_styleguide/visitors/ast/naming.py
+++ b/wemake_python_styleguide/visitors/ast/naming.py
@@ -3,7 +3,7 @@
 import ast
 import itertools
 from collections import Counter
-from typing import Callable, Iterable, List, Optional, Tuple, Union, cast, Set
+from typing import Callable, Iterable, List, Optional, Set, Tuple, Union, cast
 
 from typing_extensions import final
 
--- a/wemake_python_styleguide/visitors/ast/naming.py
+++ b/wemake_python_styleguide/visitors/ast/naming.py
@@ -12,11 +12,11 @@ from wemake_python_styleguide.compat.functions import get_assign_targets
 from wemake_python_styleguide.constants import (
     MODULE_METADATA_VARIABLES_BLACKLIST,
     SPECIAL_ARGUMENT_NAMES_WHITELIST,
-    VARIABLE_NAMES_BLACKLIST,
 )
 from wemake_python_styleguide.logic import functions, nodes
 from wemake_python_styleguide.logic.naming import (
     access,
+    blacklists,
     builtins,
     logical,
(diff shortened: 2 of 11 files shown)


wemake_python_styleguide/visitors/ast/naming.py
```diff
@@ -109,6 +112,17 @@ def check_attribute_name(self, node: ast.ClassDef) -> None:
                     naming.UpperCaseAttributeViolation(target, text=target.id),
                 )
 
+    @property
```
sobolevn: It can be:
1. Cached
2. Moved to `logic/`

wemake_python_styleguide/visitors/ast/naming.py
```diff
@@ -109,6 +112,17 @@ def check_attribute_name(self, node: ast.ClassDef) -> None:
...
+    def variable_names_blacklist(self) -> Set[str]:
+        if self._variable_names_blacklist is None:
+            self._variable_names_blacklist = {
+                *VARIABLE_NAMES_BLACKLIST,
+                *self._options.forbidden_domain_names,
+            }
+            for name in self._options.allowed_domain_names:
```
sobolevn: Can we use just `set` arithmetic here?

AlwxSin: Shouldn't be an error if `allowed` and `forbidden` names intersects?

AlwxSin: At least, if option variables intersect.

sobolevn: Yes, I agree. But! Only during validation. Not in runtime

wemake_python_styleguide/visitors/ast/naming.py
skarzi: > 1. Cached

Current implementation provide caching, it'a similar to `@cached_property` from `django`.
Do you want to introduce new dependency e.g. [`cached-property`](https://pypi.org/project/cached-property/) just for it?

> 2. Moved to `logic/`

I can move it there if you want, but could you tell me to which file or maybe I should create new one?

sobolevn: You can create a new one 👍

sobolevn: No new dependecies are required, you can try to use `lru_cache`

wemake_python_styleguide/visitors/ast/naming.py
skarzi: is [`_ValidatedOptions.validate_options()`](https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/options/validation.py#L72) proper place to add above mentioned validation logic?

wemake_python_styleguide/visitors/ast/naming.py
skarzi: However when the code responsible for creating variable names' blacklist will be moved to `logic/` then it should be enough to just assign the result in `__init__`:

```python
def __init__(...):
    ...
    self.variable_names_blacklist = logic.create_variable_names_blacklist()
```

What do you think?

wemake_python_styleguide/visitors/ast/naming.py
sobolevn: Yeap!

wemake_python_styleguide/visitors/ast/naming.py
sobolevn: Sounds fine!

wemake_python_styleguide/visitors/ast/naming.py
```diff
@@ -45,7 +45,7 @@
 class _NameValidator(object):
     """Utility class to separate logic from the naming visitor."""
 
-    _variable_names_blacklist: Optional[Set[str]]
+    variable_names_blacklist: Set[str]
```
sobolevn: Why is it public?

wemake_python_styleguide/logic/naming/blacklists.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+
+from typing import Set
+
+from wemake_python_styleguide.constants import VARIABLE_NAMES_BLACKLIST
+from wemake_python_styleguide.types import ConfigurationOptions
+
+
+def variable_names_blacklist_from(options: ConfigurationOptions) -> Set[str]:
```
sobolevn: Can we return `frozenset`?

sobolevn: And can we use [`lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache) here?

wemake_python_styleguide/visitors/ast/naming.py
skarzi: To be honest I don't know :D I found it more proper/clear when I was writing this code. I will correct it in following commit

wemake_python_styleguide/logic/naming/blacklists.py
skarzi: > And can we use [`lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache) here?

I don't know the whole architecture and code flow well, but caching seems to be not needed, because this functions should be ran only once (in the `_NameValidator.__init__()` which is created only once in `WringNameVisitor.__init__()` that again should be created only once by checker), am I right?

skarzi: please, let me know if you still think `lru_cache` is needed

sobolevn:
Merging into `0.14`

Thanks again!

wemake_python_styleguide/logic/naming/blacklists.py
skarzi: @sobolevn during working on #1019 I have found that I was wrong about caching and it's needed, because `Checker` is created for each file being checked, sorry!
Should I review other similar cases and add some issue? Or maybe i should just correct `variable_names_blacklist_from()` in new PR?

sobolevn: Yes, please. We can create a separate issue and discuss it there.

Issue #1113, opened 2020-01-18
Split WPS436 into two rules


sobolevn:
We need to split `WPS436` into two separate violations. Let's see two examples:

```
  1:1      WPS436 Found protected module import
  from some._internal import one
```

and

```
  1:1      WPS436 Found protected module import
  from some.internal import _one
```

The first one import from some protected module. Which is ok in some cases like: `pip`, `deal`, and `stories`. When we develop some package we can import from protected modules and packages, that's fine.

But, we can never import protected things! Protected objects, functions, classes, etc should never leave a single module.

The current problem is that people switch off `WPS436` to allow import like `from some._internal import correct` and it also switches off `from some.internal import _wrong`

This should be split into two violations:

1. `from some.internal import _wrong`
2. `from some._internal import correct`

skarzi:
Can I take it?

sobolevn:
Sure! Thanks a lot! It will be merged into `0.14`

skarzi:
How imports like `import _pytest` should be handled?
Such case is quite tricky because it both imports protected module and loads object with protected name to globals. 
Do such cases should raise `ProtectedModuleViolation`, this new violations (let's say `ProtectedModuleMemberViolation`) or both violations?

sobolevn:
It should not be allowed as "importing from protected path" or `ProtectedModuleViolation`

PR #1118, opened 2020-01-22
extract new violation - WPS450 from WPS436

skarzi:
# I have made things!

## Checklist

## Related issues

Closes: #1113 

### DIFF
--- a/wemake_python_styleguide/violations/best_practices.py
+++ b/wemake_python_styleguide/violations/best_practices.py
@@ -69,6 +69,7 @@ Summary
    StringConstantRedefinedViolation
    IncorrectExceptOrderViolation
    FloatKeyViolation
+   ProtectedModuleMemberViolation
 
 Best practices
 --------------
@@ -123,6 +124,7 @@ Best practices
 .. autoclass:: StringConstantRedefinedViolation
 .. autoclass:: IncorrectExceptOrderViolation
 .. autoclass:: FloatKeyViolation
+.. autoclass:: ProtectedModuleMemberViolation
 
 """
 
@@ -1431,21 +1433,22 @@ class ProtectedModuleViolation(ASTViolation):
         our code at any moment.
 
     Solution:
-        Do not import anything from protected modules.
+        Do not import protected modules.
         Respect the encapsulation.
 
     Example::
 
         # Correct:
+        import public_module
         from some.public.module import FooClass
 
         # Wrong:
         import _compat
         from some._protected.module import BarClass
-        from some.module import _protected
 
     .. versionadded:: 0.3.0
     .. versionchanged:: 0.11.0
+    .. versionchanged:: 0.14.0
 
     """
 
@@ -1942,3 +1945,35 @@ class FloatKeyViolation(ASTViolation):
 
     error_template = 'Found float used as a key'
     code = 449
+
+
+@final
+class ProtectedModuleMemberViolation(ASTViolation):
+    """
+    Forbids to import protected objects from modules.
+
+    Reasoning:
+        When importing protected modules' members we break a contract
+        that authors of this module enforce.
+        This way we are not respecting encapsulation and it may break
+        our code at any moment.
+
+    Solution:
+        Do not import protected objects from modules.
+        Respect the encapsulation.
+
+    Example::
+
+        # Correct:
+        from some.public.module import FooClass
+
+        # Wrong:
+        from some.module import _protected
+        from some.module import _protected as not_protected
+
+    .. versionadded:: 0.14.0
+
+    """
+
+    error_template = 'Found protected object import'
+    code = 450
--- a/wemake_python_styleguide/violations/best_practices.py
+++ b/wemake_python_styleguide/violations/best_practices.py
@@ -1426,6 +1426,8 @@ class ProtectedModuleViolation(ASTViolation):
     """
     Forbids to import protected modules.
 
+    Related to :class:`~ProtectedModuleMemberViolation`.
+
     Reasoning:
         When importing protected modules we break a contract
         that authors of this module enforce.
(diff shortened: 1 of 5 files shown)


wemake_python_styleguide/violations/best_practices.py
```diff
@@ -1942,3 +1945,35 @@ class FloatKeyViolation(ASTViolation):
 
     error_template = 'Found float used as a key'
     code = 449
+
+
+@final
+class ProtectedModuleMemberViolation(ASTViolation):
```
skarzi: I am not sure if this name is clear and meaningful enough, if you have some suggestions please let me know

sobolevn: It is fine! Thanks!

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -1942,3 +1945,35 @@ class FloatKeyViolation(ASTViolation):
...
+    Example::
+
+        # Correct:
+        from some.public.module import FooClass
+
+        # Wrong:
+        from some.module import _protected
+        from some.module import _protected as not_protected
+
+    .. versionadded:: 0.14.0
```
sobolevn: Can we please have a link to the previous violation like so:

```python
"""
Related to :class:`~ProtectedModuleViolation`.
"""
```

And the same for `ProtectedModuleViolation`, please.

wemake_python_styleguide/visitors/ast/imports.py
```diff
@@ -101,9 +123,11 @@ def visit_Import(self, node: ast.Import) -> None:
         Used to find wrong ``import`` statements.
 
         Raises:
-            SameAliasImportViolation
             DottedRawImportViolation
             NestedImportViolation
+            ProtectedModuleViolation,
+            SameAliasImportViolation
+            VagueImportViolation,
```
sobolevn: Extra `,`

wemake_python_styleguide/visitors/ast/imports.py
```diff
@@ -101,9 +123,11 @@ def visit_Import(self, node: ast.Import) -> None:
         Used to find wrong ``import`` statements.
 
         Raises:
-            SameAliasImportViolation
             DottedRawImportViolation
             NestedImportViolation
+            ProtectedModuleViolation,
```
sobolevn: Extra `,`

wemake_python_styleguide/visitors/ast/imports.py
```diff
@@ -27,7 +28,7 @@
 ErrorCallback = Callable[[BaseViolation], None]  # TODO: alias and move
 
 
-@final
+@final  # noqa: WPS214
```
sobolevn: Can we do something with this violation here?

Can we refactor this class to be more simple? Or just split it?

skarzi: In the latest [commit](https://github.com/wemake-services/wemake-python-styleguide/pull/1118/commits/f8dc24e45cc18d0d810797c0b9052e2d0c67698d) I splitted this validator into 3 classes:

+ `ast.Import` validator
+ `ast.ImportFrom` validator
+ base class for both above mentioned validators

If you have any ideas how to make it clearer please let me know.

wemake_python_styleguide/visitors/ast/imports.py
```diff
@@ -73,60 +62,120 @@ def check_alias(self, node: AnyImport) -> None:
...
-    def check_protected_import(self, node: AnyImport) -> None:
-        import_names = [alias.name for alias in node.names]
-        for name in chain(imports.get_import_parts(node), import_names):
+
+@final
+class _ImportValidator(_BaseImportValidator):
+    """Validator of ``ast.Import`` nodes."""
+
+    def validate(self, node: AnyImport) -> None:
+        node = cast(ast.Import, node)
```
skarzi: I dislike this casting :D
Do you know if there is any better way to type such strict subclasses which provides implementation only for one type of the type from `Union` defined in base class?

wemake_python_styleguide/visitors/ast/imports.py
```diff
@@ -74,39 +62,74 @@ def check_alias(self, node: AnyImport) -> None:
...
-        )
-        self._check_protected_names(
-            [alias.name for alias in node.names],
-            ProtectedModuleMemberViolation(node),
+@final
+class _ImportValidator(_BaseImportValidator):
+    """Validator of ``ast.Import`` nodes."""
+
+    def validate(self, node: AnyImport) -> None:
+        node = cast(ast.Import, node)
```
sobolevn: We try not use `cast`. It should be a bug in Python typing to actually use it.
Can you please refactor this place once again?

skarzi: Sure, but to do that I need to use everywhere `AnyImport` instead of concrete import type that is validating in given method, is that ok?

e,g `def _check_dotted_raw_import(self, node: AnyImport) -> None:` instead of `def _check_dotted_raw_import(self, node: ast.Import) -> None:`

wemake_python_styleguide/visitors/ast/imports.py
```diff
@@ -39,30 +40,18 @@ def __init__(
         self._error_callback = error_callback
         self._options = options
 
-    def check_nested_import(self, node: AnyImport) -> None:
+    @abc.abstractmethod
+    def validate(self, node: AnyImport):
```
sobolevn: `-> None`

sobolevn: We can just call this method differently. Like `_validate_any_import` and then call it in specific `.validate` method instead of `super().validate()`

skarzi: Sure, but then we need to resign from declaring `.validate()` as abstractmethod and just provide this method in 2 concrete validators to satisfy mypy, is that what you suggest?

sobolevn: Yes 👍

skarzi:
could you restart travis build? from some reason it failed on downloading poetry

PR #1122, opened 2020-01-25
use abc stdlib module to mark abstract classes

### LINKED ISSUE #1121: Make abstract classes abstract
There are several classes that are not marked as `abstract`. But, they should be!

Like:
- https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/visitors/base.py
- https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/violations/base.py

We need to import `abc` and use `metaclass=ABCMeta` for all of them.

skarzi:
# I have made things!

## Checklist

## Related issues

Closes #1121 

### DIFF
--- a/tests/test_visitors/test_base.py
+++ b/tests/test_visitors/test_base.py
@@ -2,30 +2,21 @@
 
 from unittest.mock import MagicMock
 
-import pytest
-
 from wemake_python_styleguide import constants
-from wemake_python_styleguide.visitors.base import (
-    BaseFilenameVisitor,
-    BaseVisitor,
-)
+from wemake_python_styleguide.visitors.base import BaseFilenameVisitor
 
 
-def test_visitor_raises_not_implemented(default_options):
-    """Ensures that `BaseChecker` raises `NotImplementedError`."""
-    with pytest.raises(NotImplementedError):
-        BaseVisitor(default_options).run()
-
-
-def test_base_filename_raises_not_implemented(default_options):
-    """Ensures that `BaseFilenameVisitor` raises `NotImplementedError`."""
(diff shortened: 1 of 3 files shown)


skarzi:
`abc.ABC` is just small helper - it is `object`'s subclass that use `abc.ABCMeta` metaclass, so it just makes most common abstract classes definitions more concise.
Generally any class that subclass some ABC class and do NOT override **all** abstract methods is also ABC class, that's the reason I haven't add `metaclass=abc.ABCMeta` to [`BaseFilenameVisitor`](https://github.com/wemake-services/wemake-python-styleguide/blob/53b98be42f8e8be9993c62b3f856478f9dc5ea3c/wemake_python_styleguide/visitors/base.py#L182) however I have used `@abc.abstractmethod` decorator on `visit_filename()`.

We can add `metaclass=abc.ABCMeta` to other classes, however they aren't real ABC classes, because they already implement `run()` method (which is the only abstract method they need to implement), so it will be just an information to developers

skarzi:
Similar story with all violations' base classes, but there even `BaseViolation` is not a real ABC class, because it doesn't declara any abstract method, however still we can make its class attributes abstract by using 2nd or 3rd solution that I have mentioned in issue #1121 [comment](https://github.com/wemake-services/wemake-python-styleguide/issues/1121#issuecomment-578401954)

sobolevn:
> so it will be just an information to developers
Yes, please! That was the main idea behind this change.
Sometimes developers what to see what classes are *base* ones (that how this task was originally born). And this (I hope) is going to help them.

sobolevn:
Thanks a lot! 😄

Issue #1099, opened 2020-01-06
WPS332 Found implicit ternary expression on boolean expression


sponsfreixes:
After reading https://wemake-python-stylegui.de/en/latest/pages/usage/violations/consistency.html#wemake_python_styleguide.violations.consistency.ImplicitTernaryViolation, I don't understand why the following code:
```python
if first_condition and second_condition or third_condition:
    do_stuff()
```
triggers WPS332. The closest related issue I could find is https://github.com/wemake-services/wemake-python-styleguide/issues/604. On that example, it is clear that it's a ternary expression to make an assignment. But applying that same reasoning to any boolean logic that combines `and`s and `or`s like the example from above does not seem adequate to me.

The only way I found around it is with the construct:
```python
trigger_do_stuff = third_condition if first_condition else second_condition
if trigger_do_stuff:
    do_stuff()
```
Which strikes me as confusing and not readable.

I would appreciate an explanation of how that boolean logic should be correctly expressed according to WPS, or if this case shouldn't be flagged by 332 (on that case, I would open a bug ticket).

sponsfreixes:
I've been thinking a bit more about it, and found about example. It feels incorrect that:
```
if first_condition and second_condition or third_condition:
    do_stuff()
```
raises WPS332 but
```python
if first_condition and second_condition or third_condition or fourth_condition:
    do_stuff()
```
is OK.

On both cases we are just applying basic boolean operands.  Having 3 terms raises that error but having more than 3 does not.

sobolevn:
This is a bug. This code should not raise any violations:

```python
if first_condition and second_condition or third_condition:
    do_stuff()
```

We must check that our potential "implicit ternary expression" is not inside `if` condition. 

Are you willing to fix it? Should be rather simple!

sponsfreixes:
I'll give it a try on the following weeks!

sobolevn:
Awesome, thanks!

sponsfreixes:
I've been thinking about the right way of telling if an expression is used as ternary expression or boolean logic. Only checking for an `if` would mean that this is correct:
```python
if (product_too_expensive and we_are_low_on_budget) or we_have_no_money:
    reject_product()
```

But this would we marked as incorrect:
```python
we_cant_afford_product = (product_too_expensive and we_are_low_on_budget) or we_have_no_money
if we_cant_afford_product:
    reject_product()
```
[...]

PR #1126, opened 2020-01-26
Remove ImplicitTernaryViolation - WPS332 (#1099)

sponsfreixes:
We are removing this rule because we found out that it was not
applicable to all cases.

This is the first time we remove an existing rule, so a mechanism to
keep track of them and satisfy test_no_holes was needed. The implemented
mechanism involves just adding the deprecated code to the
DEPRECATED_CODES tuple, which must be defined on the file where the
violation class lived. This variable is not defined at all if there are
no deprecated codes on the file.

# I have made things!

## Checklist

## Related issues


sobolevn:
Related #1099

Issue #1125, opened 2020-01-26
Add `__call__` to list of methods that should be on top


sobolevn:
I guess that the order should be the following:

- `__new__`
- `__init__`
- `__call__`

Things to edit:
1. https://github.com/wemake-services/wemake-python-styleguide/blob/06e94265bc27263b87d58a640a08435bae96d47c/wemake_python_styleguide/violations/consistency.py#L1516
2. https://github.com/wemake-services/wemake-python-styleguide/blob/e84e7671aad08a5acc87646f1c86f20cd65ce400/wemake_python_styleguide/visitors/ast/classes.py#L410
3. https://github.com/wemake-services/wemake-python-styleguide/blob/f5447577113bd4c7c397a0e7970a62908bd11f73/tests/test_visitors/test_ast/test_classes/test_method_order.py

gillesigot:
May I take it?

sobolevn:
Sure! Thanks a lot!

sobolevn:
Closed by https://github.com/wemake-services/wemake-python-styleguide/pull/1129

PR #1129, opened 2020-01-30
Add `__call__` to list of methods that should be on top #1125

gillesigot:
# I have made things!

## Checklist

## Related issues
* Too many returns for the` _ideal_order` method (classes.py), had to refactor a bit...

### DIFF
--- a/wemake_python_styleguide/visitors/ast/classes.py
+++ b/wemake_python_styleguide/visitors/ast/classes.py
@@ -412,10 +412,10 @@ class ClassMethodOrderVisitor(base.BaseNodeVisitor):
             '__init__': 4,
             '__call__': 3,
         }
-        public_magic_methods_priority = 2
+        public_and_magic_methods_priority = 2
 
         if access.is_protected(first):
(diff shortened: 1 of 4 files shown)


wemake_python_styleguide/visitors/ast/classes.py
```diff
@@ -407,12 +407,15 @@ def _check_method_order(self, node: ast.ClassDef) -> None:
...
+        base_methods_order = {
+            '__new__': 5,  # highest priority
+            '__init__': 4,
+            '__call__': 3,
+        }
         if access.is_protected(first):
             return 1
         if access.is_private(first):
             return 0  # lowest priority
+        if first in base_methods_order.keys():
```
skarzi: this `if`, however it's totally correct, can be avoided:

```python
return base_methods_order.get(first, 2)
```

Also some constant can be used instead of number `2` to increase readability

sobolevn: 👍

PR #1133, opened 2020-02-02
lru_cache variable_names_blacklist_from() logic function

### LINKED ISSUE #1130: cache all things dependent on configuration options
Hi,

in #1129 domain's names options that allow developers to modify variable names' blacklist were introduced. Construction of this blacklist can be performed only once, because it depends only on some constants and configuration options, that are also static (loaded from CLI flags or ini config), so there is no need to compute them for every file being checked.
We already have discussed using `lru_cache` decorator on `variable_names_blacklist_from` function, but because of my misunderstanding of flake8 and `Checker` lifecycle I haven't used it.
I have `grep`ped `logic` module to check if there is any other usage of configuration options that can be cached, but I haven't found anything else.
Can I add `lru_cache` decorator this time? Well, better late than never :smile:  

Regards,
Łukasz

skarzi:
# I have made things!

## Checklist

## Related issues

Closes #1130 

To make using [`lru_cache` decorator](https://docs.python.org/3/library/functools.html#functools.lru_cache) on `variable_names_blacklist_from()` possible - `_ValidatedOptions` has to be hashable, which implies its immutability (I've used [`attrs` `frozen`](https://www.attrs.org/en/stable/examples.html#immutability) argument) that btw is really expected in this case, because configuration options shouldn't be changed after they are loaded and validated.
This change also required updating `ConfigurationOptions` protocol - all attributes need to be unsettable, because  `_ValidatedOptions` is now frozen/immutable.

If you dislike current solution (personally I don't like new definition of `ConfigurationOptions` protocol, however according to few issues I have found on `mypy` repo, it's the only way to define unsettable protocol's attributes) we can change current implementation of `variable_names_blacklist_from()` a bit, so it will accept `allowed_domain_names` and `forbidden_domain_names` immutable arguments instead of `_ValidatedOptions`, however then `_ValidatedOptions` will be defined as mutable.

___

### DIFF
--- a/wemake_python_styleguide/types.py
+++ b/wemake_python_styleguide/types.py
@@ -38,7 +38,7 @@ Reference
 """
 
 import ast
-from typing import List, Tuple, Type, Union
+from typing import Tuple, Type, Union
 
 from typing_extensions import Protocol, final
 
@@ -83,7 +83,7 @@ AnyAccess = Union[
 ]
 
 
-@final
+@final  # noqa: WPS214
 class ConfigurationOptions(Protocol):
     """
     Provides structure for the options we use in our checker and visitors.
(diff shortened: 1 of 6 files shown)


wemake_python_styleguide/types.py
```diff
@@ -98,35 +98,93 @@ class or structure.
...
 
     # General:
-    min_name_length: int
-    i_control_code: bool
-    max_name_length: int
-    max_noqa_comments: int
-    nested_classes_whitelist: List[str]
-    allowed_domain_names: List[str]
-    forbidden_domain_names: List[str]
+    @property
```
sobolevn: Why do we need a `@property` decorator here?

skarzi: Because of following error:

```
wemake_python_styleguide/checker.py:139: error: Incompatible types in assignment (expression has type "_ValidatedOptions", variable has type "ConfigurationOptions")
wemake_python_styleguide/checker.py:139: note: Protocol member ConfigurationOptions.max_call_level expected settable variable, got read-only attribute
```

By default all protocol's attributes are settable and according to few issues on [`mypy` repo](https://github.com/python/mypy/issues/6002#issuecomment-550421647) or [`typing` repo](https://github.com/python/typing/issues/622) the only wayto define unsettable protocol's attribute is to use `@property` without defining `setter` and/or `deleter`.

Why `ConfigurationOptions` protocol was introduced?
Doesn't it possible to make `_ValidatedOptions` public and use it in all annotations instead?

sobolevn: This is really strange. Why this error happens now? (I am asking because I cannot tell the reason from the code changes)

Maybe because of the `frozen=True`?

skarzi: Yes, it's because of `frozen=True`
I wrote some explanation in issue description, but in short, to use `lru_cache` any function's argument must be hashable (so also immutable, because otherwise it would be really inconsistent and error-prone) and I have used `frozen=True` to make `_ValidatedOptions` instances immutable and because protocol `ConfigurationOptions` is used as type annotation for `_ValidatedOptions` instances it had to be changed to be compatible with now immutable `_ValidatedOptions`

sobolevn: Got it, thanks!

sobolevn:
Thanks a lot! This is cool! 

Do you think that there are possibly other places where caching can be applied?

skarzi:
> Do you think that there are possibly other places where caching can be applied?

I haven't found anything else after quick look on `logic/` and `visitors/` modules.
But I will remember about checking that during new features/fixes implementation :+1:

PR #1147, opened 2020-02-11
Issue #1140

### LINKED ISSUE #1068: Consider using libcst
We can try it: https://github.com/Instagram/LibCST

This might solve a lot of our problems.
Like:
1. New lines in brackets
2. Brackets around operators (like #385 and #214 and #454)

### LINKED ISSUE #1140: Forbid to use `some .method()`
# Rule request

## Thesis

Python shocks me everyday. I found out (the hard way) that one can write `some .method()` and even `some . method()`

Here's what `WPS` says:

```
» flake8 ex.py

ex.py

  1:1      C101  Coding magic comment not found
  some .set(1, 2)
  ^

  1:1      F821  undefined name 'some'
  some .set(1, 2)
  ^

```

This is insane!

Original issue: https://github.com/django/django/pull/12426

This might be a consistency rule. I guess that we should use `tokenize` or `libcst` for this.

PrVrSs:
# I have made things!

## Checklist

## Related issues

- Closes #1068
- Closes #1140


wemake_python_styleguide/checker.py
```diff
@@ -118,6 +123,23 @@ def __init__(
...
         self.filename = filename
         self.file_tokens = file_tokens
+        self.cst = self.build_cst()
+
+    def read_py3(self) -> str:
+        """Read the source code."""
+        with tokenize.open(self.filename) as src_code:
+            return src_code.read()
+
+    def build_cst(self) -> libcst.Module:
```
sobolevn: Can and should we cache this call?

wemake_python_styleguide/checker.py
```diff
@@ -118,6 +123,23 @@ def __init__(
...
+        """Build CST from source code."""
+        if self.filename in frozenset(('stdin', '-', None)):
+            src = utils.stdin_get_value()
+        elif pathlib.Path(self.filename).is_file():
+            src = self.read_py3()
+        else:
+            return libcst.parse_module('')
```
sobolevn: should be `src = ''`

wemake_python_styleguide/visitors/base.py
```diff
@@ -271,3 +276,45 @@ def run(self) -> None:
...
+
+    This class should be used as a base class for all ``cst`` based checkers.
+    Method ``visit()`` is defined in ``CSTNode`` class.
+
+    Attributes:
+        tree: ``libcst.Module`` tree to be checked.
+
+    """
+
+    METADATA_DEPENDENCIES = (PositionProvider,)  # noqa: WPS115
```
sobolevn: Can you please document and type this attribute?

wemake_python_styleguide/visitors/cst/attributes.py
```diff
@@ -0,0 +1,49 @@
+# -*- coding: utf-8 -*-
+
+from libcst import (
```
sobolevn: Let's stick to the same convention as we use in `ast`.
We do `import ast`.

So, let's use `import libcst` instead

pyproject.toml
```diff
@@ -80,6 +80,7 @@ flake8-rst-docstrings = "^0.0.12"
 flake8-executable = "^2.0"
 pep8-naming = "^0.9.1"
 darglint = "^1.1"
+libcst = "^0.3.1"
```
sobolevn: One more thing: can you please move this dependency at the very top? Near `astor`

wemake_python_styleguide/checker.py
PrVrSs: hm, I can do that
```
@lru_cache(None)
def build_cst(self, filename) -> libcst.Module: ...
```
but checker is created for every files and I didn't found that flake8 cached ast.(only use git for find diff) What do you think about it?

pyproject.toml
PrVrSs: done

wemake_python_styleguide/visitors/cst/attributes.py
PrVrSs: done

wemake_python_styleguide/visitors/base.py
PrVrSs: done

wemake_python_styleguide/checker.py
PrVrSs: done

wemake_python_styleguide/checker.py
sobolevn: So, we call this function once. Ok, no action is required. Thanks!

wemake_python_styleguide/visitors/base.py
```diff
@@ -271,3 +277,50 @@ def run(self) -> None:
...
+class BaseCSTVisitor(BaseVisitor, CSTVisitor, metaclass=abc.ABCMeta):
+    """
+    Allows to store violations while traversing node cst.
+
+    This class should be used as a base class for all ``cst`` based checkers.
+    Method ``visit()`` is defined in ``CSTNode`` class.
+
+    Attributes:
+        tree: ``libcst.Module`` tree to be checked.
+        METADATA_DEPENDENCIES: ``Collection[ProviderT]`` metadata interface
```
uhbif19, line 291: As far as I understand, the class inherits both libcst and WPS visitors. I think here it is not clear, that METADATA_DEPENDENCIES is libcst functionality, which is used implicitly by `self.tree.visit`, and this better be noted.

uhbif19:
Very cool stuff.

wemake_python_styleguide/visitors/base.py
PrVrSs: I suggest we discuss it. why not clear. `BaseCSTVisitor` inherits `BaseVisitor` and `BaseVisitor` doesn't know anything about `self.tree`. But mb should be noted what used MetadataWrapper. what do u think about next text. 
`Around a ``libcst.Module`` is used ``MetadataWrapper``, that stores associated metadata for that module.`

uhbif19: Suppose somebody not familiar with libcst, tries to make a plugin. I suppose he may not recognize, that `METADATA_DEPENDENCIES` is related with `CSTVisitor` at all.

sobolevn:
Sorry, but we are not going to add `libcst` to this project.
Because:
- `libcst` does not work for 3.9 and 3.10
- future versions will be partially written in Rust, which may cause build issues
- I am writing another tool with libcst support

Thanks a lot for your time and effort!

PR #1157, opened 2020-02-16
add WPS235 violation related to import from members number

### LINKED ISSUE #1151: Add new complexity rule for max imported names for a single module
# Rule request

## Thesis

Currently [`TooManyImportedNamesViolation`](https://wemake-python-stylegui.de/en/latest/pages/usage/violations/complexity.html#wemake_python_styleguide.violations.complexity.TooManyImportedNamesViolation) only works on module level. 

So, you can import up to `50` names from a single package! Like this:

```python
from some import (
    name1,
    name2,
    ...
    name50,
)
```

And it won't be an issue.

This needs to be fixed. I suggest to add `TooManyImportedModuleNamesViolation` that will enforce to have less than `7?` names per each module.

Good: `import some` and `some.name1`, `some.name50`
Bad: example above

This rule needs to be configurable.

skarzi:
# I have made things!

## Checklist

## Related issues

Closes #1151 


wemake_python_styleguide/options/defaults.py
```diff
@@ -112,6 +112,9 @@
 #: Maximum number of call chains.
 MAX_CALL_LEVEL: Final = 3
 
+#: Maximum number of names that can be imported from module.
+MAX_IMPORT_FROM_MEMBERS = 8  # guessed
```
sobolevn: Should be `Final`

wemake_python_styleguide/violations/complexity.py
```diff
@@ -1058,3 +1060,40 @@ class TooLongCallChainViolation(ASTViolation):
...
+        This rule is configurable with ``--max-import-from-members``.
+        Default:
+        :str:`wemake_python_styleguide.options.defaults.MAX_IMPORT_FROM_MEMBERS`
+
+    .. versionadded:: 0.15.0
+
+    """
+
+    error_template = (
+        'Found ``from ... import ...`` with too many imported names: {0}'
```
sobolevn: `Found too many imported names from a module: {0}`

wemake_python_styleguide/visitors/ast/complexity/counts.py
```diff
@@ -81,11 +82,7 @@ def _post_visit(self) -> None:
             )
 
 
-@final
-@alias('visit_any_import', (
-    'visit_ImportFrom',
-    'visit_Import',
-))
+@final  # noqa: WPS214
```
sobolevn: Why do we have a violation here? Can we remove it?

skarzi: there is too many methods in class, because I had to use 2 methods specific to each import node (now `ImportFrom` node has one check more than `Import`).
I can remove it by:

+ using old implementation and call this check only if `isinstance(node, ast.ImportFrom)`
+ move check implementation from separate method to `.visit_ImportFrom()`
+ extract some methods to module level functions/classes

What do you think?

sobolevn: I always prefer `extract some methods to module level functions/classes` if you don't mind.

skarzi: Generally I am also fan of this approach, but in this case every method either use some class' instance attributes (`.options`) or update instance's attributes (`._imports_count` and `_imported_names_count`) and these methods don't have much in common (maybe except `._check_import_count()` and `._check_imported_names_count()`), so I don't have good idea which methods should be extracted.
Maybe I should create `ImportFromMembersNumberValidator` similar to `ImportFromValidator` from `imports` visitor and just use it in `visit_ImportFrom()`, what do you think?

sobolevn: Whatever you think is best! 👍

skarzi: Ok, so now, when I have extracted some logic to new class, another problem appeared - there is to many module members. Can I ignore this violation or should we split `wemake_python_styleguide/visitors/ast/complexity/counts.py` into multiple files?

sobolevn: No, ignoring complexity violations is not an option, sorry: https://sobolevn.me/2019/10/complexity-waterfall

We can split this module:
1. method related counts can be moved to `classes.py`
2. logic can be moved to `logic/`

wemake_python_styleguide/options/defaults.py
```diff
@@ -120,3 +120,6 @@
 
 #: Maximum number of nested annotations.
 MAX_ANN_COMPLEXITY: Final = 3
+
+#: Maximum number of names that can be imported from module.
+MAX_IMPORT_FROM_MEMBERS: Final = 8  # guessed
```
sobolevn: We have set this to `8` by default, by our own code has `20`

What do you think? Should we adjust this value? Or refactor our own code?
I am asking because `8` looks great. But, linter has found new problem in our code that we might fix. Or just ignore.

wemake_python_styleguide/violations/complexity.py
```diff
@@ -1095,3 +1097,38 @@ class TooComplexAnnotationViolation(ASTViolation):
...
+
+
+@final
+class TooManyImportedModuleMembersViolation(ASTViolation):
+    """
+    Forbids ``from ... import ...`` with too many imported names.
+
+    Reasoning:
+        Importing too many names from one import is easy way to cause
+        `WPS203`.
```
sobolevn: We use double backtics.

sobolevn: And I also suggest to add more detailed description: even I don't remember what `WPS203` is 🙂

wemake_python_styleguide/options/defaults.py
skarzi: I have checked some of my personal and work projects and 8 is usually the maximum number of imports from one module.
I have set this value for our project to such big number, because it's library, however now I double checked and there is only 7 `from ... import ...` that import more than 8 members and they are related to:

+ violations - so we can just import namespace and use namespaced violations (imho it's even clearer)
+ ast - same story as with violations
+ typing - again same story, but if you prefer to import everything explicitly from `typing` we can add some option to pass module names that can be skipped for this violation
+ visitors imports in presets - same story like with `typing` imports

What's your opinion?

sobolevn: - violations: I agree, let's do this
- ast: agreed
- typing: yes, I feel like this is a good idea. But quite problematic. One can have a custom module named `typing`, `typing` might have submodules like `.re` or `.decorators`. So we will have to add a lot of configuration for this. I don't like configuration. I guess we can just leave `noqa` comment for this case
- visitors: `noqa` looks like the best option

PR #1185, opened 2020-02-24
Issue #1054 - unPythonicGetterSetterViolation

### LINKED ISSUE #1054: Detect getters and setters
# Rule request

## Thesis

We can inspect class and instance methods to find these pattern: if class contains two methods like `get_X` and `set_X` we should raise `UnpythonicGetterSetterViolation`

We should also detect cases like:

- class has instance or class level attribute `x` or `_x` or `__x`
- class has method called `get_x` or `set_x`

It is `oop` violation.

## Reasoning

Because python does not need this abstraction:
- Either use `@property`
- Or just make the property public and change it

ruwaid4:
# I have made things!

## Checklist

## Related issues
- Closes #1054 


wemake_python_styleguide/violations/oop.py
```diff
@@ -574,3 +576,53 @@ def some_method(self):
...
+            def __init__(self):
+                self.attribute = None
+
+            def set_attribute(self):
+                return self.attribute
+
+            def get_attribute(self, value):
+                self.attribute = value
+
+    .. versionadded:: 0.13.0
```
sobolevn: `.. versionadded:: 0.15.0`

wemake_python_styleguide/violations/oop.py
```diff
@@ -574,3 +576,53 @@ def some_method(self):
...
+
+    See also:
+        https://docs.python.org/3/library/functions.html#property
+        https://stackoverflow.com/a/36943813
+
+    """
+
+    error_template = 'Found getters and setters'
```
sobolevn: `error_template = 'Found unpythonic getter (or setter)'`

tests/fixtures/noqa/noqa.py
```diff
@@ -692,3 +692,130 @@ def consecutive_yields():
...
+
+
+class GetterAndSetterChild(GetterAndSetterParent):
+
+    # What is the intended behavior here?
+    def get_parent_attribute(self):
+        return 1
+
+    def set_parent_attribute(self):
+        return 1
```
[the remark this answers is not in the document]
Gwin73: Where would be a good place to move correct examples, maybe test_logic? I could ofc just remove them.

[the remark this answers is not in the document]
sobolevn: `tests/test_visitors/test_ast/test_classes/`

wemake_python_styleguide/logic/tree/functions.py
```diff
@@ -103,3 +103,11 @@ def is_generator(node: AnyFunctionDef) -> bool:
         if is_contained(node=body_item, to_check=(Yield, YieldFrom)):
             return True
     return False
+
+
+def check_decorators(node: AnyFunctionDef, name: str) -> bool:
+    """Check if name is in decorator name."""
+    for decorator in node.decorator_list:
```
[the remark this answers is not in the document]
ruwaid4: Hard to avoid `line too long` without `too short names`.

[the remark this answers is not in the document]
sobolevn: You can break lines like:

```python
return any(
    name in source.node_to_string(decorator) 
    for ...
)
```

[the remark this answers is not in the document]
ruwaid4: Worked! Actually tried that before but over indented last bracket so it failed.

Gwin73:
Hi @sobolevn, we are done with the requested changes, but I have some trouble with the linter. It gives the `I001 isort found an import in the wrong position` for the imports in my new test class:

```python
# -*- coding: utf-8 -*-

import pytest

from wemake_python_styleguide.visitors.ast.classes import (
    WrongClassVisitor,
)
from wemake_python_styleguide.violations.oop import (
    UnpythonicGetterSetterViolation,
)
...
```
I have tried various orders but with no luck.

ruwaid4:
@sobolevn we have now made the neccesary changes.

tests/test_visitors/test_ast/test_classes/test_getter_setter.py
```diff
@@ -0,0 +1,158 @@
...
+    def set_attribute():
+        ...
+"""
+
+property_getter_and_setter = """
+class Test(object):
+    def __init__(self):
+        self.attribute = 1
+
+    @property
```
sobolevn: I am not sure about this example though. Why? Because the mist pythonic way to do it is:

```python
class Test(object):
    def __init__(self):
        self._attribute = 1

    @property
    def attribute(self):
        ...

    @attribute.setter
    def attribute(self, attribute_value):
        ...
```

Let's forbid this naming even for properties, please.

tests/test_visitors/test_ast/test_classes/test_getter_setter.py
```diff
@@ -0,0 +1,158 @@
...
+
+class TestChild(TestParent):
+    def get_attribute(self):
+        ...
+
+    def set_attribute(self):
+        ...
+"""
+
+nested_getter_and_setter = """
```
sobolevn: Let's also test this case:

```python
"""
class Test(object):
    first: int
    second = 2
    third: int = 3

    def __init__(self):
        self.fifth = 5

    def get_some(self): ...

    def set_other(self): ...
"""
```

tests/test_visitors/test_ast/test_classes/test_getter_setter.py
```diff
@@ -0,0 +1,158 @@
...
+    @property
+    def get_attribute(self):
+        ...
+
+    @property.setter
+    def set_attribute(self):
+        ...
+"""
+
+child_getter_and_setter = """
```
sobolevn: Let's also test this case:

```python
"""
class Test(object):
    def __init__(self):
        self.attribute = 1
        self.set_attribute(2)
"""
```

wemake_python_styleguide/violations/oop.py
```diff
@@ -574,3 +576,45 @@ def some_method(self):
...
+        class Example(object):
+            def __init__(self):
+                self._attribute = None
+
+        # Wrong:
+        class Example(object):
+            def __init__(self):
+                self.attribute = None
+
+            def set_attribute(self):
```
sobolevn: `get` and `set` can have just `...` as implementation, because it does not matter.

tests/test_visitors/test_ast/test_classes/test_getter_setter.py
```diff
@@ -0,0 +1,158 @@
...
+            ...
+        def set_attribute(self):
+            ...
+        get_attribute(self)
+"""
+
+class_attribute_template = """
+class Template(object):
+    def __init__(self):
+        self.{0} = 1
```
sobolevn: We also need to cover class attributes here, in forms of:

- AnnAssign like `attribute: int`
- AnnAssign like `attribute: int = 1`
- Assign like `attribute = 1`
- Assign like `attribute = other = 1`
- Assign like `attribute, other = 1, 2`

Instance cases not covered:

- AnnAssign like `self.attribute: int = 1`
- Assign like `self.attribute = self.other = 1`
- Assign like `self.attribute, self.other = 1, 2`

We also need to test every case with public, protected, and private access modificators.

wemake_python_styleguide/visitors/ast/classes.py
```diff
@@ -117,6 +119,24 @@ def _is_correct_base_class(self, base_class: ast.AST) -> bool:
...
+        class_attributes, instance_attributes = classes.get_attributes(node)
+        attribute_names = set(
+            name_nodes.flat_variable_names(class_attributes),
+        ).union({
+            instance.attr for instance in instance_attributes
+        })
+
+        for attribute in attribute_names:
+            for method_postfix in classes.getter_setter_postfixes(node):
+                if attribute.lstrip('_') == method_postfix:
```
sobolevn: We can strip attribute names earlier. Because we can dedupe them in sets and not make extra loop iterations for cases like `_some`, `some`, and `__some`

tests/test_visitors/test_ast/test_classes/test_getter_setter.py
```diff
@@ -0,0 +1,158 @@
+# -*- coding: utf-8 -*-
+
+import pytest
+
+from wemake_python_styleguide.violations.oop import (
+    UnpythonicGetterSetterViolation,
+)
+from wemake_python_styleguide.visitors.ast.classes import WrongClassVisitor
+
+module_getter_and_setter = """
```
sobolevn: We also need at least one case for `@dataclass`

wemake_python_styleguide/logic/tree/functions.py
```diff
@@ -103,3 +103,11 @@ def is_generator(node: AnyFunctionDef) -> bool:
         if is_contained(node=body_item, to_check=(Yield, YieldFrom)):
             return True
     return False
+
+
+def check_decorators(node: AnyFunctionDef, name: str) -> bool:
+    """Check if name is in decorator name."""
+    return any(
+        name in source.node_to_string(decorator)
```
sobolevn: I also feel like this place can brake. Let's replace `in` with `==`

wemake_python_styleguide/logic/tree/classes.py
```diff
@@ -34,3 +39,54 @@ def is_forbidden_super_class(class_name: Optional[str]) -> bool:
...
+            if is_getter_or_setter(subnode):
+                method_postfixes.add(subnode.name.partition('get_')[2])
+                method_postfixes.add(subnode.name.partition('set_')[2])
+    return method_postfixes
+
+
+def is_getter_or_setter(node: types.AnyFunctionDef) -> bool:
+    """Checks if non property decorated function contains get or set prefix."""
+    if any(node.name.startswith(prefix) for prefix in ('get_', 'set_')):
+        is_property = functions.check_decorators(node, 'property')
```
sobolevn: Since we are replacing `in` with `==` we need to generate a full name here. It is possible?

wemake_python_styleguide/logic/tree/functions.py
ruwaid4: Sure.

wemake_python_styleguide/logic/tree/classes.py
ruwaid4: "property" and ".setter" seems to be the full `name that node_to_string(decorator)` returns.

ruwaid4: But from earlier comments you wrote i was under the impression that decorators do not have to be checked anymore since the naming `get_x` and `set_x` should be forbidden no matter the decorator. So i was thinking about removing the decorator checking part.

wemake_python_styleguide/logic/tree/functions.py
ruwaid4: `attribute.setter` will be the result from `node_to_string(decorator)` if the decorator is `@attribute.settter`. Since the word before the dot `.` can be whatever, it will be hard to find methods with property setter decorator.

ruwaid4: I think `in` should be correct since i have seen others use that for checking decorators:
```
_PROPERTY_EXCEPTIONS: Final = frozenset(('property', '.setter'))
...
def is_property_setter(node: ast.AST) -> bool:
    """Check that function decorated with ``@property.setter``."""
    if isinstance(node, FunctionNodes):
        for decorator in node.decorator_list:
            if node_to_string(decorator) in _PROPERTY_EXCEPTIONS:
                return True
    return False
```

wemake_python_styleguide/logic/tree/classes.py
sobolevn: This is ok 👍

ruwaid4: So i should remove the decorator checking?

sobolevn: Yes, please 👍

tests/test_visitors/test_ast/test_classes/test_getter_setter.py
fwald: I would like to clarify whether or not the following case should be prohibited:
```
"""
@dataclass
class DataClass(object):
    attribute: int

    def get_attribute(self):
        ...

    def set_attribute(self):
        ...
"""
```
As I understand the rule, this should not be allowed?

sobolevn: Yes, this should also be forbidden.

ruwaid4: Attributes like `attribute: int` are of type ast.Load but are they the only ones of type ast.Load? What i am asking is how do we find them? Because they are not of type attribute.

sobolevn: They are `ast.AnnAssign`

ruwaid4: Thank you! Is there a method to print the type of an ast node for next debugging purposes?

sobolevn: See https://github.com/wemake-services/wemake-python-styleguide/blob/master/CONTRIBUTING.md#helpers

ruwaid4:
@sobolevn Here is our progress.

wemake_python_styleguide/logic/tree/classes.py
```diff
@@ -43,39 +45,85 @@ def is_forbidden_super_class(class_name: Optional[str]) -> bool:
...
-            if any(sub.name.startswith(prefix) for prefix in ('get_', 'set_')):
-                method_postfixes.add(sub.name.partition('get_')[2])
-                method_postfixes.add(sub.name.partition('set_')[2])
-    return method_postfixes
+        if isinstance(sub, FunctionNodes) and functions.is_get_set(node, sub):
+            if functions.check_decorator(sub, 'classmethod'):
+                class_method_postfixes.add(sub.name.partition('get_')[2])
+                class_method_postfixes.add(sub.name.partition('set_')[2])
+                continue
+            instance_method_postfixes.add(sub.name.partition('get_')[2])
```
sobolevn: I would better create two variables from `sub.name.partition('get_')[2]` and `sub.name.partition('set_')[2]`

wemake_python_styleguide/visitors/ast/classes.py
```diff
@@ -120,23 +121,27 @@ def _is_correct_base_class(self, base_class: ast.AST) -> bool:
...
-                self.add_violation(
-                    oop.UnpythonicGetterSetterViolation(
-                        node,
-                        text=node.name,
-                    ),
-                )
+        class_attributes, instance_attributes = classes.get_all_attributes_str(
+            node,
+        )
+        class_postfixes, instance_postfixes = classes.get_set_postfixes(node)
```
sobolevn: So, why do we care about this difference here about `class_` and `instance_` attributes? I ask because we end up with the same logic for both of them.

ruwaid4: classmethods can only access class attributes while instance methods can access both and it is why we have to split the checks.

sobolevn: Can you please provide an example when this is useful? I am a bit lost 🙂

ruwaid4:
```
instance_attribute_class_getter_setter = """
class Test(object):
    def __init__(self):
        self.attribute = 1

    @classmethod
    def get_attribute(self):
        ...

    @classmethod
    def set_attribute(self):
        ...
"""
```
@classmethod decorated methods cannot access the instance attribute, so there should be no violation. But my previous code raised a violation here so i realised i needed to split up the checking.

sobolevn: Ok, now I see. Thank you for the explanation. This is an interesting case indeed.

The thing about this is not that this violation can be technically correct. We care about the idealogical part here. And from the idealogical point of view: there's no difference in `@classmethod` or instance methods. They are still java-style getters and setters.

So, I guess we should move towards restricting this example as well.

And combine these two logical conditions into one: with both instance and class attributes.

ruwaid4: Okay, then i will change to the old code and replace the tests to be expected to raise a violation.

ruwaid4: This would be commit: `Refacotr: Remove decorator checking`
(6a1eccc)

sobolevn: 👍

wemake_python_styleguide/visitors/ast/classes.py
```diff
@@ -121,27 +120,23 @@ def _is_correct_base_class(self, base_class: ast.AST) -> bool:
...
-        all_attributes = instance_attributes.union(class_attributes)
-        if variables.name_in_postfix(node, all_attributes, instance_postfixes):
-            self.add_violation(
-                oop.UnpythonicGetterSetterViolation(
-                    node,
-                    text=node.name,
-                ),
-            )
+        class_attributes, instance_attributes = classes.get_all_attributes(node)
+        attribute_names = {
```
sobolevn: I guess we can even more this logic into `get_all_attributes` function. Currently it does not have any logic inside at all.

ruwaid4: That function isnt needed, see my latest commit. I think it should still be plain as it is so that it can be reused later for whatever.

tests/test_visitors/test_ast/test_classes/test_getter_setter.py
```diff
@@ -0,0 +1,381 @@
...
+property_getter_and_setter = """
+class Test(object):
+    def __init__(self):
+        self.attribute = 1
+
+    @property
+    def attribute(self):
+        ...
+
+    @property.setter
```
skarzi:
```suggestion
    @attribute.setter
```

tests/test_visitors/test_ast/test_classes/test_getter_setter.py
```diff
@@ -0,0 +1,381 @@
...
+dataclass_property_getter_setter = """
+@dataclass
+class DataClass(object):
+    attribute: int
+
+    @property
+    def attribute(self):
+        ...
+
+    @property
```
skarzi: name `dataclass_property_getter_setter` suggest that it should be:

```suggestion
    @attribute.setter
```

tests/test_visitors/test_ast/test_classes/test_getter_setter.py
```diff
@@ -0,0 +1,381 @@
...
+dataclass_incorrect_property_getter_setter = """
+@dataclass
+class DataClass(object):
+    attribute: int
+
+    @property
+    def get_attribute(self):
+        ...
+
+    @property
```
skarzi: dataclass name suggests that it should be:

 ```suggestion
    @attribute.setter
```

File: wemake_python_styleguide/logic/tree/classes.py [hunk 20]
```diff
@@ -34,3 +38,76 @@ def is_forbidden_super_class(class_name: Optional[str]) -> bool:
...
+    return class_attributes, instance_attributes
+
+
+def get_class_attributes(
+    node: ast.ClassDef,
+) -> List[Union[ast.Assign, ast.AnnAssign]]:
+    """Returns all non annotated class attributes of a class."""
+    class_attributes = []
+    for sub in ast.walk(node):
+        correct_context = nodes.get_context(sub) == node
```
skarzi: what do you think about renaming `correct_context` to `is_context_correct`?
It sounds more proper for boolean and then next `if` statement can be read like "plain english" sentence

ruwaid4: Sure!

ruwaid4:
@sobolevn Think this is what you asked for 😄

ruwaid4:
@sobolevn You are a great maintaner, thank you for the feedback! Awesome project!

tests/fixtures/noqa/noqa.py
```diff
@@ -693,6 +693,15 @@ def consecutive_yields():
...
 *numbers, = [4, 7]  # noqa: WPS356
 
+
+class AttributeGetter(object):  # noqa: WPS614
+
+    def __init__(self):
+        self.attribute = 1
+
+    def get_attribute(self):
+        return 1
```
sobolevn: Can we replace this line with `anti_wps = ...`? Because `return` is not important here.

ruwaid4: What does `anti_wps` represent? Would the line become:
```
def get_attribute(self):
        anti_wps614 = 1
```
?

wemake_python_styleguide/logic/tree/classes.py
```diff
@@ -34,3 +39,87 @@ def is_forbidden_super_class(class_name: Optional[str]) -> bool:
...
+            class_attributes.append(sub)
+
+    return class_attributes
+
+
+def get_instance_attributes(node: ast.ClassDef) -> List[ast.Attribute]:
+    """Returns all instance attributes of a class."""
+    instance_attributes = []
+    for sub in ast.walk(node):
+        if isinstance(sub, ast.Attribute) and isinstance(sub.ctx, ast.Store):
```
sobolevn: This is missing `'self'` part check. It will also count `other.attr` as instance attribute.

ruwaid4: Should it matter for this violation? The naming should still be disallowed.

sobolevn: Yes, because this code should not produce any violations:

```python
class Some(object):
    def __init__(self, other):
        other.attr = self.some()

   def get_attr(self):
        return something.unrelated()
```

Here `get_attr` technically is not a getter. But this check will fail. See https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/logic/tree/classes.py#L44

ruwaid4: If they are, can you give some example / test cases which i can use?

ruwaid4: Guess i could just use the new implementation.

ruwaid4: Shouldn't the shadowing violation also apply for AnnAssign type attributes?

sobolevn: Nope, shadowing should not report `AnnAssign`:

```python
class Test(object):
    attr: int   # class does not have `attr` in `__dict__` at this point

    def __init__(self):
         self.attr = 1  # now it is in `__dict__` of this instance
```

sobolevn: But! If `AnnAssign` has a value. Than - yes! That's a bug.

Can you please report it?

ruwaid4: sure!

ruwaid4: Hi this seems to have already been checked:
´´´
class_attribute_annotated = """
class ClassWithAttrs(object):
    {0}: int = 0

    def __init__(self) -> None:
        self.{1} = 2
"""
´´´´
This test case raises the shadowing violation as it should. Seems to be working without checking for AnnAssign.

sobolevn: Yes, it is covered indeed 👍

PR #1259, opened 2020-03-16
introduce violation WPS358

### LINKED ISSUE #1256: List should be treated as invalid unpacking targets
# Rule request

## Thesis

Currently, this code passes our linter:

```python
[first, second] = some()
```

But, it is wrong. We need to force:

```python
first, second = some()
```

So, we need to disallow `ast.List` as an unpacking target.

## Reasoning

There's an already a working way of doing unpacking assignment: tuples.
We don't need one more way to do it.

I would say this is a consistency rule.

skarzi:
Add violation WPS358 forbidding unpacking iterables to lists.

# I have made things!

## Checklist

## Related issues

Closes #1256 

### DIFF
--- a/tests/test_visitors/test_ast/test_builtins/test_assign/test_unpacking_rules.py
+++ b/tests/test_visitors/test_ast/test_builtins/test_assign/test_unpacking_rules.py
@@ -17,10 +17,24 @@ tuple_assignment2 = '{0}, second = (1, 2)'
 
 list_assignment = '[first, second] = (1, 2)'
 
+nested_list_assignment0 = '[first, second], third = ((1, 2), 3)'
+nested_list_assignment1 = 'first, [second, third] = (1, (2, 3))'
+nested_list_assignment2 = 'first, (second, [third, fourth]) = (1, (2, (3, 4)))'
+
+multiple_level_nested_list_assignment = (
+    'first, [second, [third, fourth]] = (1, (2, (3, 4)))'
+)
+
 spread_assignment1 = '{0}, *second = [1, 2, 3]'
 spread_assignment2 = 'first, *{0} = [1, 2, 3]'
 
 spread_assignment_in_list = '[first, *rest] = [1, 2, 3]'
+
+nested_spread_assignment_in_list = 'first, [second, *rest] = [1, (2, 3, 4)]'
+
+multiple_level_nested_spread_assign_in_list = (
+    'first, [second, [*rest, last]] = (1, (2, (3, 4, 5)))'
+)
 
 for_assignment = """
 def wrapper():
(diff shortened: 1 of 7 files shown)


tests/test_visitors/test_ast/test_builtins/test_assign/test_unpacking_rules.py
```diff
@@ -12,9 +15,13 @@
 tuple_assignment1 = 'first, {0} = (1, 2)'
 tuple_assignment1 = '{0}, second = (1, 2)'
 
+list_assignment = '[first, second] = (1, 2)'
```
sobolevn: What about cases like these? 

```python
>>> [a, b], c = ((1, 2), 3)
>>> a, [b, c] = (1, (2, 3))
```

wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -274,6 +276,12 @@ def _check_unpacking_targets(
                     best_practices.WrongUnpackingViolation(node),
                 )
 
+    def _check_unpacking_target_type(self, node: ast.Assign) -> None:
+        if isinstance(node.targets[0], ast.List):
```
sobolevn: Turns out we need to check all elements inside `targets` because of:

```python
>>> [a, b], c = ((1, 2), 3)
>>> a, [b, c] = (1, (2, 3))
```

skarzi: that's true. Should we stop detection on first spotted list or raise violations for all?

skarzi: unpacking to nested list actually raise `WPS414` - ` WrongUnpackingViolation`.
Should the second violation be raised?
 
I think we should do one of the following then:

+ update my PR to raise `WPS414` and update `WPS414` violations docstrings (I dislike this, because know `WPS414` can be raised in 2 totally different cases)
+ change name of `WPS414` to something less confusing and more compliant with its docstring title (`Forbids to have tuple unpacking with side-effects.`) and refactor check raising this violation, because unpacking to list doesn't have any side effects if it's not dynamic e.g `a, b[0] = (1, 3)` but `b[0]` is not parsed as `ast.List` by `ast`

What do you think?

sobolevn: I think that these two violations should be independent. And both of them can be raised at the same time.

tests/test_visitors/test_ast/test_builtins/test_assign/test_unpacking_rules.py
```diff
@@ -13,15 +13,29 @@
...
+)
+
 spread_assignment1 = '{0}, *second = [1, 2, 3]'
 spread_assignment2 = 'first, *{0} = [1, 2, 3]'
 
 spread_assignment_in_list = '[first, *rest] = [1, 2, 3]'
 
+nested_spread_assignment_in_list = 'first, [second, *rest] = [1, (2, 3, 4)]'
+
+multiple_level_nested_spread_assign_in_list = (
```
sobolevn: One more case to test: `z = (a, [b, c]) = (1, (2, 3))`

skarzi: I have added support for such cases and much more in new commits.
I am happy that unpacking iterables is not supported when using walrus operator :smile:

sobolevn (approved):

PR #1315, opened 2020-04-06
Add rule for f-string complexity (Issue #1239)

### LINKED ISSUE #1239: Lint `f` strings better!
# Rule request

## Thesis

Currently we just do not allow people to use `f` strings.
But, what if our user just ignores this violation? Now it is possible to do very bad things inside `f` strings and we do not cover it at all.

I suggest to add `TooComplexFormattedStringViolation` that will be throw in case of:
0. `f'{one_name}'` should raise: because we only a single name inside `f` string
1. Too complex expression inside `f` string: like calling a lot of function, making conditions, etc
2. Too many expressions used in `f` string
3. `f` strings that dont have any values inside like: `f'abc'`
4. Things like `assert f'a {value}'` should raise proper violations (`WPS444` in this case), but `assert f'{value}'` should not

This is just a discussion at the moment.

stevensdavid:
# I have made things!
This PR introduces a new rule that performs an additional check on
f-strings. While the previous rule simply disallowed all f-strings,
this new rule allows f-strings that are not too complex. Complexity
is measured by whitelisting what is allowed, namely:

- Value of a variable itself
- Value of a collection through lookup with a Name, Num or Str
- Return value of a call without arguments

Co-authored by @ruwaid4

## Checklist

## Related issues
Closes #1239 


tests/test_visitors/test_ast/test_builtins/test_strings/test_formatted_string.py
```diff
@@ -18,6 +21,16 @@
...
+f_complex_f_string = """
+    f'{reverse(\"url-name\")}?{\"&\".join(\"user=\"+uid for uid in user_ids)}'
+"""
+f_variable_lookup = "f'smth {value}'"
+f_dict_lookup_str_key = "f'smth {dict_value[\"key\"]}'"
+f_list_index_lookup = "f'smth {list_value[0]}'"
+f_function_empty_args = "f'smth {user.get_full_name()}'"
+f_function_with_args = "f'smth {func(arg)}'"
+f_dict_lookup_function_empty_args = "f'smth {dict_value[func()]}'"
+f_list_slice_lookup = "f'smth {list[:]}'"
```
sobolevn, line 33: Can we please also test cases like:
- `some.call().attr`
- `some.attr.call()`
- `some.first[0].attr.other`

tests/test_visitors/test_ast/test_builtins/test_strings/test_formatted_string.py
```diff
@@ -53,3 +65,39 @@ def test_wrong_string(assert_errors, parse_ast_tree, code, default_options):
...
+
+@pytest.mark.parametrize('code', [
+    f_complex_f_string,
+    f_function_with_args,
+    f_dict_lookup_function_empty_args,
+    f_string,
+    f_list_slice_lookup,
+])
+def test_complex_f_string(assert_errors, parse_ast_tree, code, default_options):
+    """Testing that ..."""
```
sobolevn, line 78: Missing docstring entry.

tests/test_visitors/test_ast/test_builtins/test_strings/test_formatted_string.py
```diff
@@ -53,3 +65,39 @@ def test_wrong_string(assert_errors, parse_ast_tree, code, default_options):
...
+
+
+@pytest.mark.parametrize('code', [
+    f_dict_lookup_str_key,
+    f_function_empty_args,
+    f_list_index_lookup,
+    f_variable_lookup,
+])
+def test_simple_f_string(assert_errors, parse_ast_tree, code, default_options):
+    """Testing that ..."""
```
sobolevn, line 97: Also

wemake_python_styleguide/violations/complexity.py
```diff
@@ -1139,3 +1141,48 @@ class TooManyImportedModuleMembersViolation(ASTViolation):
 
     error_template = 'Found too many imported names from a module: {0}'
     code = 235
+
+
+@final
+class TooComplexFormattedStringViolation(ASTViolation):
+    """
+    Forbids f-strings that are more complex .
```
sobolevn, line 1149: `that are more complex` than ...?

wemake_python_styleguide/violations/complexity.py
```diff
@@ -1139,3 +1141,48 @@ class TooManyImportedModuleMembersViolation(ASTViolation):
...
+
+    Reasoning:
+        Complex `f` strings are often difficult to understand,
```
sobolevn, line 1152: We use double backquotes like so:

```
``f``
```

wemake_python_styleguide/violations/complexity.py
```diff
@@ -1139,3 +1141,48 @@ class TooManyImportedModuleMembersViolation(ASTViolation):
...
+        making the code less readible.
+
+    Solution:
+        Use `.format()` or assign complex expressions to variables
```
sobolevn, line 1156: The same about `.format()`.

wemake_python_styleguide/violations/complexity.py
```diff
@@ -1139,3 +1141,48 @@ class TooManyImportedModuleMembersViolation(ASTViolation):
...
+        before formatting.
+
+    Example::
+
+        # Wrong:
+        f'{reverse("url-name")}?{"&".join("user="+uid for uid in user_ids)}'
```
sobolevn, line 1162: Just one correct and wrong example are fine.

wemake_python_styleguide/violations/complexity.py
```diff
@@ -1139,3 +1141,48 @@ class TooManyImportedModuleMembersViolation(ASTViolation):
...
+        f'smth {list_value[0]}'
+        f'smth {user.get_full_name()}'
+        def get_full_name(self):
+            return f'{self.first_name} {self.second_name} {self.last_name}'
+
+    .. versionadded:: 0.16.0
+
+    """
+
+    error_template = 'Found a too complex `f` string'
```
sobolevn, line 1187: We also need to link to `FormattedStringViolation` here. And explain why we do have these two violations that might seem redundant.

wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -124,6 +130,39 @@ def _check_modulo_patterns(
         if self._modulo_string_pattern.search(text_data):
             self.add_violation(consistency.ModuloStringFormatViolation(node))
 
+    def _check_complex_f_string(self, node: ast.JoinedStr) -> None:
+        # Whitelists all simple uses of f strings
+        # Checks if list, dict, function call with no parameters or variable
+        for string_component in node.values:
+            if isinstance(string_component, ast.FormattedValue):
+                if self._valid_f_value(string_component):
```
sobolevn, line 138: We tend to use `_is_valid_...` naming style
I would also suggest to name it `_is_valid_formatted_value`

wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -124,6 +130,39 @@ def _check_modulo_patterns(
...
+        elif isinstance(f_value, ast.Subscript):
+            if self._valid_f_lookup(f_value):
+                return True
+        # Variable lookup is okay
+        elif isinstance(f_value, ast.Name):
+            return True
+        return False
+
+    def _valid_f_lookup(self, subscript: ast.Subscript) -> bool:
+        allowed_types = (ast.Str, ast.Num, ast.Name)
```
sobolevn, line 161: We also tend to move these "enums" to class level.

tests/test_visitors/test_ast/test_builtins/test_strings/test_formatted_string.py
ruwaid4: These should be allowed i guess?

sobolevn: Nope, these all are forbidden.

stevensdavid: So where should the boundary be drawn here? Is all chaining prohibited, do we allow a single chained lookup, or restrict chaining to one call on `self`? E.g.
- `self.attr`
- `self.func()`
- `user.id`

sobolevn: > a single chained lookup

This one resonates the best with me. What do you think?

ruwaid4:
@sobolevn We have made progress but still have some linter errors which will be fixed tomorrow. However we have found no way to write:
`
for _ in range(5):
                chained_parts.append(next(parts))
`
So that it passes the linter. How would you recommendus to do this? We are also wondering if we can add a new class `WrongFormatVisitor`?

Feel free to give feedback on the already pushed changes since the logic is finished.

wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -79,6 +79,9 @@ class WrongStringVisitor(base.BaseNodeVisitor):
         flags=re.X,  # flag to ignore comments and whitespaces.
     )
 
+    _valid_f_lookup_types = (ast.Str, ast.Num, ast.Name)
```
sobolevn, line 82: This needs to be typed: `ClassVar[AnyNodes]`

And please, do not name things like `_f_` it is hard to read.

sobolevn: Use `_formatted_` instead

wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -133,34 +136,90 @@ def _check_modulo_patterns(
...
+                res = False
+            if len(chained_parts) == 4:
+                # If there are 4 elements, they must have exactly 2 of
+                # subscript or call. This is the alternating case fcn().fcn()
+                n_calls = sum(
+                    isinstance(part, self._allowed_chains)
+                    for part in chained_parts
+                )
+                res = n_calls == 2
+            elif len(chained_parts) == 3:
```
sobolevn, line 205: This function is going to be reported as too complex.

ruwaid4: Thank you for the feedback! We are however still struggling with how to rewrite 
```
for _ in range(5):
    chained_parts.append(next(parts))
```
To pass the linter.

wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -478,7 +510,7 @@ def _check_set_elements(
             # Non-constant nodes raise ValueError,
             # unhashables raise TypeError:
             with suppress(ValueError, TypeError):
-                # Similar value:
+                # Similar val:
```
sobolevn, line 513: Was there a problem with this comment? I am asking in fear of any false positives.

stevensdavid:
@sobolevn I believe that we've resolved all the issues you've pointed out so far - thank you for your feedback! Is there anything else that needs to be done with this PR?

stevensdavid:
Ping! @sobolevn

tests/test_visitors/test_ast/test_builtins/test_strings/test_formatted_string.py
```diff
@@ -16,8 +22,33 @@
...
+# Allowed
+f_single_chained_attr = "f'{attr1.attr2}'"
+f_variable_lookup = "f'smth {value}'"
+f_dict_lookup_str_key = "f'smth {dict_value[\"key\"]}'"
+f_list_index_lookup = "f'smth {list_value[0]}'"
+f_function_empty_args = "f'smth {user.get_full_name()}'"
+f_attr_on_function = "f'{fcn().attr}'"
+f_single_chained_functions = "f'{f1().f2()}'"
+f_calling_returned_function = "f'{calling_returned_function()()}'"
+f_double_indexing = "f'{list[0][1]}'"
```
sobolevn, line 34: Should double indexes be allowed? I though we were explicitly allowing only single access operations. 

See discussion at https://github.com/wemake-services/wemake-python-styleguide/pull/1315/files#r405069203

wemake_python_styleguide/violations/complexity.py
```diff
@@ -1139,3 +1141,50 @@ class TooManyImportedModuleMembersViolation(ASTViolation):
...
+    Solution:
+        Use ``.format()`` or assign complex expressions to variables
+        before formatting.
+
+    Example::
+
+        # Wrong:
+        f'{reverse("url-name")}?{"&".join("user="+uid for uid in user_ids)}'
+
+        # Correct:
```
sobolevn, line 1176: Let's swap `Correct` and `Wrong` places (this is how we usually order them).
And leave just a single correct example.

wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -125,6 +124,128 @@ def _check_modulo_patterns(
...
+        """
+        Forbids use of ``f`` strings and too complex ``f`` strings.
+
+        Raises:
+            FormattedStringViolation
+            TooComplexFormattedStringViolation
+
+        """
+        self.add_violation(consistency.FormattedStringViolation(node))
+        self.generic_visit(node)
```
sobolevn, line 156: Should be the last line.

sobolevn: Can you please open a new issue that we have to test it?

wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -125,6 +124,128 @@ def _check_modulo_patterns(
...
+                if self._is_valid_formatted_value(format_value):
+                    continue
+                # Everything else is too complex
+                self.add_violation(
+                    complexity.TooComplexFormattedStringViolation(node),
+                )
+                break
+
+    def _is_valid_formatted_value(self, format_value: ast.AST) -> bool:
+        if isinstance(format_value, (ast.Call, ast.Subscript, ast.Attribute)):
```
sobolevn, line 185: Can be replaced with `self._chainable_types`

wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -125,6 +124,128 @@ def _check_modulo_patterns(
             self.add_violation(consistency.ModuloStringFormatViolation(node))
 
 
+@final
+class WrongFormatStringVisitor(base.BaseNodeVisitor):
+    """Restricts usage of ``f`` strings."""
+
+    _valid_format_index: ClassVar[AnyNodes] = (
+        ast.Str,
```
sobolevn, line 132: Do we allow `Bytes`? Use this alias: https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/compat/aliases.py#L16

wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -125,6 +124,128 @@ def _check_modulo_patterns(
...
+        ast.Num,
+        ast.Name,
```
sobolevn, line 134: What about `False`, `None`, and `True`?

wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -125,6 +124,128 @@ def _check_modulo_patterns(
...
+        parts = iter(attributes.parts(format_value))
+        chained_parts: List[ast.AST] = []
+
+        iteration = 0
+        try:
+            # Should throw on at latest the fifth attempt
+            while iteration < 5:
+                chained_parts.append(next(parts))
+                iteration += 1
+        except StopIteration:
```
sobolevn, line 216: We try not to use exceptions, they are hard to test. Can you please rewrite this part?

wemake_python_styleguide/visitors/ast/builtins.py
ruwaid4: Yes but then it won't pass Mypy since `format_value` will still be considered AST, and `_is_valid_chaining(self, format_value: AnyChainable)` takes AnyChainable as parameter.

Following error will be displayed:
```
wemake_python_styleguide/visitors/ast/builtins.py:186: error: Argument 1 to "_is_valid_chaining" of "WrongFormatStringVisitor" has incompatible type "AST"; expected "Union[Attribute, Subscript, Call]"
```

tests/test_visitors/test_ast/test_builtins/test_strings/test_formatted_string.py
ruwaid4: We debated this and found it readable. But if you prefer this to be disallowed then it will be fixed.

wemake_python_styleguide/visitors/ast/builtins.py
ruwaid4: You mean as indexes? Such as `first[True]`? Is this to be allowed?

wemake_python_styleguide/visitors/ast/builtins.py
sobolevn: This will work if you move `_any_chainable` to module level. And declare it `final`. 
That's a bug in mypy, but we can live with it. Just drop a comment on why it is done this way.

tests/test_visitors/test_ast/test_builtins/test_strings/test_formatted_string.py
sobolevn: Yes, that's true.

ruwaid4: Just to clarify, double indexing and double function calls should NOT be allowed?

wemake_python_styleguide/visitors/ast/builtins.py
sobolevn: I guess so

tests/test_visitors/test_ast/test_builtins/test_strings/test_formatted_string.py
sobolevn: They should not be allowed. Only single level of chaining is fine.

stevensdavid:
@sobolevn Thank you again for the feedback! We've adjusted the code now. Do you have any more comments?

jtpavlock:
Anything left besides the conflicts to get this merged?

sobolevn:
@jtpavlock I will merge this today. Sorry for the long wait. I didn't have much to work on this.

PR #1320, opened 2020-04-08
added __subclasses__ and __mro__ to exclusions of DirectMagicAttributeAccessViolation

### LINKED ISSUE #1314: False positive DirectMagicAttributeAccessViolation with __subclasses__
# Bug report

## What's wrong
```python
class MyClass(object):
    """My class."""

subs = MyClass.__subclasses__()
```

This piece of code raises
`WPS609 Found direct magic attribute usage: __subclasses__`

## How is that should be
`__subclasess__` method should be excluded from the check because it's a class method and can be used outside of class.

Maybe there are other methods that should be excluded?

mattany:
## Checklist

## Related issues

Closes #1314

### DIFF
--- a/tests/test_visitors/test_ast/test_attributes/test_magic_attributes.py
+++ b/tests/test_visitors/test_ast/test_attributes/test_magic_attributes.py
@@ -140,45 +140,7 @@ def test_magic_attribute_is_restricted(
 
 
 @pytest.mark.parametrize('attribute', [
-    'regular',
-    '__doc__',
-    '__name__',
-    '__class__',
-    '__qualname__',
-])
-@pytest.mark.parametrize('code', [
-    magic_attribute_assigned,
-    magic_attribute_accessed,
-    magic_method_called,
-    magic_method_called_params,
-    magic_container_attribute,
-    magic_container_method,
-    magic_callable_attribute,
-])
-def test_regular_attributes(
-    assert_errors,
-    parse_ast_tree,
-    code,
-    attribute,
-    default_options,
-    mode,
-):
-    """Ensures that it is impossible to use magic attributes."""
-    tree = parse_ast_tree(mode(code.format(attribute)))
-
-    visitor = WrongAttributeVisitor(default_options, tree=tree)
-    visitor.run()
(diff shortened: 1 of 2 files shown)


mattany:
I looked at some of the test cases in the file ```test_magic_attributes.py``` and many of them won't compile for sure but do pass the tests without errors. This is because the "method" and "attribute" tests are done in conjunction. Is this on purpose? 
For example in the test ```test_magic_attribute_is_restricted``` the case ```magic_method_called``` will replace {0} with __class__ :
```'some.__class__()'``` 
which throws no style exception.
I'm guessing this is okay because we trust the person to try and run the code before running the linter

mattany:
Also I think that the names of the last two functions in the test are quite misleading: 
```test_regular_attributes``` and ```test_whitelist_magic_attributes_are_allowed``` run on the same ```attribute``` parameter list which includes both ```'regular'``` and the special cases ```__name__``` etc. The difference is only in the code fragments that the functions run on:
-```test_regular_attributes```  runs on the 'wrong' cases.
-```test_whitelist_magic_attributes_are_allowed``` runs on the 'correct' cases.
Both tests pass iff no errors were thrown.

mattany:
> It means that we need to raise errors on both calls and attributes.
> You can rename tests the way you wish! Thanks for the suggestion.

So how should I proceed from here? My understanding is that the exception needs to be thrown under different circumstances. Should I try and fix the place where the exception is called and then update it and update the tests and issue a new pull request?

sobolevn:
@mattany sorry, I am a bit lost in the discussion 🙂 

Why cannot we just add the needed values here: https://github.com/wemake-services/wemake-python-styleguide/blob/58b96219a39390c35353aa9a29d483d5ed3f1cf7/tests/test_visitors/test_ast/test_attributes/test_magic_attributes.py#L181 ?

mattany:
It's no problem to add them there and in the other test, but since we established that the exception doesn't capture some cases, I assumed we would want to fix that first. I guess your right though - it's better to first close this issue first before addressing the next.

PR #1324, opened 2020-04-09
added __subclasses__ and __mro__ to exclusions of DirectMagicAttributeAccessViolation

### LINKED ISSUE #1314: False positive DirectMagicAttributeAccessViolation with __subclasses__

mattany:
## Remarks
Title and:
I added the two new whitelisted attributes to the test suite and also changed it in the following ways:
### Before the changes there were 3 tests:

1. "wrong" access to magic attributes not on the whitelist (should raise exception).
2. "wrong" access to magic attributes on whitelist or regular attributes (shouldn't raise exception).
3. "correct" access to magic attributes on whitelist or regular attributes (shouldn't raise exception).

### After the changes there are 3 tests:
1. "wrong" access to magic attributes not on the whitelist (should raise exception).
2. "correct" access to magic attributes not on the whitelist (should'nt raise exception). **I added this one since it wasn't covered before**
3. access to magic attributes on whitelist or regular attributes (shouldn't raise exception). **I merged tests 2. and 3. from the old test to make this one since the contents are identical.**

## Checklist

## Related issues

Closes #1314

### DIFF
--- a/tests/test_visitors/test_ast/test_attributes/test_magic_attributes.py
+++ b/tests/test_visitors/test_ast/test_attributes/test_magic_attributes.py
@@ -157,7 +157,6 @@ def test_magic_attribute_is_restricted(
     magic_super_cls_attribute,
     magic_super_cls_method,
 ])
-
 def test_magic_attribute_correct_contexts(
     assert_errors,
     parse_ast_tree,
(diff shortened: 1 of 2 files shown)


sobolevn:
@mattany I have fixed the CI, can you please rebase your PR? So we can check that everything works now. 👍

sobolevn:
🎉 🎉 🎉

PR #1419, opened 2020-05-25
Enable Deepsource

RJ722:
# Add DeepSource

~~Add `.deepsource.toml` to run Continuous Quality analysis on the repo with [DeepSource](https://deepsource.io). Upon enabling DeepSource, quality analysis will be run on every PR to detect 500+ types of problems in the changes — including bug risks, anti-patterns, security vulnerabilities, etc.~~

I just checked that this repository already had DeepSource enabled in the past. My bad, closing this PR!

Would you have any feedback for us? Is there something we can improve?

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,0 +1,12 @@
+version = 1
+
+[[analyzers]]
+name = "python"
+enabled = true
+
+  [analyzers.meta]
+  runtime_version = "3.x.x"
+
+[[analyzers]]
+name = "docker"
+enabled = true


sobolevn:
I know that Deepsource uses a bunch of rules from `wemake-python-styleguide` inside. And tools like `pylint`, `flake8`, etc.

1. We can always add these tools to the CI directly rather than using 3rd party service. 
2. All things that we really care about we cover with our own rules in the core product
3. (personal, but you have asked a Russian for a feedback, so you will get an honest one 😆) I don't like your business model. Your service is based on lots of opensource tools other people are working on for free. AFAIK, DeepSource didn't support these project in a financial way. Correct me if I am wrong. Projects like `pylint` and `flake8` really deserve a support, they are great and serve our Python community for many years.

PR #1432, opened 2020-05-30
Issue 1406: Allow unused variable in for loop iterator

### LINKED ISSUE #1406: WPS122 should allow a no-op loop var
# Bug report

## What's wrong

Imagine I want to do something 3 times but I don't care about the attempt number. So I'd normally write something like this:
```python
for _ in range(3):
    do_something()
```

But then the linter complains with `WPS122 Found all unused variables definition: _` which is incorrect. While things like assignment allow to fix this with
```diff
-_something_unused = get_useless_result()
+get_useless_result()
```
it is impossible with for-loops not to define this loop variable at all.

## How is that should be

This case shouldn't be a violation of the given rule.

## System information

N/A

nn3un:
# I have made things!

## Checklist

   (Moved a test case from unallowed to allowed)
   (The [documentation](https://wemake-python-stylegui.de/en/latest/pages/usage/violations/naming.html#wemake_python_styleguide.violations.naming.UnusedVariableIsDefinedViolation) makes no mention of this specific case)

## Related issues
Closes #1406

(The issue description talks about comprehensions, but I didn't find any place,including the [comprehension tests](https://github.com/wemake-services/wemake-python-styleguide/tree/58b96219a39390c35353aa9a29d483d5ed3f1cf7/tests/test_visitors/test_ast/test_loops/test_comprehensions), where comprehensions were being checked for WPS122, and manually testing code such as `my_list = [3 for _ in range(4)]` was throwing no WPS122 errors, so I ended up doing nothing for that)

### DIFF
--- a/tests/test_visitors/test_ast/test_naming/conftest.py
+++ b/tests/test_visitors/test_ast/test_naming/conftest.py
@@ -214,27 +214,38 @@ _FORBIDDEN_UNUSED_TUPLE = frozenset((
     for_variable,
 ))
 
-_FORBIDDEN_UNUSED = _FORBIDDEN_UNUSED_TUPLE | {
+# Raw unused variables return True for logic.naming.access.is_unused().
+# Example: _, __.
+# Protected unused variables return True for logic.naming.access.is_protected().
+# Example: _protected.
+_FORBIDDEN_BOTH_RAW_AND_PROTECTED_UNUSED = frozenset((
+    unpacking_variables,
+    variable_def,
+    with_variable,
     variable_typed_def,
     variable_typed,
     exception,
-}
+))
 
 if PY38:
-    _FORBIDDEN_UNUSED |= {
+    _FORBIDDEN_BOTH_RAW_AND_PROTECTED_UNUSED |= {
         assignment_expression,
     }
+
+_FORBIDDEN_RAW_UNUSED = _FORBIDDEN_BOTH_RAW_AND_PROTECTED_UNUSED | {
+    static_attribute,
+    static_typed_attribute,
+    static_typed_annotation,
+}
+
+_FORBIDDEN_PROTECTED_UNUSED = _FORBIDDEN_BOTH_RAW_AND_PROTECTED_UNUSED | {
+    for_variable,
+}
 
 
 @pytest.fixture(params=_ALL_FIXTURES)
 def naming_template(request):
     """Parametrized fixture that contains all possible naming templates."""
-    return request.param
-
-
-@pytest.fixture(params=_FORBIDDEN_UNUSED)
-def forbidden_unused_template(request):
-    """Returns template that can be used to define wrong unused variables."""
     return request.param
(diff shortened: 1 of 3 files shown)


tests/test_visitors/test_ast/test_naming/conftest.py
```diff
@@ -211,7 +211,6 @@ def container():
     unpacking_variables,
     variable_def,
     with_variable,
-    for_variable,
```
sobolevn, line 214: Can you please add a test that explicitly shows that this is now allowed?

nn3un:
Added tests that test the complement of _FORBIDDEN_RAW_UNUSED_TEMPLATE, which includes for loops. So now tests run for: 

```
for _ in []:
for __ in []:
for _unused in []:
```
in method, function, and module contexts.

(Please let me know if the docstrings need to be improved.)

sobolevn:
@nn3un here's when I start to see a problem:
1. `for _ in []:` is allowed
2. `for __ in []:` is not allowed
3. `for _unused in []:` is not allowed

nn3un:
> @nn3un here's when I start to see a problem:
> 
> 1. `for _ in []:` is allowed
> 2. `for __ in []:` is not allowed
> 3. `for _unused in []:` is not allowed

Oh, I didn't realize that. 
- `for __ in range(n)` still raises  ` WPS123 Found wrong unused variable name: __`. So is it ok if it doesn't raise WPS122? I think it will make the code a bit cleaner, since I can call [access.is_unused](https://github.com/wemake-services/wemake-python-styleguide/blob/256de41bebbe53f60cd5ce87cb4031f120eb9615/wemake_python_styleguide/logic/naming/access.py#L9-L26), which returns `True` for both _ and __.

- Should `for _, _ in result:` raise error?

sobolevn:
Yes to both questions 👍

nn3un:
Sorry for the delay. I tried different ways to change the tests to accommodate for the change, and the best I could think of required a decent bit of restructuring. 

My basic idea was unused variables are of two kinds: raw (`_`,`__`) and protected (`_protected`). So as long as I checked for raw forbidden, raw allowed, protected forbidden, and protected allowed, I checked for all unused definition use cases.

Git makes the `test_unused_defintion.py` changes a bit complex to understand so here’s a summary of what I did:

The 2 tuple tests, `test_used_variable_definition` and `test_raw_unused_variable_definition` remain unchanged (with slight comment changes and renaming)

Deleted `test_unused_variable_definition_allowed`, which is now covered by `test_protected_unused_variable_definition_allowed`, and `test_raw_unused_variable_definition_allowed`. 

`test_unused_variable_definition_allowed` is now `test_protected_unused_var_definition_allowed`. 

Added `test_raw_unused_variable_definition_allowed` (which tests for the no-op loop var)

nn3un:
Hurrah! Thank you @sobolevn for all the help and guidance!

PR #1439, opened 2020-06-02
Add example with suppress to UselessContinueViolation

### LINKED ISSUE #1438: False positive UselessContinueViolation
# Bug report

## What's wrong

```python
def func(obj_iterable) -> Optional[str]:
    for some_obj in obj_iterable:
        try:
            return do_smth(some_obj)
        except Exception:
            continue
    return None

WPS327 Found useless `continue` at the end of the loop [wemake-python-styleguide]
  for some_obj in obj_iterable:
  ^
```

## How is that should be

Not sure is it a bug or on purpose?

For me, it's valid function if I don't care about intermediate results of `do_smth(some_obj)`

AlwxSin:
Closes #1438

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -14,9 +14,36 @@
 import os
 import sys
 
+import sphinx
 import tomlkit
 
 sys.path.insert(0, os.path.abspath('..'))
+
+
+# -- Monkeypatches -----------------------------------------------------------
+
+def _monkeypatch(cls):
+    """Decorator to monkey-patch methods."""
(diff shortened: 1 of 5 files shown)


sobolevn:
Can you please also fix the CI? https://github.com/dry-python/returns/commit/e2e9905a66096f1474db64ea1b9990739b3539b9

AlwxSin:
@sobolevn looks like fixed

sobolevn:
Thanks a lot! 🍻

Issue #1448, opened 2020-06-05
[docs] Violation set of pages is broken: code numbers missing


webknjaz:
# Bug report

I used to look up rule explanations by searching the rule pages with `Ctrl+F`.

## What's wrong

The rule numbers/codes are not rendered in docs anymore.

## How is that should be

It should be possible to see the WPS codes on pages like https://wemake-python-stylegui.de/en/latest/pages/usage/violations/naming.html

sobolevn:
This happened because of `sphinx` update to `3.0` 😞

Dreamsorcerer:
I don't have it set up to test out. But, looking at the documentation, I found:
> Setting None or True to the value is equivalent to giving only the option name to the directives.
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_default_options

So, maybe just changing this line to `True` will do the trick?
https://github.com/wemake-services/wemake-python-styleguide/blob/master/docs/conf.py#L113

Also, it should be changed to `autodoc_default_options`, as that documentation (in the paragraph above) says `autodoc_default_flags` is deprecated.

sobolevn:
Fixed!
<img width="1083" alt="Снимок экрана 2020-06-07 в 20 41 49" src="https://user-images.githubusercontent.com/4660275/83975930-54bcaf00-a8ff-11ea-8db3-a200c600e4ac.png">

PR #1449, opened 2020-06-05
Fix a missing "number" reference in the WPS114 msg

webknjaz:
This is pretty trivial. Saying `WPS114: Found underscored name pattern:
IS_PYTEST_4_PLUS` is confusing and doesn't match the class name/purpose.
I uncovered this confusion when I hit #1448.
I don't think this deserves even a log entry.


sobolevn (approved):

sobolevn:
Thanks a lot!

PR #1471, opened 2020-06-19
Issue 1274

Dreamsorcerer:
Will need a little cleanup and probably some extra tests.

This adds:

- Explicit --create-baseline
- Remove unused violations from baseline and keep the file in sync with the code base.
- Handle file renames.
- Catch more cases with fuzzy matching.

### DIFF
--- a/wemake_python_styleguide/patches/baseline.py
+++ b/wemake_python_styleguide/patches/baseline.py
@@ -101,6 +101,9 @@ def _patch_start(manager: Type[Manager]) -> None:  # noqa: C901,WPS231
     def start(self, paths=None) -> None:  # noqa: WPS430
         # --- patch start
         self._wps_baseline = None
+        if not self.options.baseline:  # noqa: WPS513
+            self.options.baseline = baseline.BASELINE_FILE
+
         if self.options.create_baseline:
             if paths is not None or self.arguments:
                 response = input(  # noqa: WPS421
@@ -110,8 +113,6 @@ def _patch_start(manager: Type[Manager]) -> None:  # noqa: C901,WPS231
                 if not response.lower().startswith('y'):  # noqa: WPS513
                     sys.exit(3)
         else:
-            if not self.options.baseline:  # noqa: WPS513
-                self.options.baseline = baseline.BASELINE_FILE
             self._wps_baseline = baseline.load_from_file(self.options.baseline)
 
         if self._wps_baseline is None and not self.options.create_baseline:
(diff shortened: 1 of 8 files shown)


wemake_python_styleguide/patches/baseline.py
```diff
@@ -78,13 +84,89 @@ def _handle_results(  # noqa: WPS210, WPS430
...
 
+def _patch_start(manager: Type[Manager]) -> None:
+    original_start = manager.start
+
+    def start(self, paths=None) -> None:  # noqa: WPS430
+        # --- patch start
+        self._wps_baseline = None
+        if self.options.create_baseline:
+            if paths is not None or self.arguments:
+                response = input("This will create a new baseline for only the given files. Continue? (y/n) ")
```
sobolevn, line 106: Is this a development only `input`?

Dreamsorcerer: No, is there some flake8 mechanism for getting user input?

The idea here is to double check the user's intentions if they've passed explicit files, as I expect in most cases that would be a mistake. Usually, you'd just run it across the entire project.

I added this on a hunch. But, a couple of hours later, this saved me from making a mistake.

sobolevn: No, there's no such mechanism. Let's leave it there for now, but we cannot not keep it in the final solution.

Dreamsorcerer:
@sobolevn Sorry for all the noise, I had difficulty getting the tests fully working locally, so used the CI for my testing.
I'd suggest merging with a squash commit.

Anyway, tests are sorted now, so we should be good to proceed with any final changes.
I also put a few too many # noqa comments in, so feel free to iterate on this and clean up some of them.

Dreamsorcerer:
@sobolevn How's the other project going? Do you have an idea when you'll get round to looking at this?

Dreamsorcerer:
@sobolevn *nudge* :P
If you get a chance to look over this shortly, I should have some time next week to do any extra tasks necessary.

sobolevn:
Hi, @Dreamsorcerer! I have decided to ignore this task until `0.15` is released.
This is overly complex for me right now.

I can only handle simple rules development right now.

sobolevn:
No, `--baseline` won't be added with this PR. Reasons:
1. It does not detect code movements
2. It heavily depends on flake8's internals
3. It is hacky

Dreamsorcerer:
>     1. It does not detect code movements

If you mean changes to code over time, I think I've handled that fairly well in this PR.

vkurilin:
I would add that `flakehell` has a community-maintained fork [`flakehaven`](https://github.com/flakeheaven/flakeheaven). Maybe instead of deleting all mentions, we can replace them with `flakehaven` and thus support the project?

PR #1548, opened 2020-08-12
Issue 1539

### LINKED ISSUE #1539: Add Exceptions for TooShortNameViolation
# Rule request

I tend to set min-name-length to 2 in my projects [default is 3]

Then, I get a WPS111 error, for example, for `import numpy as np`

## Thesis

I feel it would be great if there were an allowlist for names that are too short given the configuration but still allowed.

## Reasoning

No matter if min-name-length is 2 or 3, there will always be common idiomatic names that are short.

Examples: np (numpy), df (common name when working with only one Pandas DataFrame), sns (Seaborn), sa (sqlalchemy), ...

Alternative: Provide a default allowlist with common short names that do not cause an error.

deveshsangwan:
# I have made things!

## Checklist

## Related issues

Closes #1539

Added an allowlist for commonly used aliases/short forms.
 


wemake_python_styleguide/logic/naming/logical.py
```diff
@@ -109,7 +109,15 @@ def is_too_short_name(
     >>> is_too_short_name('xy', min_length=2, trim=True)
     False
 
+    >>> is_too_short_name('np', min_length=3)
+    False
+
     """
+    allow_list = ['np', 'pd', 'df', 'plt', 'sns', 'tf', 'cv']
```
sobolevn, line 116: We tend to move these values to https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/constants.py

deveshsangwan: ok will move them

deveshsangwan: @sobolevn You can check the changes now

sobolevn:
Thanks a lot!

PR #1588, opened 2020-09-05
Add "wrapper" to nested functions whitelist

### LINKED ISSUE #1582: Add "wrapper" to NESTED_FUNCTIONS_WHITELIST?
# Rule request

## Thesis and reasoning

Decorators are often created inside functions. These decorators are then supposed to be used on top of other functions, which means they must accept one, and also that they will often create and return a new function by wrapping the original one. It's a level-3 nesting of functions. Maybe WPS430 should be relaxed, or `NESTED_FUNCTIONS_WHITELIST` could include another name like `wrapper`?

Code example:

```python
from functools import wraps


def python(versions):
    """Run through multiple Python versions."""

    def decorator(func):
        @wraps(func)
        def wrapper(context, *args, **kwargs):
            for version in versions:
                # do things with version and context
                func(context, *args, **kwargs)

        return wrapper

    return decorator


@python(["3.6", "3.7", "3.8"])
def task(context):
    ...
```

What do you think 🙂 ?

pawamoy:
## Checklist

## Related issues

Closes #1582

### DIFF
--- a/wemake_python_styleguide/constants.py
+++ b/wemake_python_styleguide/constants.py
@@ -278,6 +278,7 @@ ALLOWED_BUILTIN_CLASSES: Final = frozenset((
 NESTED_FUNCTIONS_WHITELIST: Final = frozenset((
     'decorator',
     'factory',
+    'wrapper',
 ))
 
 #: List of allowed ``__future__`` imports.


sobolevn (approved):

sobolevn:
Thanks a lot!

pawamoy:
Oh OK, I was planning to add tests and docs a bit later. Maybe they are automated though.

sobolevn:
Yes, they are (in this case) 🙂

Issue #1440, opened 2020-06-02
WPS414 for attributes


Dreamsorcerer:
```
self._reader, self._writer = await asyncio.open_connection(...)
```
This gives: `WPS414 Found incorrect unpacking target`.

While the example for WPS414 with assigning into a dict looks dodgy. Assigning to `self` attributes looks like normal behaviour to me. Should this check be changed?

sobolevn:
Well, not really. Assigning to `self` might be possible. But, general multiple assignment to attributes is totally incorrect. 

I will think a little bit more about `self` case. I, personally, never write a code like this. But, I don't have an explanation why yet. Only a gut feeling. 

Let's take some more time to think about it.
Thanks a lot for the report! 👍

sobolevn:
I still think that 

```python
reader, writter = call()
self._reader = reader
self_writter = writter
```

is better.

Dreamsorcerer:
I can see the argument there, as we are modifying the state of the class, so every modification should be explicit and on it's own line?

If that's the reasoning, maybe we can add something to the rationale in the docs?

sobolevn:
Yes, PRs are welcome!

PR #1635, opened 2020-10-02
Add information about attributes for WPS 414

Dreamsorcerer:
Discussion from https://github.com/wemake-services/wemake-python-styleguide/issues/1440


sobolevn (approved):
💯

Issue #1525, opened 2020-07-28
Bad English writing


wyuenho:
# Bug report

Hi! Thanks for this awesome project. This has alleviated a lot of configuration pains. However, I have some suggestions to make:

## What's wrong

Broken English in documentation and error messages all over the place, a good example is [this](https://wemake-python-stylegui.de/en/latest/pages/usage/violations/consistency.html#wemake_python_styleguide.violations.consistency.WrongLoopIterTypeViolation). What do you mean "incorrect"? It's correct, the linter just suggests an alternative style with little examples and bad English to explain the rationale.

This does not help the user to improve their code, this just adds confusion, or worse, nudge the users to turn off everything they don't understand.

There are also lots of spelling errors, wrong usage of tenses, redundant words etc that make reading very unpleasant.

Even the Github issue template is bad. If should be "How should it be", not "How is that should be".

## How is that should be

Get an English speaker on the team, or write strings in German, use `gettext` to extract them, and throw it into a service like [Lokalise](https://lokalise.com/) or [Smartling](https://www.smartling.com/) to translate into English properly.

sobolevn:
Any help in this field is appreciated! 👍

siva432:
I would like to contribute to this :-)

sobolevn:
@siva432 thanks a lot!

siva432:
where or which HTML file should I edit?

sobolevn:
@siva432 there are no HTML files to edit actually. We have two sources of docs:
- `.rst` files: https://github.com/wemake-services/wemake-python-styleguide/tree/master/docs and https://github.com/wemake-services/wemake-python-styleguide/tree/master/docs/pages
- `.py` files, like: https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/violations/best_practices.py

You can edit them! 👍

tangowithfoxtrot:
I have made some grammar and spelling corrections to the best_practices.py file. Would you mind providing some guidance on how to contribute them? Please and thank you!

sobolevn:
@tangowithfoxtrot do you mean #1540 ?

siva432:
ok @tangowithfoxtrot go ahead even I'm working on it, I'm just little busy now u take care or it.

siva432:
@sobolevn I have finished all changes how should resend the same PR, I tried by saving the file, git add , git push .
but I'm, not receiving any PR.

[...]

PR #1636, opened 2020-10-02
Update summary strings for conciseness.

Dreamsorcerer:
Updating with suggestions from: https://github.com/wemake-services/wemake-python-styleguide/issues/1525#issuecomment-675714883

This also means the summaries are more concise and the tense has changed (to the imperative form, I think), which also matches PEP 257.


sobolevn (approved):
Awesome! 👍

PR #1648, opened 2020-10-05
Forbid non-trivial expressions as an argument to  (#1599)

### LINKED ISSUE #1599: Forbid `except TypeError or ValueError`
This code should be invalid:

```python
try:
    ...
except TypeError or ValueError:
    ...

```

lensvol:
# I have made things!

## Checklist

## Related issues

Closes #1599.

### DIFF
--- a/wemake_python_styleguide/visitors/ast/exceptions.py
+++ b/wemake_python_styleguide/visitors/ast/exceptions.py
@@ -26,8 +26,6 @@ from wemake_python_styleguide.violations.refactoring import (
 )
 from wemake_python_styleguide.visitors.base import BaseNodeVisitor
 
-TRIVIAL_EXCEPT_ARG_TYPES = (ast.Name, ast.Attribute)
-
 
 def _find_returing_nodes(
     node: ast.Try,
@@ -163,6 +161,8 @@ class WrongExceptHandlerVisitor(BaseNodeVisitor):
 
     _base_exception: ClassVar[str] = 'BaseException'
 
+    _trivial_except_arg_nodes: ClassVar[AnyNodes] = (ast.Name, ast.Attribute)
+
     def visit_ExceptHandler(self, node: ast.ExceptHandler) -> None:
         """
         Checks all ``ExceptionHandler`` nodes.
@@ -208,12 +208,12 @@ class WrongExceptHandlerVisitor(BaseNodeVisitor):
         if node.type is None:
             return
 
-        if isinstance(node.type, TRIVIAL_EXCEPT_ARG_TYPES):
+        if isinstance(node.type, self._trivial_except_arg_nodes):
             return
 
         if isinstance(node.type, ast.Tuple):
             all_elements_are_trivial = all((
-                isinstance(element, TRIVIAL_EXCEPT_ARG_TYPES)
+                isinstance(element, self._trivial_except_arg_nodes)
                 for element in node.type.elts
             ))
             if all_elements_are_trivial:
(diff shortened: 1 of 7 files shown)


wemake_python_styleguide/visitors/ast/exceptions.py
```diff
@@ -198,3 +200,21 @@ def _check_exception_type(self, node: ast.ExceptHandler) -> None:
...
+        # Catch-all 'except' is actually okay in this case
+        if node.type is None:
+            return
+
+        if isinstance(node.type, (ast.Name, ast.Attribute)):
+            return
+
+        if isinstance(node.type, ast.Tuple):
+            all_elements_are_trivial = all((
+                isinstance(element, (ast.Name, ast.Attribute))
```
skarzi, line 214: NIP: let's add some constant for trivial elements types, so it will be easier to extend in the future and thanks to this code repetitions (this tuple is also used in line 209) can be avoided

sobolevn: 👍

sobolevn (changes requested):
Thanks a lot! I agree with @skarzi that constant would be a good fit here.

lensvol:
Okay, I moved them into module-level constant.

wemake_python_styleguide/visitors/ast/exceptions.py
```diff
@@ -26,6 +26,8 @@
 )
 from wemake_python_styleguide.visitors.base import BaseNodeVisitor
 
+TRIVIAL_EXCEPT_ARG_TYPES = (ast.Name, ast.Attribute)
```
sobolevn, line 29: Let's make it a class-level contant like we always do. 
Example: https://github.com/wemake-services/wemake-python-styleguide/blob/3e679c1a23dc149b664657df5a6374af71fdd164/wemake_python_styleguide/visitors/ast/exceptions.py#L59

sobolevn (changes requested):

wemake_python_styleguide/visitors/ast/exceptions.py
sobolevn: Sorry I had to be more specific!

sobolevn:
One more thing (I know, that it is unrelated, but we have to clear the mess sometimes): let's move `_find_returning_nodes` (https://github.com/wemake-services/wemake-python-styleguide/blob/3e679c1a23dc149b664657df5a6374af71fdd164/wemake_python_styleguide/visitors/ast/exceptions.py#L32) to `wemake_python_styleguide.logic.tree.exceptions`

lensvol:
Done.

wemake_python_styleguide/visitors/ast/exceptions.py
```diff
@@ -1,12 +1,13 @@
...
 from collections import Counter
-from typing import ClassVar, Set, Tuple
+from typing import ClassVar, Set
 
 from typing_extensions import final
 
 from wemake_python_styleguide.compat.aliases import FunctionNodes
 from wemake_python_styleguide.logic import nodes
 from wemake_python_styleguide.logic.tree import exceptions
+from wemake_python_styleguide.logic.tree.exceptions import find_returning_nodes
```
sobolevn, line 10: We have `exceptions` and `from exception` imports. Why so?

There's a proposed rule to forbid this style.

lensvol:
Because if I use it as `exceptions.find_returning_nodes` then _E501_ is triggered (85 symbols instead of 80), and I've already spent enough time at this completely unrelated change.

What do you propose?

sobolevn (approved):

sobolevn:
We will then fix it when the rule will officially land into the project. Thanks a lot! 👍

PR #1650, opened 2020-10-06
WPS531: Forbid just returning a boolean constant in `if` statements

### LINKED ISSUE #1638: Return the boolean condition directly without using `if`
# Rule request

## Thesis

We need to track:

```python
    if cond:
        return True  # or False
    else:
        return False  # or True
```

and:

```python
    if cond:
        return True  # or False
    return False  # or True
```

Instead one should use:

```python
    return cond
```

Inspiration: https://github.com/MartinThoma/flake8-simplify/blob/master/flake8_simplify.py#L487

## Reasoning

We don't need extra conditions to this simple thing.

Jrryy:
# I have made things!

Extends the functionality of WPS502: it is forbidden to use `if-else` or early returning `if` statements inside a function to check a condition and immediately returning a boolean depending on its outcome. As an example, it is forbidden to do:

```python
def some_function():
    if condition:
        return True
    else:
        return False
```

when this could be easily translated to:

```python
def some_function():
    return some_function
```

or also:

```python
def some_function():
    return bool(some_function)
```

## Checklist

## Related issues

Closes #1638


tests/fixtures/noqa/noqa.py
```diff
@@ -500,6 +500,18 @@ class MyBadException(BaseException):  # noqa: WPS418
 else:
     some_dict['x'] = False
 
+def wrong_if():
```
sobolevn, line 503: You can delete this one, just one `WPS502` is enough

wemake_python_styleguide/violations/refactoring.py
```diff
@@ -201,11 +202,37 @@ class SimplifiableIfViolation(ASTViolation):
         # Wrong:
         my_bool = True if some_call() else False
 
+        # In functions:
```
sobolevn, line 205: We need a brand new violation for this.

sobolevn (changes requested):
Thanks a lot!

Some more work is required 🙂

setup.cfg
```diff
@@ -41,6 +41,8 @@ ignore = D100, D104, D401, W504, RST303, RST304, DAR103, DAR203
 per-file-ignores =
   # These function names are part of 3d party API:
   wemake_python_styleguide/visitors/ast/*.py: N802
+  # This module contains many visitors
+  wemake_python_styleguide/visitors/ast/compares.py: N802, WPS202
```
sobolevn, line 45: If we have too many visitors, we should refactor our module 🙂 
Not just hide it!

wemake_python_styleguide/violations/refactoring.py
```diff
@@ -1222,3 +1224,56 @@ class ImplicitNegativeIndexViolation(ASTViolation):
...
+        code.
+
+    Solution:
+        Instead of testing the condition and returning a boolean, return the
+        condition itself.
+
+    Example::
+
+        # Correct:
+        def some_function():
```
sobolevn, line 1246: Just one correct and just one wrong examples are fine.

wemake_python_styleguide/visitors/ast/compares.py
```diff
@@ -416,6 +382,72 @@ def _is_simplifiable_assign(
...
+        Ensures that ``if`` in a function or method do not just return a bool.
+
+        Raises:
+            SimplifiableReturningIfStatementViolation
+
+        """
+        self._check_simplifiable_conditional_statement(node)
+        self.generic_visit(node)
+
+    def _returns_bool(self, node: ast.AST) -> bool:
```
sobolevn, line 400: We can move this to `logic`

sobolevn: And other logic parts as well.

sobolevn (changes requested):
Thanks a lot! Great progress.

This now looks way better! Let's polish it some more. 👍

setup.cfg
Jrryy: I guessed so, but I couldn't just do that on my own without asking about it first. I'd need to know which ones I should move and how I should call the new visitors module.

wemake_python_styleguide/violations/refactoring.py
Jrryy: I agree on leaving one wrong example but from my experience using this project at work, I'd prefer to leave the two first correct examples so as to avoid confusion because those two examples are pretty different from one another and still correct. I'll leave only one if you insist though.

Jrryy:
I've just ran into a few problems I would like to ask about. I've noticed that there was an `IfStatementVisitor` in `conditions.py` which seemed more fitting to implement this in. Doing this, I've seen that `UselessReturningElseViolation` (503) already existed and would throw violations in case anyone wanted to return anything in an `else` statement. 

This means that whenever someone tried to do:

```python
def func():
    if x:
        return True
    else:
        return False
```

This would always throw both `503` and `531`,  and even if in the `else` they tried to return something that is not a boolean, it would still throw `503`. The only way for this violation not to be thrown is to use early returns. That is:

```python
def func():
    if x:
        return True
    return False
```

This would be the only way to throw only `531`.

Having said all of this, my question is: should I implement this violation only to be thrown in the case of the second block of code, assuming that if the first is used `503` will be thrown instead, or do I implement it to be thrown also in the first case and modify the tests of `503` adding also `531` in the test cases that use a block of code as such?

sobolevn:
>

tests/test_visitors/test_ast/test_conditions/test_returning_else/test_redundant_returning_else.py
```diff
@@ -88,6 +88,57 @@ def test_else_that_can_be_removed_in_function(
     assert_errors(visitor, [UselessReturningElseViolation])
 
 
+@pytest.mark.parametrize('template', [
+    multiple_ifs1,
+    multiple_ifs2,
+])
+@pytest.mark.parametrize('code', [
+    'return True',
```
sobolevn, line 96: What about `return False` here?

tests/test_visitors/test_ast/test_conditions/test_returning_else/test_redundant_returning_else.py
```diff
@@ -88,6 +88,57 @@ def test_else_that_can_be_removed_in_function(
...
+
+    assert_errors(visitor, [UselessReturningElseViolation])
+
+
+@pytest.mark.parametrize('template', [
+    function_level_condition,
+    for_loop_level_condition,
+])
+@pytest.mark.parametrize('code', [
+    'return True',
```
sobolevn, line 120: Let's add `return False`

tests/test_visitors/test_ast/test_conditions/test_simplifiable_if_statements.py
```diff
@@ -0,0 +1,206 @@
...
+])
+def test_complex_early_returning_if_inside(
+    assert_errors,
+    parse_ast_tree,
+    comparators,
+    default_options,
+):
+    """These more complex early returning ifs can not be simplified (inside)."""
+    tree = parse_ast_tree(
+        complex_early_returning_if_inside.format(*comparators),
```
sobolevn, line 94: Let's join it with `complex_early_returning_if_outside` case

wemake_python_styleguide/logic/tree/ifs.py
```diff
@@ -6,11 +6,16 @@
...
 
-def has_elif(node: ast.If) -> bool:
+def is_elif(node: ast.If) -> bool:
     """Tells if this node is a part of a ``if`` chain or just a single one."""
     return getattr(node, 'wps_if_chain', False)  # noqa: WPS425
 
 
+def has_elif(node: ast.If) -> bool:
+    """Tells if this statement has an ``elif`` chained or not."""
+    return bool(tuple(chain(node))[1:-1])
```
sobolevn, line 16: Let's rewrite `[1:-1]`. I cannot understand it 😞

wemake_python_styleguide/violations/refactoring.py
```diff
@@ -1195,3 +1196,44 @@ class ImplicitNegativeIndexViolation(ASTViolation):
...
+        Instead of testing the condition and returning a boolean, return the
+        condition itself. This applies to early returning ifs too.
+
+    Example::
+
+        # Correct:
+        def some_function():
+            return some_condition
+
+        def some_function():
```
sobolevn, line 1221: Let's remove this example.

sobolevn (changes requested):
Thanks a lot! 👍 

This is almost done!

sobolevn (approved):

sobolevn:
Thanks a lot! Great work!

PR #1661, opened 2020-10-12
Forbid assigning to a slice #1629

### LINKED ISSUE #1629: Forbid assigning to a slice
# Rule request

## Thesis

There might be some crazy examples of mutation, when assigning to a slice:

```python
my_list = [1, 2, 3]
my_list[0:3] = [0, 0, 0]
```

So, let's ban this.

However, we still allow regular index assignments:

```python
my_list = [1, 2, 3]
my_list[0] = [0, 0, 0]
```

Is ok.

sudo-k-runner:
Forbid assigning to a slice #1629

## Checklist

## Related issues

Closes #1629


tests/test_visitors/test_ast/test_subscripts/test_slice_assignment.py
```diff
@@ -0,0 +1,50 @@
+import pytest
+
+from wemake_python_styleguide.violations.consistency import (
+    AssignToSliceViolation,
+)
+from wemake_python_styleguide.visitors.ast.subscripts import SubscriptVisitor
+
+usage_template = 'constant[{0}]= [1,2,3]'
```
sobolevn, line 8: `'constant[{0}]= [1, 2, 3]'`

tests/test_visitors/test_ast/test_subscripts/test_slice_assignment.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@pytest.mark.parametrize('expression', [
+    ':7',
+    '1:7:2',
+    '3:',
+    '3::',
+    '::2',
+    ':2:',
```
sobolevn, line 17: Let's also test this case:

```python
>>> l = [1, 2, 3]
>>> l[slice(1)]
[1]
```

wemake_python_styleguide/violations/consistency.py
```diff
@@ -2197,3 +2199,30 @@ class FloatZeroViolation(TokenizeViolation):
...
     code = 358
     error_template = 'Found a float zero (0.0)'
+
+
+@final
+class AssignToSliceViolation(ASTViolation):
+    """
+    Forbid assignment to a subscript slice.
+
+    Reasoning:
```
sobolevn, line 2209: We are missing `Solution` section.

wemake_python_styleguide/violations/consistency.py
```diff
@@ -2197,3 +2199,30 @@ class FloatZeroViolation(TokenizeViolation):
...
+        We do this for consistency.
```
sobolevn, line 2210: Consistency is not the main reason here. Hard to spot bugs and list that changes its size implicitly and strangely are our reasons.

wemake_python_styleguide/violations/consistency.py
```diff
@@ -2197,3 +2199,30 @@ class FloatZeroViolation(TokenizeViolation):
...
+
+    Example::
+
+        # Correct:
+        a[5] = [1]
+
+        # Wrong:
+        a[1:3] = [1,2]
```
sobolevn, line 2218: One example is enough.

sobolevn (changes requested):
Thanks a lot! 👍

tests/test_visitors/test_ast/test_subscripts/test_slice_assignment.py
sudo-k-runner: Thanks will take care of that. 
@sobolevn  Also wanted to check do we allow something like this? replacing the full list or forbid that as well?
```
>>> l = [1, 2, 3]
>>> l[:] = [4, 5]
```

sobolevn: Should be banned as well!

tests/test_visitors/test_ast/test_subscripts/test_slice_assignment.py
sudo-k-runner: Done

kxepal:
Sorry to be late for the party, but why? Slice assignment is the only way for inplace array multiple replacement when you need that. Any other alternatives on the table? Shouldn't any forbid rule follow with howto one?

And no, explicit index assignment isn't an alternative since you replace one shot operation which could be optimized by underlying structure (like numpy.array) via O(N) one which wouldn't much be optimal in most of cases.

sobolevn:
It can be switched off for `numpy` array. In regular `python` this does not make any sence.

kxepal:
The sense is quite the same: provide the way for multiple sequential items replacement in a single shot. 

Dummy example how it works:
```
In [1]: l = list(range(100500))                                                                                                                                                                                                               

In [2]: def a(l, r=list(range(100,500))): 
   ...:     l[100:500] = r 
   ...:                                                                                                                                                                                                                                       

In [3]: def b(l, r=list(range(100,500))): 
   ...:     for i in r: 
   ...:         l[i] = i 
   ...:                                                                                                                                                                                                                                       

In [4]: %timeit a(l)                                                                                                                                                                                                                          
1.09 µs ± 27.4 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [5]: %timeit b(l)                                                                                                                                                                                                                          
14.4 µs ± 268 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

So at least this rule forces to do not an optimal operations. Any other reasons to not do slice assignments?

kxepal:
The quite common and useful example which violates this rule is inplace list replacement via `[:]` - this helps to keep the same object reference while it content could be completely erased or replaced with the new one. Shared list thing.

sobolevn:
@kxepal thanks a lot! This is a valuable addition. 

The main idea of this rule is that `list` is mutable. And assigning to a slice changes the size of the list implicitly.

```python
>>> l = [0, 1, 2, 3, 4]
>>> l[0:2]
[0, 1]
>>> l[0:2] = ['a', 'b', 'c']
>>> l
['a', 'b', 'c', 2, 3, 4]
```

And this is really confusing. How can I tell the final size of the list after `l[0:2] = some()`?
We can't even do this with `mypy`.

I think that this code is ugly. And I would be really unpleased to something like it in our production code.
On the other hand:

```python
# We need this operation to be fast, because this is a hot path,
# so we use a hack to insert a part of another list in one bunch.
new_list_part = some()
assert len(new_list_part) == 500 - 100
my_list[100:500] = some_call()  # noqa: WPS-whatever
```

This code looks reasonable to me.

sobolevn:
@sudo-k-runner can you please 
1. rebase your branch on current `master`?
2. add bits from our discussion with @kxepal to the docs? 

Thanks!

sudo-k-runner:
@sobolevn Sure, apologies for the delay, got side tracked. Will updated the PR.

sobolevn (approved):
Great work, @sudo-k-runner 
Thanks!

PR #1670, opened 2020-10-14
Fix false positive in WPS513

### LINKED ISSUE #1632: Stray WPS513 warning.
# Bug report

`WPS513` is emitted on `if`/`else`/`if` chains that contain other statements in the last `if` block.

## What's wrong

The following snippet:

```python
if private_key is not None:
    if isinstance(private_key, PKey):
        key = private_key
    else:
        if isinstance(private_key, str):  # WPS513 Found implicit `elif` condition
            stream = io.StringIO()
            stream.write(private_key)
            stream.seek(0)
            private_key = stream
        key = RSAKey.from_private_key(private_key)
```

triggers a stray WPS513 warning.  Seems like the assignment to `key` is not taken into account.

## How is that should be

`WPS513` should only be emitted if the dangling `if` statement is the only one present in the `else` block.

Jrryy:
# I have made things!

Solve a bug that throws a WPS513 when the else has more code in it below the if inside of it, thus making the elif not implicit. For example, this structure:

```python
if some:
    ...
else:
    if other:
        ...
    ...
```

Should not throw a WPS513 since there is more code after the inner if, thus making an elif not possible.

## Checklist

## Related issues

- Closes #1632

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ Semantic versioning in our case means:
 - Make `WPS326` work when there is comment between string literals
 - Allowed yield statements in call method
 - Allow to use `^` with `1`
+- Fixes false positives in WPS513
 
 ### Misc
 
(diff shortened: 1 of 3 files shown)


sobolevn (approved):

sobolevn:
Thanks!

PR #1676, opened 2020-10-19
Fix WPS323 false positive thrown when using some psycopg2 and datetime functions

### LINKED ISSUE #1329: False positive for WPS323 on date/time formatting
# Bug report

Some date and time formatting cases are treated as violation:

    some_datetime.strftime('%Y-%m-%d-%H-%M')
    '{0:%Y-%m-%d-%H-%M}'.format(some_datetime)

But others are not:

    some_datetime.strftime('%H-%M')
    '{0:%H-%M}'.format(some_datetime)

## What's wrong

There is no easy to use alternative for custom date/time formatting except C standard codes.

## How is that should be

The "%" symbol should be allowed at least as part of format specification for new-style format strings.

Jrryy:
# I have made things!

Fix that WPS323 is thrown when executing functions such as:

```python
some_date = datetime.strptime("01-01-2020 (10:20:30)", "%d-%m-%Y (%H:%M:%S)")
formatted_datetime = some_date.strftime("%d-%m-%Y (%H:%M:%S)")
cur.execute
```

## Checklist

## Related issues

- Closes #1329

### DIFF
--- a/tests/test_visitors/test_ast/test_builtins/test_strings/test_modulo_formatting.py
+++ b/tests/test_visitors/test_ast/test_builtins/test_strings/test_modulo_formatting.py
@@ -145,6 +145,32 @@ def test_regular_modulo_string(
 
 
 @pytest.mark.parametrize('code', [
+    'datetime.strftime("%d-%m-%Y (%H:%M:%S)")',
+    'datetime.strptime("01-01-2020 (10:20:30)", "%d-%m-%Y (%H:%M:%S)")',
+    'date.strftime("%d-%m-%Y (%H:%M:%S)")',
+    'date.strptime("01-01-2020", "%d-%m-%Y")',
+    'time.strftime("%H:%M:%S")',
+    'time.strptime("10:20:30", "%H:%M:%S")',
+    'strptime("01-01-2020 (10:20:30)", "%d-%m-%Y (%H:%M:%S)")',
+    'cur.execute("SELECT * FROM table WHERE column = %s", ("some_column"))',
+    'execute("SELECT * FROM table WHERE column = %s", ("some_column"))',
+])
+def test_functions_modulo_string(
+    assert_errors,
+    parse_ast_tree,
+    code,
+    default_options,
+):
+    """Testing that the strings violate the rules."""
+    tree = parse_ast_tree(code)
+
+    visitor = WrongStringVisitor(default_options, tree=tree)
+    visitor.run()
+
+    assert_errors(visitor, [], (FormattedStringViolation,))
+
+
+@pytest.mark.parametrize('code', [
     'x % 1',
     '"str" % 1',
     'name % value',
--- a/tests/test_visitors/test_ast/test_builtins/test_strings/test_modulo_formatting.py
+++ b/tests/test_visitors/test_ast/test_builtins/test_strings/test_modulo_formatting.py
@@ -145,6 +145,7 @@ def test_regular_modulo_string(
 
 
 @pytest.mark.parametrize('code', [
+    'dt.strftime("%A, %d. %B %Y %I:%M%p")',
     'datetime.strftime("%d-%m-%Y (%H:%M:%S)")',
     'datetime.strptime("01-01-2020 (10:20:30)", "%d-%m-%Y (%H:%M:%S)")',
     'date.strftime("%d-%m-%Y (%H:%M:%S)")',
--- a/wemake_python_styleguide/visitors/ast/builtins.py
+++ b/wemake_python_styleguide/visitors/ast/builtins.py
@@ -26,6 +26,7 @@ from wemake_python_styleguide.compat.aliases import (
 from wemake_python_styleguide.logic import nodes, safe_eval, source, walk
 from wemake_python_styleguide.logic.naming.name_nodes import extract_name
 from wemake_python_styleguide.logic.tree import operators, strings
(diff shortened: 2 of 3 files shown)


wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -112,6 +112,30 @@ def _check_is_alphatbet(
...
+        modulo patterns because they must have them for the functions to work
+        properly.
+        """
+        exceptions = (
+            'strftime',  # For date, time, and datetime.strftime()
+            'strptime',  # For date, time, and datetime.strptime()
+            'execute',  # For psycopg2's cur.execute()
+        )
+
+        parent = nodes.get_parent(node)
```
sobolevn, line 129: Can we reuse this function? https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/logic/tree/functions.py#L13

sobolevn (commented):
Thanks a lot!

wemake_python_styleguide/visitors/ast/builtins.py
Jrryy: Sure, there was some issue with the typing but I'll fix it tomorrow.

tests/test_visitors/test_ast/test_builtins/test_strings/test_modulo_formatting.py
```diff
@@ -144,6 +144,32 @@ def test_regular_modulo_string(
     assert_errors(visitor, [], (FormattedStringViolation,))
 
 
+@pytest.mark.parametrize('code', [
+    'datetime.strftime("%d-%m-%Y (%H:%M:%S)")',
```
sobolevn, line 148: Also, please check `dt.`

sobolevn (approved):

wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -112,6 +112,29 @@ def _check_is_alphatbet(
...
+        """
+        exceptions = (
+            'strftime',  # For date, time, and datetime.strftime()
+            'strptime',  # For date, time, and datetime.strptime()
+            'execute',  # For psycopg2's cur.execute()
+        )
+
+        if parent and isinstance(parent, ast.Call):
+
+            func_name = getattr(parent.func, 'attr', None)
```
sobolevn, line 131: Oh, one more thing: what about `given_function_call` here? https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/logic/tree/functions.py#L13

Jrryy: This method has a problem: if, let's say, `datetime.strptime()` is executed, it will check that the whole `datetime.strptime` is in the exceptions, but in case the user imports `datetime` as `dt`, or anything else, it will fail. With functions like `strftime` this is even more of a problem, because it's used in datetime objects stored in variables with names defined by the users. I can use `given_function_called` for this PR, but this would require to modify that function to check only the function name instead of the function and the modules it belongs to.

I propose creating an issue to modify its behavior, either by splitting the modules and function's string by dots and comparing only the function name, or by using `endswith` which would allow the developers to use function and modules names in the `to_check` parameter.

sobolevn: @Jrryy yes, please!

wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -111,6 +116,28 @@ def _check_is_alphatbet(
...
 
+    def _is_modulo_pattern_exception(self, parent: Optional[ast.AST]) -> bool:
+        """
+        Check if the string with modulo patterns is in an exceptional situation.
+
+        Basically we have some function names in which we allow strings with
+        modulo patterns because they must have them for the functions to work
+        properly.
+        """
+        exceptions = (
```
sobolevn, line 127: Let's make it:

```python
_exceptions: FrozenSet[str] = frozenset((
```

sobolevn (changes requested):
Thanks a lot! Great work!

wemake_python_styleguide/visitors/ast/builtins.py
sobolevn: And we need a new line before it.

wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -49,6 +49,14 @@
 _HashItems = Sequence[Optional[ast.AST]]
 
 
+#: names of functions in which we allow strings with modulo patterns.
+_modulo_pattern_exceptions: ClassVar[FrozenSet[str]] = frozenset((
```
sobolevn, line 53: This is a module-level, isn't it? 🤔 

nitpick: `#: n` -> `#: N`

Jrryy: Ah god dammit I'm dumb

sobolevn (approved):

sobolevn:
Thanks a lot! This hack is ugly, but really useful! 😆

Jrryy:
I know, I tried to refactor it visiting the Mod operator and it worked fine when the left operator was a constant string, but the ast can't load a variable's value so it was pretty much in vain and had to come back to this ugly patch.

PR #1688, opened 2020-10-21
Improve given_function_called

### LINKED ISSUE #1682: Method `given_function_called` should only take the function name into account.
# Bug report

## What's wrong

The method `given_function_called(node: ast.Call, to_check: Container[str]) -> str` in `logic.tree.functions` is described as a method that returns the name of the function being called in `node`, in case it is included in `to_check`. For example:
 ```python
# Let's imagine we are visiting the Call node in `print(123, 456)` stored in `node`
called_function = given_function_called(node, ['print'])
print(called_function)
# Prints `print`
# But, if we are visiting `datetime.timedelta(days=1)`
called_function = given_function_called(node, ['timedelta'])
print(called_function)
# Prints an empty string, as if `timedelta` had not been called. The way for it to be shown is:
called_function = given_function_called(node, ['datetime.timedelta'])
print(called_function)
# Prints `datetime.timedelta`
```

This is related to https://github.com/wemake-services/wemake-python-styleguide/pull/1676#discussion_r508471791

## How is that should be

```python
# If we are visiting `datetime.timedelta(days=1)`
called_function = given_function_called(node, ['timedelta'])
print(called_function)
# Prints `timedelta`
```

Jrryy:
# I have made things!

Added a parameter `split_modules: bool` to `given_function_called` in `wemake_python_styleguide.logic.tree.functions`. Defaults to False.

When False, behaves as always.
When True, it only checks that a function has been called, regardless of the modules it comes from. For example:

```python
import ast

module = ast.parse('datetime.timedelta(days=1)')
called_function = given_function_called(module.body[0].value, ['datetime.timedelta'], split_modules=False)
print(called_function) 
# Prints `datetime.timedelta`

called_function = given_function_called(module.body[0].value, ['timedelta'], split_modules=False)
print(called_function)
# Prints `` (empty string)

called_function = given_function_called(module.body[0].value, ['timedelta'], split_modules=True)
print(called_function)
# Prints `timedelta`
```

## Checklist

## Related issues

- Closes #1682

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Semantic versioning in our case means:
 - WPS531: Forbids testing conditions to just return booleans when it is possible to simply return the condition itself
 - Forbids to use unsafe infinite loops
 - Forbids to use raw strings `r''` when not necessary
+- Add parameter to `given_function_called` so that it checks if a function has been called by its name regardless of the modules or classes it is in
 
 ### Bugfixes
 
--- a/wemake_python_styleguide/logic/tree/functions.py
+++ b/wemake_python_styleguide/logic/tree/functions.py
@@ -10,9 +10,16 @@ from wemake_python_styleguide.types import (
 )
 
 
-def given_function_called(node: Call, to_check: Container[str]) -> str:
+def given_function_called(
+    node: Call,
+    to_check: Container[str],
+    split_modules: bool = False,
+) -> str:
     """
     Returns function name if it is called and contained in the container.
+
+    If `split_modules`, takes the modules or objects into account. Otherwise,
+    it only cares about the function's name.
 
     >>> import ast
     >>> module = ast.parse('print(123, 456)')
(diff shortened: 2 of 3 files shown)


wemake_python_styleguide/logic/tree/functions.py
```diff
@@ -10,10 +10,17 @@
 )
 
 
-def given_function_called(node: Call, to_check: Container[str]) -> str:
+def given_function_called(
+    node: Call,
+    to_check: Container[str],
+    split_modules: bool = False,
```
sobolevn, line 16: Lets makeit `kwonly`: 
```python
*,
split_modules: bool = False,
```

wemake_python_styleguide/logic/tree/functions.py
```diff
@@ -22,8 +29,25 @@ def given_function_called(node: Call, to_check: Container[str]) -> str:
     >>> given_function_called(module.body[0].value, ['adjust'])
     ''
 
+    >>> module = ast.parse('datetime.timedelta(days=1)')
```
sobolevn, line 32: We can remove all doctest, because they are replaced with new test cases.

sobolevn (changes requested):
Thanks!

sobolevn (approved):

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ Semantic versioning in our case means:
 - WPS531: Forbids testing conditions to just return booleans when it is possible to simply return the condition itself
 - Forbids to use unsafe infinite loops
 - Forbids to use raw strings `r''` when not necessary
+- Add parameter to `given_function_called` so that it checks if a function has been called by its name regardless of the modules or classes it is in
```
sobolevn, line 30: Btw, we only put user-facing changes here. Som you can delete this line in the future PR.

sobolevn:
Thanks again! 👍

CHANGELOG.md
Jrryy: ah, gotcha

PR #1694, opened 2020-10-24
Fix #1689: lambda in loop false positive WPS426

### LINKED ISSUE #1689: WPS426 triggers on safe lambdas
# Bug report

## What's wrong

```python
def _show_titles(number_of_posts: int) -> Sequence[FutureResultE[str]]:
    return [
        # Notice how easily we compose async and sync functions:
        _fetch_post(post_id).map(lambda post: post['title'])
        # TODO: try `for post_id in (2, 1, 0):` to see how async errors work
        for post_id in range(1, number_of_posts + 1)
    ]
```

Raises:

```
  32:9     WPS426 Found `lambda` in loop's body
  _fetch_post(post_id).map(lambda post: post['title'])
  ^
```

But it should not!

## How is that should be

Why? Because this `lambda` does not use `post_id` in its body. It is safe to be used.
So, if `lambda` uses only its args - we can allow it.

### flake8 information

Contents of `flake8 --bug-report`:

### pip information

Contents of `pip freeze`:

jaydesl:
# I have made things!

Allow a `lambda` in a loop as long as it does not act on variables outside of its scope. 

Lambda functions using only their own arguments are considered safe.

## Checklist

## Related issues

Closes #1689 

### DIFF
--- a/wemake_python_styleguide/visitors/ast/loops.py
+++ b/wemake_python_styleguide/visitors/ast/loops.py
@@ -177,15 +177,8 @@ class WrongLoopVisitor(base.BaseNodeVisitor):
         self,
         node: Union[AnyLoop, AnyComprehension],
     ) -> None:
-        container_names = self._containers.get(type(node), ())
-        for container in container_names:
-            body = getattr(node, container, [])
-            if not isinstance(body, list):
-                body = [body]
-
-            for subnode in body:
-                if walk.is_contained(subnode, ast.Lambda):
-                    self.add_violation(LambdaInsideLoopViolation(node))
+        for _ in walk.get_subnodes_by_type(node, ast.Lambda):
+            self.add_violation(LambdaInsideLoopViolation(node))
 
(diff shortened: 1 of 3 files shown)


wemake_python_styleguide/visitors/ast/loops.py
```diff
@@ -177,15 +177,17 @@ def _check_lambda_inside_loop(
...
-        container_names = self._containers.get(type(node), ())
-        for container in container_names:
-            body = getattr(node, container, [])
-            if not isinstance(body, list):
-                body = [body]
-
-            for subnode in body:
-                if walk.is_contained(subnode, ast.Lambda):
-                    self.add_violation(LambdaInsideLoopViolation(node))
+        for lambda_node in walk.get_subnodes_by_type(node, ast.Lambda):
```
sobolevn, line 180: Great solution! 👍

sobolevn (approved):
Thanks a lot! 👍 
Extra karma points for a quick fix!

PR #1697, opened 2020-10-25
Forbid to have getter functions that returns nothing (#1382)

### LINKED ISSUE #1382: `get_*` functions must return something.
# Rule request

## Thesis

Functions and methods with `get_` prefix must return something (have `return` or `yield` in the body).

## Reasoning

`get_` function is expected to return what it promises to return. For example, `get_user` returns a user object. If it's not so, it's an error or a bad interface.

hhsdev:
# I have made things!

- Add `GetterWithoutReturnViolation`, triggered when functions and
  methods whose name starts with `get_` returns nothing

- Rename `FunctionArgumentVisitor` to `FunctionSignatureVisitor`,
  in preparation to add code that checks the function name as well

- Add code checking for `GetterWithoutReturnViolation` to
  `FunctionSignatureVisitor`

- Add noqa tests

- Add unit test for `GetterWithoutReturnViolation` in
  `test_functions/test_getter_without_return.py`

- Update CHANGELOG.md appropriately

## Checklist

- Closes #1382

### DIFF
--- a/tests/test_visitors/test_ast/test_functions/test_getter_without_return.py
+++ b/tests/test_visitors/test_ast/test_functions/test_getter_without_return.py
@@ -1,0 +1,289 @@
+import pytest
+
+from wemake_python_styleguide.violations.best_practices import (
+    GetterWithoutReturnViolation,
+)
+from wemake_python_styleguide.visitors.ast.functions import (
+    FunctionSignatureVisitor,
+)
+
+getter_function_with_implicit_return = """
+def get_foo():
+    print('Hello world!')
+"""
+
+getter_function_with_bare_return = """
+def get_foo():
+    return
+"""
+
+getter_function_with_valued_return = """
+def get_foo():
+    return 1
+"""
+
+getter_function_with_explicit_none_return = """
+def get_foo():
+    return None
+"""
+
+getter_function_with_bare_yield = """
+def get_foo():
+    yield
+"""
+
+getter_function_with_valued_yield = """
+def get_foo():
+    yield 1
+"""
+
+getter_function_with_explicit_none_yield = """
+def get_foo():
+    yield None
+"""
+
+getter_function_with_yield_from = """
+def get_foo():
+    yield from [1]
+"""
+
+getter_method_with_implicit_return = """
+class Foo:
+    def get_foo(self):
+        print('Hello world')
+"""
+
+getter_method_with_bare_return = """
+class Foo:
+    def get_foo(self):
+        return
+"""
+
+getter_method_with_valued_return = """
+class Foo:
+    def get_foo(self):
+        return 1
+"""
+
+getter_method_with_explicit_none_return = """
+class Foo:
+    def get_foo(self):
+        return None
+"""
+
+getter_method_with_bare_yield = """
+class Foo:
+    def get_foo(self):
+        yield
+"""
+
+getter_method_with_valued_yield = """
+class Foo:
+    def get_foo(self):
+        yield 1
+"""
+
+getter_method_with_explicit_none_yield = """
+class Foo:
+    def get_foo(self):
+        yield None
+"""
+
+getter_method_with_yield_from = """
+class Foo:
+    def get_foo(self):
+        yield from [1]
+"""
+
+regular_function_with_bare_return = """
+class Foo:
+    def foo(self):
+        return
+"""
+
+regular_function_with_implicit_return = """
+class Foo:
+    def foo(self):
+        print('Hello World!')
+"""
+
+regular_function_with_bare_return = """
+class Foo:
+    def foo(self):
+        return
+"""
+
+regular_function_with_bare_yield = """
+class Foo:
+    def foo(self):
+        yield
+"""
+
+regular_method_with_bare_return = """
+class Foo:
+    def foo(self):
+        return
+"""
+
+regular_method_with_implicit_return = """
+class Foo:
+    def foo(self):
+        print('Hello world')
+"""
+
+regular_method_with_bare_return = """
+class Foo:
+    def foo(self):
+        return
+"""
+
+regular_method_with_bare_yield = """
+class Foo:
+    def foo(self):
+        yield
+"""
+
+getter_method_with_branched_return = """
+def get_foo():
+    if bar:
+        return 1
+"""
+
+getter_stub_with_docstring = """
+def get_foo():
+    '''Gets foo.'''
+"""
+
+getter_stub_with_ellipsis = """
+def get_foo():
+    ...
+"""
+
+getter_stub_with_raise = """
+def get_foo():
+    raise ValueError('Error')
+"""
+
+getter_stub_with_docstring_and_ellipsis = """
+def get_foo():
+    '''Gets foo.'''
+    ...
+"""
+
+getter_stub_with_docstring_and_raise = """
+def get_foo():
+    '''Gets Foo.'''
+    raise ValueError('Error')
+"""
+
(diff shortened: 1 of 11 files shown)


tests/test_visitors/test_ast/test_functions/test_getter_without_return.py
```diff
@@ -0,0 +1,249 @@
...
+from wemake_python_styleguide.violations.best_practices import (
+    GetterWithoutReturnViolation,
+)
+from wemake_python_styleguide.visitors.ast.functions import (
+    FunctionSignatureVisitor,
+)
+
+getter_function_with_implicit_return = """
+def get_foo():
+    ...
```
sobolevn, line 12: This should be allowed, because we can use WPS with stub files.

tests/test_visitors/test_ast/test_functions/test_getter_without_return.py
```diff
@@ -0,0 +1,249 @@
...
+    code,
+    mode,
+):
+    """Testing that getters which output values are allowed."""
+    tree = parse_ast_tree(mode(code))
+
+    visitor = FunctionSignatureVisitor(default_options, tree=tree)
+    visitor.run()
+
+    assert_errors(visitor, [])
```
sobolevn, line 201: Let's also test a function / method with just a docstring.
It should be allowed: intreface definitions, etc.

We should also allow `get_` function with just `raise` (+ an optional docstring). It is used in interfaces as well.

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -2255,3 +2257,47 @@ class InfiniteWhileLoopViolation(ASTViolation):
 
     error_template = 'Found an infinite while loop'
     code = 457
+
+
+class GetterWithoutReturnViolation(ASTViolation):
```
sobolevn, line 2262: Should be `@final`

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -2255,3 +2257,47 @@ class InfiniteWhileLoopViolation(ASTViolation):
...
+        A ``get_`` function is generally expected to return a value. Otherwise,
+        it is most likely either an error or bad naming.
+
+    Solution:
+        Make sure getter functions ``return`` or ``yield`` a value on all
+        execution paths, or rename the function.
+
+    Example::
+
+        # Correct:
```
sobolevn, line 2278:
```python
# Correct
def get_random_number():
     return random.randint(1, 10)

# Wrong
def get_random_number():
     print('I do not return a value!')
```

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -2255,3 +2257,47 @@ class InfiniteWhileLoopViolation(ASTViolation):
...
+
+            def get_data_into_database(self):  # load or update is better
+                self.database.update(http.request(url))
+
+
+    .. versionadded:: 0.15.0
+
+    """
+
+    error_template = 'Found getter function without a return value'
```
sobolevn, line 2302: `Found a getter without a return value`

wemake_python_styleguide/visitors/ast/functions.py
```diff
@@ -1,6 +1,6 @@
 import ast
+import typing
```
sobolevn, line 2: Please, don't use `import typing`
Because typing is already verbose, extra `typing.` makes it even worse 😢

sobolevn (changes requested):
Thanks a lot! Let's add some more test to be sure that we have covered every possible corner case.

wemake_python_styleguide/visitors/ast/functions.py
hhsdev: I was bugged about it as well, but it's saying too many imported members (10 > 8)

Extra types imported by me were `Type`, `Tuple` and `Iterator`, all needed for `_get_function_exit_nodes`. I thought about using `walk.get_subnodes_by_type` instead, but that function doesn't seem to take tuples as type.

sobolevn: Maybe we can extract some parts to `logic/`?

sobolevn:
Btw, there are some merge conflicts. Please, rebase your build 👍

tests/test_visitors/test_ast/test_functions/test_getter_without_return.py
hhsdev: Do we also allow if the `get_` function has extra statements besides the raise statement?
```
def get_foo():
    print('Debug: get_foo is called')
    raise ValueError('error!')
```

sobolevn: Nope, only a single `raise`

tests/test_visitors/test_ast/test_functions/test_getter_without_return.py
```diff
@@ -0,0 +1,249 @@
...
+"""
+
+
+@pytest.mark.parametrize('code', [
+    getter_function_with_implicit_return,
+    getter_function_with_bare_return,
+    getter_function_with_bare_yield,
+    getter_method_with_implicit_return,
+    getter_method_with_bare_return,
+    getter_method_with_bare_yield,
```
sobolevn, line 159: Bare `yield` should be allowed. Because `yield` can be used to define different constructs like `contextmanager` and `fixture`.

Can we please add this code as a test:

```python
def wrapper():
    print('context manager')
    yield
```

This should be allowed.

wemake_python_styleguide/visitors/ast/functions.py
```diff
@@ -432,21 +453,53 @@ class FunctionArgumentsVisitor(base.BaseNodeVisitor):
...
+        has_explicit_function_exit = False
+        for function_exit_node in self._get_function_exit_nodes(node):
+            has_explicit_function_exit = True
+            if function_exit_node.value is None:
+                self.add_violation(GetterWithoutReturnViolation(node))
+
+        if not has_explicit_function_exit:
+            self.add_violation(GetterWithoutReturnViolation(node))
+
+    def _get_function_exit_nodes(
```
sobolevn, line 494: We can move this to `logic/`

sobolevn (changes requested):
Thanks a lot! There are some other corner-cases that I found.

tests/test_visitors/test_ast/test_functions/test_getter_without_return.py
hhsdev: that would be allowed because the function name doesn't start with `get_`. should I add the same function with name `get_wrapper`?

sobolevn: Yeap 👍

tests/test_visitors/test_ast/test_functions/test_getter_without_return.py
```diff
@@ -0,0 +1,249 @@
...
+"""
+
+getter_function_with_yield_from = """
+def get_foo():
+    yield from [1]
+"""
+
+getter_method_with_implicit_return = """
+class Foo:
+    def get_foo():
```
skarzi, line 52: NIP: please, let's add `self` argument to all methods - it doesn't change anything, but method without `self` argument or `staticmethod` decorator is confusing

sobolevn: Good catch 👍

sobolevn (approved):
Thanks a lot, awesome feature!

Issue #1711, opened 2020-10-30
Forbid empty comments


sobolevn:
# Rule request

## Thesis

We should ban comments like this one:

```python
# 
my_var = 1
```

But, empty lines in the middle are fine:

```python
# First line
# 
# Samples:
# One
# Two
my_var = 1
```

We ban only empty lines in the beginning and in the end.

## Reasoning

Why not?

Elton-Lin:
Hello there,
@huangyunzen and I would like to work on this new rule. We have just started with the setup in the contribution page. We will report back when we get things ready and develop an initial plan. Thanks!

sobolevn:
Thanks a lot!

Elton-Lin:
We'd like to get some clarifications for the alerting style for consecutive empty comments. We think that it might make sense to only alert on the first empty line in this case, examples shown below:
```
#           <--- alert
#
my_var = 5
#           <--- alert
#
```

```
#           <--- alert
#
# A cat
#           <--- alert
#
```

What are you thoughts?

sobolevn:
Yes, the second one is fine.

sobolevn:
In other words: we ban only empty lines with any extra associated content.

Elton-Lin:
> Yes, the second one is fine.

Could you elaborate on this? Did you mean we shouldn't alert on my second example because there is actual content within the block of comments? I thought empty comments are only allowed if they are wrapped with actual content around like the example you gave previously. Thanks

sobolevn:
@Elton-Lin updated, sorry. I have to read more carefully 🙂

Elton-Lin:
Thanks for the update. Just to be extra sure, if we only alert on the first line for my second example, does it imply the following case is valid? (I'm guessing no) or we would still alert line 2. (essentially we focus on one block of comments at a time, so there will only be one alert for a block of comments, if needed)

```
# A cat
#           <--- alert or not
#
```

sobolevn:
Yes, this should also alert.

Elton-Lin:
For writing unit tests under tests/test_visitors/test_tokenize/test_comments/, we are wondering if we need to assert the line number of the alert besides confirming the correct violation is created. If so, what would be the best way to do that ? (extracting the line number of the violation and assert with the expected one)

sobolevn:
No, we don't usually assert the line number. We just assert that violation is there.

Elton-Lin:
[...]

PR #1747, opened 2020-11-26
WPS464: Forbid empty comments

Elton-Lin:
# I have made things!

WPS464: It is forbidden to have unnecessary empty comments. For example:
```python
# 
my_var = 1
```

But, empty lines in the middle are fine:
```python
# First line
# 
# Samples:
# One
# Two
my_var = 1
```
In other words, we ban only empty lines in the beginning and in the end. Also, all empty inline comments are forbidden. Overall, this rule is very straightforward though there are a few subtle edge cases, but it's more just implementation details.

## Checklist

## Related issues
Closes [#1711 ](https://github.com/wemake-services/wemake-python-styleguide/issues/1711)


Elton-Lin:
I'm passing all but one. I could not reproduce the error reported by wps/build  locally. Also, I feel like it could be a false positive since there's more code after the if block. I understand that why it's flagged as implicit elif, but can we just have that ignored? Thanks.

wemake_python_styleguide/visitors/tokenize/comments.py
```diff
@@ -51,6 +52,16 @@ def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
         self._no_cover_count = 0
 
+        # Variables for WPS#463 EmptyCommentViolation
+        self._prev_comment_line_num = -1
+        self._prev_non_empty = -1
+        self._in_same_block = True
+        self._block_alerted = False
+        self._reserved_token = tokenize.TokenInfo(
```
sobolevn, line 60: Can you please explain this variable?

wemake_python_styleguide/visitors/tokenize/comments.py
```diff
@@ -60,9 +71,18 @@ def visit_comment(self, token: tokenize.TokenInfo) -> None:
             WrongMagicCommentViolation
 
         """
+        self._line_num = token.start[0]
```
sobolevn, line 74: We try to keep the logic out of `visit_` methods

Elton-Lin: Sure, that's reasonable. The intention was to put that in a helper function, but adding one more function makes this class exceed the default limit of number of functions/methods in one class. If we put it back into ``_check_empty_comment``, it would violate this rule (WPS231), exceeding the limit of code complexity. What is your suggestion?

sobolevn: Let's create a new visitor than! 👍

wemake_python_styleguide/visitors/tokenize/comments.py
Elton-Lin: Our approach is parsing a block of comment at once since there should only be at most one error alerted per block. The only time``self._in_same_block`` is false is when we are reading a new block, i.e. the first line of a comment block. (If it's empty, we know to alert immediately) ``self._block_alerted`` is used to prevent more than one alert within the same block.

``self._prev_non_empty`` is used to keep track of the line number of previous non-empty comment. It is mostly for this kind of situations: 
```python
# Non-empty 1
# Non-empty 2
#
#
# We are not here yet, the existence of this line affects whether we alert on line 3 or not.
```
When reading line 3, the token for line 3 will be assigned to ``self._reserved_token`` for a potential alert. And line 4 will be ignored because it is not right after a non-empty comment. (We did ``current line number - self._prev_non_empty`` == 1 to check). It is "reserved" because we should only alert if there is no non-empty comment in the same block later.

wemake_python_styleguide/visitors/tokenize/comments.py
Elton-Lin: Sounds good! Just making sure, this means we will make a new class within ``comments.py`` and we will call it like ``EmptyCommentVisitor()``? Thanks

sobolevn: Yes 👍

wemake_python_styleguide/visitors/tokenize/comments.py
```diff
@@ -60,9 +71,18 @@ def visit_comment(self, token: tokenize.TokenInfo) -> None:
...
+        self._in_same_block = (
+            self._line_num - self._prev_comment_line_num == 1 and  # is consec
```
skarzi, line 76: NIP: does `consec` mean `consecutive`? if so:

```suggestion
            self._line_num - self._prev_comment_line_num == 1 and  # is consecutive
```

You can even make some properties for this and the next one expressions when a new visitor will be created

Elton-Lin: Thanks for the review. Could you elaborate on "make some properties for this and the next one expressions"? Do you mean having a variable like ``is_consecutive = (self._line_num - self._prev_comment_line_num == 1)``?

skarzi: I mean `@property` but having a well-named variable also sounds good :+1:

wemake_python_styleguide/visitors/tokenize/comments.py
```diff
@@ -98,6 +99,79 @@ def _post_visit(self) -> None:
...
+                self._has_reserved = True
+                self._reserved_token = token
+        else:
+            self._prev_non_empty = self._line_num
+            if self._in_same_block:
+                self._has_reserved = False
+
+        self._prev_comment_line_num = token.start[0]
+
+    def _check_same_block(self, token: tokenize.TokenInfo):
```
skarzi, line 159:
```suggestion
    def _check_same_block(self, token: tokenize.TokenInfo) -> None:
```

wemake_python_styleguide/visitors/tokenize/comments.py
```diff
@@ -98,6 +99,79 @@ def _post_visit(self) -> None:
...
+        self._in_same_block = (
+            self._is_consecutive(self._prev_comment_line_num) and
+            token.line.lstrip()[0] == '#'  # is inline comment
+        )
+        if not self._in_same_block:
+            self._block_alerted = False
+
+    def _is_consecutive(self, prev_line_num):
```
skarzi, line 167:
```suggestion
    def _is_consecutive(self, prev_line_num: int) -> None:
```

wemake_python_styleguide/visitors/tokenize/comments.py
```diff
@@ -98,6 +99,79 @@ def _post_visit(self) -> None:
...
+
+    def __init__(self, *args, **kwargs) -> None:
+        """Initializes fields to track empty comments."""
+        super().__init__(*args, **kwargs)
+
+        self._prev_comment_line_num = -1
+        self._prev_non_empty = -1
+        self._in_same_block = True
+        self._block_alerted = False
+        self._has_reserved = False
```
skarzi, line 114: Where is `_reserved_token` defined/declared? It's good practice to define/declare all variables modified in methods etc in `__init__`

Elton-Lin: I see. I just thought initializing it with a token object is strange and setting it to ``None`` is also prohibited. Anyway, I still added back the initialization.

skarzi: It's just a quick thought, but [`TokenInfo`](https://github.com/python/cpython/blob/cfb2f796342c224599eeab847e8593343a775408/Lib/tokenize.py#L45) is `namedtuple`, that means it's immutable, so we can define some constant like `EMPTY_TOKEN` or `SENTINEL_TOKEN` with some "empty"/"dummy" `TokenInfo` value, that shouldn't be created normally by `tokenize` and assign it to `_reserved_token` in `__init__`.
Maybe this will also make `_has_reserved` optional because you can always compare `_reserved_token` with `EMPTY_TOKEN`.

Elton-Lin: I like that idea! It seems better for clarity and readability. I'll look into that in a bit.

wemake_python_styleguide/visitors/tokenize/comments.py
```diff
@@ -37,6 +37,10 @@
 )
 from wemake_python_styleguide.visitors.base import BaseTokenVisitor
 
+EMPTY_TOKEN = tokenize.TokenInfo(
```
skarzi, line 40: let's use `Final` from `typing` to indicate that this variable shouldn't be reassigned:

```suggestion
EMPTY_TOKEN: Final = tokenize.TokenInfo(
```

wemake_python_styleguide/visitors/tokenize/comments.py
```diff
@@ -37,6 +37,10 @@
...
+    type=0, string='', start=(0, 0), end=(0, 0), line='',
```
skarzi, line 41: NIP: token types are defined as constants in `token` module, so just for readability, `token.ENDMARK` can be used instead of `0`

wemake_python_styleguide/visitors/tokenize/comments.py
skarzi: What do you say for the `SENTINEL_TOKEN` name? This constant is mostly used as a sentinel in the end

skarzi (commented):
Nice work, code is really succinct now :+1:

wemake_python_styleguide/visitors/tokenize/comments.py
Elton-Lin: Looks like ``Final`` from ``typing`` is a ``3.8`` feature. How should I proceed?

wemake_python_styleguide/visitors/tokenize/comments.py
Elton-Lin: Sure, is doing ``from token import ENDMARK`` ok? I had to do it to avoid name clash with ``token`` being used as an argument to almost all functions.

wemake_python_styleguide/visitors/tokenize/comments.py
skarzi: Ahh, yes, so let's use `typing_extensions` for compatibility with older python versions:

```python
from typing_extensions import Final
```

wemake_python_styleguide/visitors/tokenize/comments.py
skarzi: IMHO it's fine to import `ENDMARK` from `token` as you wrote :+1:

wemake_python_styleguide/visitors/tokenize/comments.py
```diff
@@ -37,8 +38,12 @@
 )
 from wemake_python_styleguide.visitors.base import BaseTokenVisitor
 
-EMPTY_TOKEN = tokenize.TokenInfo(
-    type=0, string='', start=(0, 0), end=(0, 0), line='',
+SENTINEL_TOKEN: Final = tokenize.TokenInfo(
+    type=ENDMARKER,
+    string='',
+    start=(ENDMARKER, ENDMARKER),
+    end=(ENDMARKER, ENDMARKER),
```
skarzi, line 45: let's use `0` here because as far I know it's for row and column, so `ENDMARKER` usage is quite confusing here

Elton-Lin: Opps, I misunderstood your last comment. Fixed now.

skarzi (approved):
thanks a lot for super contribution :+1:

skarzi:
only one small thing have left - https://github.com/wemake-services/wemake-python-styleguide/pull/1747/checks?check_run_id=1476340314:

> ./wemake_python_styleguide/visitors/tokenize/comments.py:0:1: WPS226 Found string constant over-use: '' > 3

@sobolevn should we create some constant like `EMPTY_STRING: Final = ''` or this is some issue with `WPS226` and `''`?

sobolevn:
Yes, please add some constants for strings. And that's it! 👍

sobolevn (approved):
Thanks!

sobolevn:
Awesome work! And really useful feature. Thanks!

sobolevn:
🎉

Elton-Lin:
No problem! It was a great experience working on this feature and project with you guys. (@huangyunzen and I both enjoyed it.) Thanks @sobolevn and @skarzi !

PR #1760, opened 2020-12-05
--allowed-no-effect-binary-operators implemented

### LINKED ISSUE #1737: Make WPS428 configurable (Airflow example)
# Rule change request

Make [WPS428](https://wemake-python-stylegui.de/en/latest/pages/usage/violations/best_practices.html#wemake_python_styleguide.violations.best_practices.StatementHasNoEffectViolation) (Found statement that has no effect) configurable.

## Thesis

Let WPS428 treat some operators as having side effects and not violating the original rule.

## Reasoning

WPS introduces rule `WPS428: Found statement that has no effect`. The majority of Airflow-based codebases violate this rule because of using `>>` operator to define dependencies between tasks.

Please see the last line of [the following example](https://airflow.apache.org/docs/stable/tutorial.html#example-pipeline-definition) from the official documentation: `t1 >> [t2, t3]`. It defines that both tasks `t2` and `t3` are run only when `t1` is completed. And this line violates WPS428 since its result is not assigned to anything.

There is also a method-style way to define dependencies: `t2.set_upstream(t1)`, `t3.set_upstream(t1)`. However, [bitshift composition is recommended](https://airflow.apache.org/docs/stable/concepts.html#bitshift-composition) and is more common.

I don't want to disable this check completely and would rather like to ignore bitshift operators only since they are very uncommon in Airflow-based codebases for purposes other than setting up dependencies.

--

Also huge thanks for you work! It has dramatically simplified teaching my team to use best coding practices without much participation from my side!

bryzgaloff:
This PR introduces `--allowed-no-effect-binary-operators` configuration option which allows to ignore StatementHasNoEffectViolation for selected binary operators. As stated in the issue, this is useful to disable the check for binary shift operators (`>>` and `<<`) in specific domains like Airflow DAG codebases: these operators are used to define dependencies between tasks and a hardly ever used as numbers-related shift operators (e.g. `1 << 2`). This allows to safely disable check for such codebase.

I need some help with the following:
* Current code violates WPS rules and does not pass flake8 check due to the following reasons:
  * A new added test (`test_ignored_statement_with_no_effect`) is the 8th function within `tests/test_visitors/test_ast/test_statements/test_statements_with_no_effect.py` file. Maximum number of module members is set to 7 (WPS202). I have tried to move the test to a separate module but I need all these templates defined in the file (`module_template`, `if_template`, etc, parameterized by pytest). And the full module name is too long (`from tests.test_visitors.test_ast.test_statements.test_statements_with_no_effect` is already >80 characters long). So what should I do? Put this file to per-file-ignores? Or still move the test to a dedicated file and copy-paste templates? They are used by all the tests in `test_statements_with_no_effect.py` file.
  * The new option parameterizes `StatementsWithBodiesVisitor` so I had to add `__init__` method and it is the 8th method within `StatementsWithBodiesVisitor` class. Maximum class members is set to 7 (WPS214). How should I avoid the violation? Create a mixin class with `allowed_no_effect_binary_operators`-related logic? Or use a `noqa` comment on a class?
* Finally I want `--allowed-no-effect-binary-operators` option to be in sync with set of available values defined in `StatementsWithBodiesVisitor._string_to_binary_operator`. Currently if unsupported operator is provided, a `KeyError` is raised at `StatementsWithBodiesVisitor.__init__` which is unobvious. Any ideas on how to do that more user-friendly? `--allowed-no-effect-binary-operators` seems to be the first string-typed comma-separated option with fixed set of values so I have not seen an example on how to implement this.
* Should I manually change html docs? Or are they built automatically? I have found nothing related in [the contribution guide](https://wemake-python-stylegui.de/en/latest/pages/api/contributing.html#before-submitting). Or is it what mentioned in action point no.5? I have changed docstrings but not html files.

## Checklist

## Related issues

- Closes #1737

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ FROM python:3.8.6-alpine
 LABEL maintainer="sobolevn@wemake.services"
 LABEL vendor="wemake.services"
 
-ENV WPS_VERSION='0.14.1'
+ENV WPS_VERSION='0.14.2'
 ENV REVIEWDOG_VERSION='v0.11.0'
 
 RUN apk add --no-cache bash git wget
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ style = "styles/nitpick-style-wemake.toml"
 
 [tool.poetry]
 name = "wemake-python-styleguide"
-version = "0.14.1"
+version = "0.14.2"
 description = "The strictest and most opinionated python linter ever"
 
 license = "MIT"
--- a/wemake_python_styleguide/violations/best_practices.py
+++ b/wemake_python_styleguide/violations/best_practices.py
@@ -1202,8 +1202,13 @@ class StatementHasNoEffectViolation(ASTViolation):
             8 + 2
             print
 
+    Configuration:
+        This rule is configurable with `--allowed-no-effect-binary-operators``.
+        Default:
+        :str:`wemake_python_styleguide.options.defaults.ALLOWED_NO_EFFECT_BINARY_OPERATORS`
+
     .. versionadded:: 0.5.0
-    .. versionchanged:: 0.11.0
+    .. versionchanged:: 0.14.2
     """
 
     error_template = 'Found statement that has no effect'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,14 @@ Semantic versioning in our case means:
 - Updates `reviewdog` version to `0.11.0` and adds `action-depup`
 
 
+## 0.14.2
+
+### Features
+
+- Adds an option: `--allowed-no-effect-binary-operators`, which allows ignoring
+  no effect violation for selected binary operators #1737.
+
+
 ## 0.14.0 aka The Walrus fighter
 
 This release was focused on adding `python3.8` support,
--- a/tests/test_visitors/test_ast/test_statements/test_statements_with_no_effect.py
+++ b/tests/test_visitors/test_ast/test_statements/test_statements_with_no_effect.py
@@ -177,6 +177,16 @@ async def container():
     '3 > 4',
     '1 + 2',
     '-100',
+    'x - 1',
+    'x * 1',
+    'x / 1',
+    'x ** 1',
+    'x ^ 1',
+    'x % 1',
+    'x >> 1',
+    'x << 1',
+    'x & 1',
+    'x | 1',
 ])
 def test_statement_with_no_effect(
     assert_errors,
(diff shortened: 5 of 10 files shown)


Dockerfile
```diff
@@ -20,7 +20,7 @@ FROM python:3.8.6-alpine
 LABEL maintainer="sobolevn@wemake.services"
 LABEL vendor="wemake.services"
 
-ENV WPS_VERSION='0.14.1'
+ENV WPS_VERSION='0.14.2'
```
sobolevn, line 23: There's no such version.

pyproject.toml
```diff
@@ -8,7 +8,7 @@ style = "styles/nitpick-style-wemake.toml"
 
 [tool.poetry]
 name = "wemake-python-styleguide"
-version = "0.14.1"
+version = "0.14.2"
```
sobolevn, line 11: The same, please, don't bump this version.

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -1202,8 +1202,13 @@ def some_function():
...
             print
 
+    Configuration:
+        This rule is configurable with `--allowed-no-effect-binary-operators``.
+        Default:
+        :str:`wemake_python_styleguide.options.defaults.ALLOWED_NO_EFFECT_BINARY_OPERATORS`
+
     .. versionadded:: 0.5.0
-    .. versionchanged:: 0.11.0
+    .. versionchanged:: 0.14.2
```
sobolevn, line 1211: `.. versionchanged:: 0.15.0`
And an extra empty line.

sobolevn (commented):
Are there any other use-cases we want to support? 
Maybe some other frameworks or other `AST` nodes?

bryzgaloff:
> Are there any other use-cases we want to support?
> Maybe some other frameworks or other `AST` nodes?

I am not aware of other frameworks. The implementation I have provided is extensible enough so if someone decides to add some new AST nodes being concerned about specific case, they may do it.

@sobolevn Could you please give any advices on the questions I have asked in the "I need some help with the following" section of [PR description](https://github.com/wemake-services/wemake-python-styleguide/pull/1760#issue-533071186)?

Once we decide what to do with it, I will apply all the required changes.

sobolevn:
The only thing I dislike here is that `--allowed-no-effect-binary-operators` is very specific to your use-case.
Maybe we can come up with some other configuration format? 

We can consider using `astpath`: https://github.com/hchasestevens/astpath 
or something similar. Or even write our own tool.

bryzgaloff:
> The only thing I dislike here is that `--allowed-no-effect-binary-operators` is very specific to your use-case.
> Maybe we can come up with some other configuration format?

I believe it covers a vast majority of cases. The only uncovered syntaxes are unary operators (`+x`) and getting attributes (`x.attr` which may have side effects if `attr` is a `property`). Any others? Maybe literals (`{}`, `()`, etc) but I don't really believe this is the case.

Do you insist on implementing any of this within current PR? I can add attribute getter to a list of ignorable operators as `.` but it will require changing argument name to… what? `getattr` is a sort of binary operator, but not exactly. Or a separate option like `--allow-getattr-no-effect`? I suggest leaving this for a separate PR. Combining binary and unary operators under current format is not doable in an intuitive way, so I suggest leaving this for the future enhancement if anyone finds it useful. Or decide on other easy to remember format (e.g. `_+` as unary plus, means left operand is absent).

> We can consider using `astpath`: https://github.com/hchasestevens/astpath
> or something similar. Or even write our own tool.
> 
> Imaginary example: `--allow-no-effect-expressions='.//BinOp/op/BitOr|RShift'`
> It will allow only `BinOp` operators with `|` and `>>` inside.

This seems to me to be an overcomplication since the number of cases is quite limited and may have a more intuitive fotmat. Also I have looked through `astpath` codebase: unfortunately, it is very XML-related and we cannot easily extract `ASTPath -> ast.operator` implementation so we may only borrow an idea, not a convenient tool.

Any suggestions on the questions I have asked in the "I need some help with the following" section of PR description? Or may I just put noqa's?

bryzgaloff:
@sobolevn please see my reply: https://github.com/wemake-services/wemake-python-styleguide/pull/1760#issuecomment-741750933

sobolevn:
> A new added test (test_ignored_statement_with_no_effect) is the 8th function within tests/test_visitors/test_ast/test_statements/test_statements_with_no_effect.py file. Maximum number of module members is set to 7 (WPS202). I have tried to move the test to a separate module but I need all these templates defined in the file (module_template, if_template, etc, parameterized by pytest). And the full module name is too long (from tests.test_visitors.test_ast.test_statements.test_statements_with_no_effect is already >80 characters long). So what should I do? Put this file to per-file-ignores? Or still move the test to a dedicated file and copy-paste templates? They are used by all the tests in test_statements_with_no_effect.py file.

In this case we use per-folder `conftest.py` with new fixtures.

> The new option parameterizes StatementsWithBodiesVisitor so I had to add __init__ method and it is the 8th method within StatementsWithBodiesVisitor class. Maximum class members is set to 7 (WPS214). How should I avoid the violation? Create a mixin class with allowed_no_effect_binary_operators-related logic? Or use a noqa comment on a class?

Options proccessing should not happen this late. Let's process them somewhere in `options/`. At this moment you should already have the working `self.options._attr` object.

> Any ideas on how to do that more user-friendly? 

A `TypedDict` maybe?

>     Should I manually change html docs? Or are they built automatically? I have found nothing related in the contribution guide. Or is it what mentioned in action point no.5? I have changed docstrings but not html files.

They are pre-built from the comments. See: https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/options/config.py#L59

What is more important, I am still not convinced that adding `--allowed-no-effect-binary-operators` is a good thing.
For the reasons I have already emphasised: mainly be too specific.

I would probably love to do some more research based on `astpath`. Because it looks like a good chance to get very powerful tool to ignore different violations based on semantics and not location.

sobolevn:
I am going to close this one, because I would love to see something based on `astpath`.
It would make this feature totally awesome!

sobolevn:
Anyways, thanks a lot for your idea and implementation 👍 💪

Issue #1416, opened 2020-05-24
WPS440/441 on multiple loops


Dreamsorcerer:
With code like:
```
        for widget in self._widgets:
            await widget.update()
        ...
        for widget in self._widgets:
            await widget.set_something(state)
```

I get WPS440 and WPS441 due to reusing the widget variable name in the second loop.

I think that WPS441 shouldn't apply, as we are defining a new variable, as opposed to reusing the previous value from the last loop.

Then I think also, that WPS440 shouldn't apply, as the WPS441 stops you from reusing the first variable anyway, so shadowing a variable you're not supposed to be using is not an issue. i.e. If WPS441 is going to make you treat them as block-scoped variables, then shadowing them shouldn't be taken into account as they shouldn't be treated as being in the enclosing scope.

anatoly-scherbakov:
Another example of the same issue with `except` blocks.

```python
        try:
            sqs_response = self.client.send_message(
                QueueUrl=self.url,
                MessageBody=message_body,
            )

        except self.client.exceptions.QueueDoesNotExist as err:
            raise SQSQueueDoesNotExist(queue=self) from err

        except self.client.exceptions.ClientError as err:
            if self._error_code_is(err, 'InvalidParameterValue'):
                raise MessageTooLarge(
                    max_supported_size=MAX_MESSAGE_SIZE,
                    message_body=message_body,
                )
```

cause 440 and 441:

```
  38:9     WPS440 Found block variables overlap: err
  except self.client.exceptions.ClientError as err:
  ^

  39:36    WPS441 Found control variable used after block: err
  if self._error_code_is(err, 'InvalidParameterValue'):
```

huangyunzen:
Hello, @Elton-Lin and I would like to work on fixing this. Thanks!

sobolevn:
Awesome! Thanks!

huangyunzen:
> Another example of the same issue with `except` blocks.
> 
> ```python
>         try:
>             sqs_response = self.client.send_message(
>                 QueueUrl=self.url,
>                 MessageBody=message_body,
>             )
> 
>         except self.client.exceptions.QueueDoesNotExist as err:
>             raise SQSQueueDoesNotExist(queue=self) from err
> 
>         except self.client.exceptions.ClientError as err:
>             if self._error_code_is(err, 'InvalidParameterValue'):
>                 raise MessageTooLarge(
>                     max_supported_size=MAX_MESSAGE_SIZE,
[...]

PR #1768, opened 2020-12-10
Fixed WPS441 to allow variable names to be reused in multiple loops

huangyunzen:
# I have made things!

With code like:

```
         for widget in self._widgets:
             await widget.update()
         ...
         for widget in self._widgets:
             await widget.set_something(state)
```
 
WPS441 is triggered due to reusing the widget variable name in the second loop.
 
WPS441 shouldn't apply, as we are defining a new variable, as opposed to reusing the previous value from the last loop.

## Checklist

## Related issues
#1416


wemake_python_styleguide/visitors/ast/blocks.py
```diff
@@ -255,7 +255,8 @@ def _add_to_scope(self, node: ast.AST, names: Set[str]) -> None:
     def _check_variable_usage(self, node: ast.Name) -> None:
         context = cast(ast.AST, get_context(node))
         blocks = self._block_variables[context][node.id]
-        if all(is_contained_by(node, block) for block in blocks):
+
+        if not blocks or any(is_contained_by(node, block) for block in blocks):
```
sobolevn, line 259: Let's create a new variables out of this condition. I am a bit lost reading it. 
We need some extra meta context.

sobolevn (changes requested):
Thanks!

wemake_python_styleguide/visitors/ast/blocks.py
huangyunzen: Added a new variable and some context to understand it.

tests/test_visitors/test_ast/test_blocks/test_control_variables.py
```diff
@@ -43,6 +43,14 @@ def wrapper():
     print(wrapper)
 """
 
+correct_for_multi_loops = """
```
sobolevn, line 46: Please, let's add some tests: I guess we need to test that `for` can reuse the variable this way. But other types of blocks cannot. Let's say that two `for` loops are ok, but a `for` loop followed by `with` block with the same variable names is not ok.

```python
wrong_loop_variable_reuse1 = """
def wrapper():
    for my_var in range(3):
        ...
    with some() as my_var:
        ...
"""
```

sobolevn (changes requested):
Thanks! I have looked through the tests and figured out that we don't have any tests that cover the case I have described above.

Let's add them! 👍

tests/test_visitors/test_ast/test_blocks/test_control_variables.py
huangyunzen: I have added the test you included, and added a variable that checks whether the block variable is reused in other types of blocks.

sobolevn (approved):
Thanks a lot!

PR #1771, opened 2020-12-11
 WPS465: Forbid to use `&` and `|` with booleans #1657 

### LINKED ISSUE #1657: Forbid to use `&` and `|` with booleans
# Rule request

## Thesis

Yesterday I saw this code: 

```python
my_result = (first > 0 & second != 0) | other is None
```

And I was 😮 

So, we need to check all operations around `&` and `|` and detect any booleans there:
- Explicit `True` and `False`
- Compares with `is` or `==` or `!=`
- Compares with `>`, `>=`, etc
- Logical operations `not`, `or`, `and`

If we can detect at least one of these operations, then we should raise a violation. 
Let's call it `BinaryAndBooleanMixupViolation`.

Note: other binary operations are still fine!
So, these lines should not raise:
- `first & second`
- `first | 10`
- etc

## Reasoning

This case indicates that a person confused `&` with `and` and `|` with `or`.
This can be the case if a person is comming from another language.

thammaji:
# I have made things!

WPS465: It is forbidden to use '|' and '&' (bitwise operators) with booleans. For example:

```
(x >= y) & True
(first > 0 & second != 0) | (other is None)
```

But usage of bitwise operators numbers, variables or with other binary operators is fine:

```
first & second
(5 + 324 + 35) & y
```
Usage of bitwise operators with explicit `True` or `False`, boolean operators (`and` `or`), unary operators (`not`), and comparisons (`<` `!=`) will be forbidden.

## Checklist

## Related issues

Closes #1657


tests/test_checker/test_noqa.py
```diff
@@ -237,6 +237,7 @@
     'WPS462': 1,
     'WPS463': 1,
     'WPS464': 0,  # logically unacceptable.
+    'WPS465': 0,  # logically unacceptable.
```
sobolevn, line 240: It is perfectly acceptable and should be present in `fixtures/noqa.py`

tests/test_visitors/test_ast/test_operators/test_bitwise_boolean_mixup.py
```diff
@@ -0,0 +1,56 @@
...
+from wemake_python_styleguide.violations.best_practices import (
+    BitwiseAndBooleanMixupViolation,
+)
+from wemake_python_styleguide.visitors.ast.operators import BitwiseOpVisitor
+
+# add more cases for more usages of mixup
+
+
+@pytest.mark.parametrize('expression', [
+    'True | False',
```
sobolevn, line 12: Let's also test `True | True`, `False & False`

tests/test_visitors/test_ast/test_operators/test_bitwise_boolean_mixup.py
```diff
@@ -0,0 +1,56 @@
...
+    '(x >= y) & True',
+    'first | (10 == second)',
+    '(11 != first) | (not False)',
+    '(1 or first) & (second and first)',
+    '(first > 0 & second != 0) | (other is None)',
+    '(5234 + 2345) & True',
```
sobolevn, line 18: What about `None`? Should it be counted?

tests/test_visitors/test_ast/test_operators/test_bitwise_boolean_mixup.py
```diff
@@ -0,0 +1,56 @@
...
+])
+def test_bitwise_boolean_mixup(
```
sobolevn, line 20: Can we have some tests without `()`? Or will the operator order break?

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -2504,3 +2506,36 @@ class EmptyCommentViolation(TokenizeViolation):
...
+
+    Solution:
+        Change bitwise operator to boolean operators.
+
+    Example::
+
+        # Correct:
+
+        first | 10
+        result = ((first > 0) && False)
```
sobolevn, line 2530: Just the first one is enough, please remove this line.

wemake_python_styleguide/visitors/ast/operators.py
```diff
@@ -268,3 +269,35 @@ def visit_NamedExpr(
...
+
+    def visit_BinOp(self, node: ast.BinOp) -> None:
+        """
+        Finds bad usage of bitwise operation with binary operation.
+
+        Raises:
+            BitwiseAndBooleanMixupViolation
+
+        """
+        self._check_BitOp(node)
```
sobolevn, line 286: Please, name it logically, not structurely. 

For example: `_check_logical_bitwise_operator`

wemake_python_styleguide/visitors/ast/operators.py
```diff
@@ -268,3 +269,35 @@ def visit_NamedExpr(
...
+        self.generic_visit(node)
+
+    def _check_BitOp(self, node: ast.BinOp) -> None:
+        if not isinstance(node.op, (ast.BitOr, ast.BitAnd)):
+            return
+
+        if (self._check_sides(node.left) or self._check_sides(node.right)):
```
sobolevn, line 293: you don't need `()` here.

wemake_python_styleguide/visitors/ast/operators.py
```diff
@@ -268,3 +269,35 @@ def visit_NamedExpr(
...
+            self.add_violation(BitwiseAndBooleanMixupViolation(node))
+
+    # checks either side of the Bitwise operation invalid usage
+    def _check_sides(self, node) -> bool:
+        invalid = False
```
sobolevn, line 298: We can refactor this into:

`return isinstance(node, self._invalid_nodes)`

Where `self._invalid_nodes` is a class-level constant with all the nodes we need. Check other visitor for examples, search `AnyNodes` type.

sobolevn (changes requested):
Awesome work! There's a bit of polishing required, but looks like we are almost ready!

tests/test_visitors/test_ast/test_operators/test_bitwise_boolean_mixup.py
thammaji: I think it should be counted as a violation (python will  throw an error when NoneType is used with Bitwise operators anyways).  Added `None` test case.

tests/test_visitors/test_ast/test_operators/test_bitwise_boolean_mixup.py
thammaji: The operator order will break, `1 or first & second and first` will not raise a violation, as the left and right of the BitwiseOp are just Name variables, and I don't think there is any way to traverse the AST from Name variables. This means that there will be some false negatives, but I don't think that there's a better way.

sobolevn: Ok 👍

wemake_python_styleguide/visitors/ast/operators.py
thammaji: Thanks for this suggestion, I was feeling that my code there was a little inelegant.

tests/fixtures/noqa/noqa.py
```diff
@@ -765,3 +765,5 @@ def many_raises_function(parameter):  # noqa: WPS238
 
 def get_item():  # noqa: WPS463
     return  # noqa: WPS324
+
+bad_bitwise = True | True # noqa: WPS465
```
sobolevn, line 769: Should end with a new line

wemake_python_styleguide/visitors/ast/operators.py
```diff
@@ -283,21 +290,16 @@ def visit_BinOp(self, node: ast.BinOp) -> None:
...
+    def _check_logical_bitwise_operator(self, node: ast.BinOp) -> None:
         if not isinstance(node.op, (ast.BitOr, ast.BitAnd)):
             return
 
-        if (self._check_sides(node.left) or self._check_sides(node.right)):
+        if self._check_sides(node.left) or self._check_sides(node.right):
             self.add_violation(BitwiseAndBooleanMixupViolation(node))
 
     # checks either side of the Bitwise operation invalid usage
     def _check_sides(self, node) -> bool:
```
sobolevn, line 304: Let's better name it `_is_bool_like`. What do you think?

thammaji: Sure, that's a would be more descriptive of what the function actually does.

wemake_python_styleguide/visitors/ast/operators.py
```diff
@@ -268,3 +269,37 @@ def visit_NamedExpr(
...
+
+    def _check_logical_bitwise_operator(self, node: ast.BinOp) -> None:
+        if not isinstance(node.op, (ast.BitOr, ast.BitAnd)):
+            return
+
+        if self._is_bool_like(node.left) or self._is_bool_like(node.right):
+            self.add_violation(BitwiseAndBooleanMixupViolation(node))
+
+    # checks either side of the Bitwise operation invalid usage
+    def _is_bool_like(self, node) -> bool:
```
sobolevn, line 304:
```suggestion
    def _is_bool_like(self, node) -> bool:
    def _is_bool_like(self, node: ast.expr) -> bool:
```

sobolevn (approved):
Awesome, thank you! Very cool feature 👍

thammaji:
I had a good experience working on it! Thanks for being so friendly and welcoming!

sobolevn:
Thanks 😊

PR #1792, opened 2020-12-22
do not raise ImplicitEnumerateViolation if range called with step

### LINKED ISSUE #1742: False positive ImplicitEnumerateViolation on range with step
# Bug report

## What's wrong
A simple function that splits a list into chunks 
```python
def chunks(lst, amount):
    for index in range(0, len(lst), amount):
        yield lst[index:index + amount]
```
will raise `ImplicitEnumerateViolation`. 

Documentation suggests using `enumerate(lst)` instead of `range(0, len(list))` but there is no `step`

## How it should be
`ImplicitEnumerateViolation` should not raise if `step > 0` in range arguments.

AlwxSin:
closes #1742

# I have made things!

## Checklist

## Related issues
Closes #1742


sobolevn (approved):
Thanks!

PR #1867, opened 2021-02-11
Closes #1863

### LINKED ISSUE #1863: Some async comprehensions are reported incorrectly
Currently our `InconsistentComprehensionViolation` fails on some `async` comprehensions. 
We need to fix that!

Reproduction:

```
async def wrapper():
    (
        some(number)
        async for number in matrix
        if number > 0
    )
```

Result:

```
  4:15     WPS361 Found an inconsistently structured comprehension
  async for number in matrix
        ^
```

sobolevn:
# I have made things!


tests/test_visitors/test_tokenize/test_statements/test_inconsistent_comprehension.py
```diff
@@ -11,26 +11,61 @@
 
 # List comprehension tests
 
-correct_list_empty = """
-a = []
+correct_list_empty = '[]'
+
+correct_list_one_line_comprehension1 = """
+def wrapper():
```
skarzi, line 17: why it's necessary to add `def wrapper():`?
Maybe it can be added in the test to make test cases declaration less verbose?

sobolevn: See `mode` fixture: https://github.com/wemake-services/wemake-python-styleguide/blob/master/tests/plugins/async_sync.py

PR #1898, opened 2021-02-22
Add exception for ImplicitEnumerateViolation

### LINKED ISSUE #1883: Invalid WPS518 Found implicit enumerate() call outside for loop
# Bug report

## What's wrong

The following code generates a WPS518 violation

```python
combs = combinations(
    range(len(coords)),
    degree,
)
```

## How it should be

I think that it should not report an error because the alternative

```github
combs = combinations(
    i, _ for i in enumerate(coords),
    degree,
)
```
is more verbose and less performant.

cbows:
Allow using `range(len(...))` as first argument of `itertools.combinations`.

## Checklist

## Related issues

Closes #1883

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Semantic versioning in our case means:
 ### Bugfixes
 
 - Fixes `BitwiseAndBooleanMixupViolation` work with PEP 604 union types #1884
+- Fixes `ImplicitEnumerateViolation` raised outside of for loops
 
 ## 0.15.1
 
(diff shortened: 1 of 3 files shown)


sobolevn (commented):
Thanks! Let's also fix https://github.com/wemake-services/wemake-python-styleguide/issues/1883

I guess it is pretty related! 👍

cbows:
That's actually what I wanted to fix

cbows:
Do you think the underlying issue is more generic than with just `itertools.combinations` ? Because this PR basically only handles this one case, but other functions could be added to that list of exceptions.

sobolevn:
Yes, we should check where `range(len(...))` is used. Because, there are a lot of use-cases when person wants to use it.
And we should only care for cases inside the `for` loop / comprehension. Other usages should be fine.

cbows:
ok makes sense. i will work on an update

PR #1906, opened 2021-02-24
Git ignore asdf .tool-versions file

cbows:
Asdf is a version manager that can be used for python and other programming
languages and tools. It is similar to pyenv in that regard. It uses a
`.tool-versions` file to specify versions of tools used in a project. Since
pyenv's equivalent, `.python-version`, is ignored, `.tool-versions` should be
ignored as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,6 +76,9 @@ target/
 # pyenv
 .python-version
 
+# asdf
+.tool-versions
+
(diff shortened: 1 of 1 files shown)


sobolevn (approved):

sobolevn:
Thanks!

Issue #1733, opened 2020-11-19
Forbid assigning `lambda` to object properties


sobolevn:
# Rule request

## Thesis

We should forbid code like: 

```python
class Example(object):
    def __init__(self):
        self.callback = lambda arg: arg + 1
```

## Reasoning

This will be very tricky to understand for users: `example.callback(1)` will look like a method, but it is not a method.

abhi-krish:
Hi! I'd like to work on this.

sobolevn:
Awesome! Thanks a lot!

abhi-krish:
Hey @sobolevn! This is my first time contributing and am not sure where I should start with this problem. Would you be able to point me in the right direction?

sobolevn:
Sure!

Start with reading this file: https://github.com/wemake-services/wemake-python-styleguide/blob/master/CONTRIBUTING.md
Then look at: https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/visitors/ast/classes.py

willtkwon:
Hi, can I work on this task?

sobolevn:
@willtkwon Please feel free! And thank you 🙂

AlexandrKhabarov:
Is it correct that this check should find situations like this?
```python
create_object.attr = lambda: print('Lambda!')
```
I mean that it should not distinguish lambda assignemnts inside __init__ or anywehere.

sobolevn:
We only treat `self`, `cls`, and `mcs` as self-like objects. So, we only check them. See https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/constants.py#L150

PR #1911, opened 2021-02-25
1733 add lambda instance assignment violation

AlexandrKhabarov:
# I have made things!

## Checklist


AlexandrKhabarov:
@sobolevn 
Hello, Nikita!
Can you look at my request?

tests/test_visitors/test_ast/test_classes/test_instance_assignment.py
```diff
@@ -0,0 +1,28 @@
...
+    InstanceLambdaAssignmentViolation,
+)
+from wemake_python_styleguide.visitors.ast.classes import (
+    InstanceAssignmentVisitor,
+)
+
+instance_lambda_assignment = """
+class Example(object):
+    def __init__(self):
+        self._attr = lambda: ...
```
sobolevn, line 11: We need to make this `_attr` thing here parametrized.
We need to check: `attr`, `_attr`, and `__attr`

We also need to parametrize `self` part.
We need to check: `self`, `cls`, `mcs`, `other`

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -2572,3 +2574,35 @@ def my_function(): ...
...
+class InstanceLambdaAssignmentViolation(ASTViolation):
+    """
+    Forbid lambda assignment to instance attributes.
+
+    Reasoning:
+       This will be very tricky to understand for users.
+       example.callback(1) will look like a method, but it is not a method.
+
+    Solution:
+        To forbid lambda assignment in instance attributes.
```
sobolevn, line 2589: > Do not assign ``lambda`` functions to attributes, create real methods.

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -2572,3 +2574,35 @@ def my_function(): ...
...
+
+
+@final
+class InstanceLambdaAssignmentViolation(ASTViolation):
+    """
+    Forbid lambda assignment to instance attributes.
+
+    Reasoning:
+       This will be very tricky to understand for users.
+       example.callback(1) will look like a method, but it is not a method.
```
sobolevn, line 2586: > ``example.callback(1)``

> but it is not really a method

sobolevn (changes requested):
Thanks a lot! Great work! So many awesome features!

I have several things I want to discuss:
1. What do you think about `cls.` attribute assignment?
2. What do you think about `class Some: a = lambda: ...`?

sobolevn:
Closes #1733

sobolevn:
@AlexandrKhabarov please, don't forget to use `closes #x` in your PRs, so our automation will link them to original issues 🙂

tests/test_visitors/test_ast/test_classes/test_instance_assignment.py
AlexandrKhabarov: Done

wemake_python_styleguide/violations/best_practices.py
AlexandrKhabarov: Done

wemake_python_styleguide/violations/best_practices.py
AlexandrKhabarov: Done

AlexandrKhabarov:
Yeah, I think is a good point to check assignments to class instance too. I'll add it.

AlexandrKhabarov:
@sobolevn 
Hello, Nikita?
Can you review my request?

wemake_python_styleguide/logic/tree/assignments.py
```diff
@@ -0,0 +1,87 @@
...
+
+    if isinstance(node, ast.Tuple):
+        for subnode in node.elts:
+            flatten_nodes.extend(_flat_nodes(subnode))
+    else:
+        flatten_nodes.append(node)
+    return flatten_nodes
+
+
+def get_assignments(node: ast.ClassDef) -> _AllAssignments:
```
sobolevn, line 40: This function is really similar to https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/logic/tree/classes.py#L76 

Can you please exaplain, why can't we reuse it?

tests/test_visitors/test_ast/test_classes/test_instance_assignment.py
```diff
@@ -0,0 +1,65 @@
+import pytest
+
+from wemake_python_styleguide.violations.best_practices import (
+    InstanceLambdaAssignmentViolation,
+)
+from wemake_python_styleguide.visitors.ast.classes import (
+    AttributesAssignmentVisitor,
+)
+
+instance_lambda_assignment = """
```
sobolevn, line 10: We also need `cls` template:

```python
"""
class Example(object):
    @classmethod
    def some(cls):
         {0}
"""
```

sobolevn (changes requested):
Thanks a lot, @AlexandrKhabarov! 
As always, great work! 👍

wemake_python_styleguide/logic/tree/assignments.py
AlexandrKhabarov: I wrote another one function for finding assignments only. 
[This one](https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/logic/tree/classes.py#L76) finds assignments for classes and attributes for instances and I decided to unify returned values for lambda assignment violations.

I agree that both functions are similar and I need to think about how to decreases code duplication.

AlexandrKhabarov:
@sobolevn 
Hello Nikita!
I decreased code duplication and write one more tests for class lamda assignment.
Can you review my request?

wemake_python_styleguide/logic/tree/classes.py
```diff
@@ -70,35 +72,40 @@ def get_attributes(
...
-        isinstance(node.value, ast.Name) and
-        node.value.id == 'self'
+def _get_instance_assignment(subnode: ast.AST) -> Optional[AnyAssign]:
+    return subnode if (
+        isinstance(subnode, AssignNodes) and
+        any(
+            isinstance(target, ast.Attribute) and
+            isinstance(target.ctx, ast.Store) and
+            isinstance(target.value, ast.Name) and
+            target.value.id in constants.SPECIAL_ARGUMENT_NAMES_WHITELIST
```
sobolevn, line 101: This is not correct, because `constants.SPECIAL_ARGUMENT_NAMES_WHITELIST` also has `mcs` and `cls`, so it can catch non-self assigns.

wemake_python_styleguide/visitors/ast/blocks.py
```diff
@@ -76,7 +76,7 @@ class BlockVariableVisitor(base.BaseNodeVisitor):
     )
 
     _scope_predicates: Tuple[_ScopePredicate, ...] = (
-        lambda node, names: predicates.is_property_setter(node),
+        lambda node, names: predicates.is_property_setter(node),  # noqa: WPS467
```
sobolevn, line 79: This should be ignored, because here we assign `Tuple` and not `lambda`.

We should test it explicitly.

sobolevn (changes requested):
Thanks a lot! The progress is looking great! 👍 

Please, rebase your PR on `master`, since now it has some conflicts.

PR #1920, opened 2021-03-01
1009 Eval math operations

### LINKED ISSUE #1009: Eval math operations
We need to add a tool to eval pure math and bool operations.

For example, we can check that `1 + 2` is `3` and `"a" * 2` is `"aa"`
We need it because sometimes we restrict to use several datatypes in some places: like `float` in dict keys. But, people can actually write `float` numbers as math expressions: `0.1 + 5 * 10 ^ 2`. We have to parse this information and return result nodes. 

Algorithm:
- if all elements are constants - return the evaluated result
- if anything fails - return the wrapping node itself

Refs #861

Jrryy:
# I have made things!

## Checklist

## Related issues

- Closes #1009

### DIFF
--- a/wemake_python_styleguide/logic/nodes.py
+++ b/wemake_python_styleguide/logic/nodes.py
@@ -1,6 +1,10 @@
 import ast
-from typing import Optional
+from typing import Optional, Union
 
+from wemake_python_styleguide.logic.safe_eval import literal_eval_with_names
+from wemake_python_styleguide.transformations.ast.enhancements import (
+    evaluate_operation,
+)
 from wemake_python_styleguide.types import ContextNodes
 
 
@@ -26,3 +30,18 @@ def get_parent(node: ast.AST) -> Optional[ast.AST]:
 def get_context(node: ast.AST) -> Optional[ContextNodes]:
     """Returns the context or ``None`` if node has no context."""
     return getattr(node, 'wps_context', None)
+
+
+def evaluate_node(node: ast.AST) -> Optional[Union[int, float, str, bytes]]:
+    """Returns the value of a node or its evaluation."""
+    if isinstance(node, ast.BinOp):
+        return evaluate_operation(node)
+    if isinstance(node, (ast.Str, ast.Bytes)):
+        return node.s
+    if isinstance(node, ast.Name):
+        return None
+    try:
+        signed_node = literal_eval_with_names(node)
+    except Exception:
+        return None
+    return signed_node
--- a/wemake_python_styleguide/logic/nodes.py
+++ b/wemake_python_styleguide/logic/nodes.py
@@ -33,6 +33,8 @@ def evaluate_node(node: ast.AST) -> Optional[Union[int, float, str, bytes]]:
     """Returns the value of a node or its evaluation."""
     if isinstance(node, ast.Name):
         return None
+    if isinstance(node, (ast.Str, ast.Bytes)):
+        return node.s
     try:
         signed_node = literal_eval_with_names(node)
     except Exception:
--- a/wemake_python_styleguide/logic/nodes.py
+++ b/wemake_python_styleguide/logic/nodes.py
@@ -2,9 +2,6 @@ import ast
 from typing import Optional, Union
 
 from wemake_python_styleguide.logic.safe_eval import literal_eval_with_names
-from wemake_python_styleguide.transformations.ast.enhancements import (
-    evaluate_operation,
-)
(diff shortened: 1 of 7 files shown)


wemake_python_styleguide/transformations/ast/enhancements.py
```diff
@@ -11,6 +14,21 @@
     *FunctionNodes,
 )
 
+_AST_OPS_TO_OPERATORS = MappingProxyType({
```
sobolevn, line 17: This should be `Final`

wemake_python_styleguide/transformations/ast/enhancements.py
```diff
@@ -96,3 +138,38 @@ def _apply_if_statement(statement: ast.If) -> None:
             if child in statement.orelse:
                 setattr(statement, 'wps_if_chained', True)  # noqa: B010
                 setattr(child, 'wps_if_chain', statement)  # noqa: B010
+
+
+def _evaluate_node(node: ast.AST) -> Optional[Union[int, float, str, bytes]]:
```
sobolevn, line 143: Let's move this function to somewhere else, `logic` seems like a better place

wemake_python_styleguide/transformations/ast/enhancements.py
```diff
@@ -96,3 +138,38 @@ def _apply_if_statement(statement: ast.If) -> None:
...
+
+    if left and right and op:
+        try:
+            evaluation = op(left, right)
+        except Exception:
+            evaluation = None
+    else:
+        evaluation = None
+
+    setattr(statement, 'wps_op_eval', evaluation)  # noqa: B010
```
sobolevn, line 174: Let's move this side effect to `set_constant_evaluations`

sobolevn (changes requested):
Awesome! Thanks a lot!

How do you plan the next step? Should we enable this logic in at least a single check to test things? Or just merge as-is?

Jrryy:
> Should we enable this logic in at least a single check to test things? Or just merge as-is?

Good question. I do not think that this PR should modify other checks, but I guess it would make sense to do it to a simple one and completely test if this feature works, although I feel like the ones which would make use of this are not implemented yet.

sobolevn:
We have this rule that fits 100%:

```
  1:1      WPS449 Found float used as a key
  a[1.5]
  ^
```

Jrryy:
Great, I will use that one then

wemake_python_styleguide/transformations/ast/enhancements.py
Jrryy: That makes sense, but I did it this way because it becomes a recursive functionality which expands and saves the evaluation to all the branches of a big operation, so that they don't have to be evaluated again. Moving this to `set_constant_evaluations` will mean that the evaluation will be reexecuted for all operations inside and will slow the execution

Jrryy: Nevermind, I've already solved both these problems.

wemake_python_styleguide/logic/nodes.py
```diff
@@ -26,3 +27,14 @@ def get_parent(node: ast.AST) -> Optional[ast.AST]:
...
     """Returns the context or ``None`` if node has no context."""
     return getattr(node, 'wps_context', None)
+
+
+def evaluate_node(node: ast.AST) -> Optional[Union[int, float, str, bytes]]:
+    """Returns the value of a node or its evaluation."""
+    if isinstance(node, ast.Name):
+        return None
+    try:
+        signed_node = literal_eval_with_names(node)
```
sobolevn, line 37: This can return `ast.Name`. I guess - we don't expect that, do we?

Jrryy: No, `literal_eval_with_names` does not return `ast.Name`. If anything, it could return the id of the variable representing `ast.Name`, only if `node` is an instance of it, but I take care of that right before calling this function.

sobolevn: Oh, yes! Thanks you for reminding 🙂 
https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/logic/safe_eval.py#L16

Jrryy: May I then resolve this conversation?

sobolevn (approved):
Thanks! I will merge this feature into `0.16` soon!

PR #1927, opened 2021-03-07
Closes #1880

### LINKED ISSUE #1880: Constant condition is not detected when used with `and` or `or`
This code does not raise:

```python
if some or True: ...
```

This one raises:

```python
if True: ...
```

```
  1:4      WPS314 Found conditional that always evaluates the same
  if True: ...
```

uhbif19:
# I have made things!

This code only checks for unpacking in variable assignment case outlined in issue.

I do not cover cases like `for a, *_r in smthg` or `a, *_b, *_c` because it seems unlikely that somebody will use them.
If you think that my assumption is not correct, I will update the code.

## Checklist

## Related issues

- Closes #1880

## Feedback

Related code uses term `target` in two different senses.
One is for `target1 = target2 = ...`, the same like ast module does.
Second one is for `target1, target2 = ...', which conflicts with first sence.

I think that renaming second thing will clarify situation.


wemake_python_styleguide/violations/best_practices.py
```diff
@@ -2572,3 +2574,33 @@ def my_function(): ...
...
+        first = next(iter(collection))
+
+        # Wrong:
+        first, *_rest = some_collection
+
+    .. versionadded:: 0.16.0
+
+    """
+
+    error_template = 'Found incorrect unpacking target'
```
skarzi, line 2605: This message is identical to `WrongUnpackingViolation.error_template`, let's change it to something more related to this violation, e.g.:

```suggestion
    error_template = 'Found wrong retrieving of the first element'
```

wemake_python_styleguide/logic/tree/variables.py
```diff
@@ -54,3 +54,17 @@ def _is_valid_single(node: _VarDefinition) -> bool:
...
     )
+
+
+def is_getting_first_element_by_unpacking(targets: List[ast.expr]) -> bool:
+    """Checks if unpacking targets used to get first element."""
+    return (
+        len(targets) == 2 and
+        isinstance(targets[0], ast.Name) and
+        isinstance(targets[1], ast.Starred) and
+        _is_unused_variable_name(targets[1].value)
```
sobolevn, line 65: Let's refactor this part to use: https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/logic/naming/access.py#L9

sobolevn (changes requested):
Thanks a lot!

wemake_python_styleguide/logic/tree/variables.py
sobolevn: Oh, sorry. Let's use `is_protected` then

CHANGELOG.md
```diff
@@ -16,6 +16,12 @@ Semantic versioning in our case means:
   But, in the future we might change the configuration names / logic,
   change the client facing API, change code conventions signigicantly, etc.
 
+## 0.16.0
+
+### Features
+
+* Forbids getting first element of list by unpacking
```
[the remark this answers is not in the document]
uhbif19: Done.

Maybe if the project have MD style rules, it can use some linter? I found rule that seems to fix bullet list style: https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md004---unordered-list-style

I may work on adding linter to the project.

wemake_python_styleguide/violations/best_practices.py
uhbif19: Thanks, I forget to fix this after copy-paste. Done.

wemake_python_styleguide/logic/tree/variables.py
uhbif19: This does not work too:

```
>>> is_protected('_')
False

>>> is_protected('__')
False
```

While `access.is_unused(node.id) or access.is_protected(node.id)` works.

But for me such code looks misleading. Would you consider adding different "weak" kind of `is_unused` in naming.access?

sobolevn: Ok, then let's create a new function `access.looks_like_unused()`

uhbif19: Done.

sobolevn (approved):

skarzi (approved):
great work!

PR #1939, opened 2021-03-10
1731 Forbid returning too long tuples

### LINKED ISSUE #1731: Forbid returning too long tuples
# Rule request

## Thesis

Forbid returning a too long constant-length tuple  from a function:

```python
# bad
def hello():
  return 1, 2, 3, 4
```

When there are more than 2 values must be returned, use [dataclasses.dataclass](https://docs.python.org/3/library/dataclasses.html), [typing.NamedTuple](https://docs.python.org/3/library/typing.html#typing.NamedTuple), or [attrs](https://www.attrs.org/en/stable/) instead.

## Reasoning

Code readability and maintainability: the caller code must know exact order of all components which is easy to mess up. A real-world example:

```python
time, phase, spacing, size, version, error, series_type = read_result_log()
```

:weary:

uhbif19:
# I have made things!

This PR extends `TooLongYieldTupleViolation` realization this way:

* Violation is new
* Tests and visitor was extended to cover this violation
* Constant is new (`MAX_LEN_RETURN_TUPLE`). I am not sure how do they work, but doc says they are a public API. So I decided I should not rename constant to cover both cases, to not to break API.

## Checklist

## Related issues

- Closes #1731

### DIFF
--- a/wemake_python_styleguide/constants.py
+++ b/wemake_python_styleguide/constants.py
@@ -406,8 +406,8 @@ MAGIC_NUMBERS_WHITELIST: Final = frozenset((
 #: Maximum amount of ``pragma`` no-cover comments per module.
 MAX_NO_COVER_COMMENTS: Final = 5
 
-#: Maximum length of ``yield`` ``tuple`` expressions.
-MAX_LEN_YIELD_TUPLE: Final = 5
+#: Maximum length of ``yield`` or ``return`` ``tuple`` expressions.
+MAX_LEN_TUPLE_OUTPUT: Final = 5
 
 #: Maximum number of compare nodes in a single expression.
 MAX_COMPARES: Final = 2
--- a/wemake_python_styleguide/violations/complexity.py
+++ b/wemake_python_styleguide/violations/complexity.py
@@ -47,7 +47,7 @@ Summary
    TooManyForsInComprehensionViolation
    TooManyExceptCasesViolation
    OverusedStringViolation
-   TooLongYieldTupleViolation
+   TooLongOutputTupleViolation
    TooLongCompareViolation
    TooLongTryBodyViolation
    TooManyPublicAttributesViolation
@@ -89,7 +89,7 @@ Structure complexity
 .. autoclass:: TooManyForsInComprehensionViolation
 .. autoclass:: TooManyExceptCasesViolation
 .. autoclass:: OverusedStringViolation
-.. autoclass:: TooLongYieldTupleViolation
+.. autoclass:: TooLongOutputTupleViolation
 .. autoclass:: TooLongCompareViolation
 .. autoclass:: TooLongTryBodyViolation
 .. autoclass:: TooManyPublicAttributesViolation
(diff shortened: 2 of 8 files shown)


wemake_python_styleguide/constants.py
```diff
@@ -369,6 +369,9 @@
 #: Maximum length of ``yield`` ``tuple`` expressions.
 MAX_LEN_YIELD_TUPLE: Final = 5
 
+#: Maximum length of ``return`` ``tuple`` expressions.
+MAX_LEN_RETURN_TUPLE: Final = 5
```
sobolevn, line 373: I think that we should unify this setting into a single one. How about `MAX_LEN_TUPLE_OUTPUT`?

wemake_python_styleguide/violations/complexity.py
```diff
@@ -1249,3 +1251,23 @@ class TooManyRaisesViolation(ASTViolation):
 
     error_template = 'Found too many raises in a function: {0}'
     code = 238
+
+
+@final
+class TooLongReturnTupleViolation(ASTViolation):
```
sobolevn, line 1257: Let's rename the existing `Yield` violation. I think that there's no need to create a new one.

wemake_python_styleguide/constants.py
uhbif19: Okay. Should I rename `ReturnLikeStatementTupleVisitor` too?

wemake_python_styleguide/violations/complexity.py
uhbif19: Okay.

Are there any docs in which cases changing violations or some parts of public API is ok?

uhbif19: All done. I am not sure about naming, through.

sobolevn: > Are there any docs in which cases changing violations or some parts of public API is ok?

We don't consider our code to be a part of our API, because no one in their sane minds would use it. This is a CLI tool, so only UX changes are important.

tests/test_visitors/test_ast/test_complexity/test_counts/test_function_output_length.py
```diff
@@ -1,18 +1,21 @@
 import pytest
 
-from wemake_python_styleguide.constants import MAX_LEN_YIELD_TUPLE
+from wemake_python_styleguide.constants import MAX_LEN_TUPLE_OUTPUT
 from wemake_python_styleguide.violations.complexity import (
-    TooLongYieldTupleViolation,
+    TooLongFunctionOutputTupleViolation,
```
sobolevn, line 5: Let's name it `TooLongOutputTupleViolation`

sobolevn (commented):
👍 

Thanks!

wemake_python_styleguide/violations/complexity.py
uhbif19: I just am confused by the comment lines in `constants.py`. There are separation between constants public for "end-user" and internal which are not documented. So I do not get, what these lines are talking about if it is not about API. 

https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/constants.py#L25

tests/test_visitors/test_ast/test_complexity/test_counts/test_function_output_length.py
uhbif19: Done. Also renamed test file accordingly.

sobolevn (approved):
Awesome! Thanks!

I am going to release `0.15.3` and then merge this PR into `0.16`

sobolevn (approved):

PR #1996, opened 2021-04-22
Ignore WPS114 in notebook test

MarcoGorelli:
This should avoid the sporadic CI failure observed in #1995 - I ran the test several times on my machine and it always passes now


sobolevn (approved):
Thanks a lot for the fix 👍

PR #2025, opened 2021-05-11
Fix WPS516 documentation

dmitry-naumenko:
# I have made things!

Just correction of documentation (WPS516)


sobolevn (approved):
Thanks! Good catch!

PR #2072, opened 2021-06-22
Issue 1951 Check that generator expressions are safe

GibranHL0:
# I have made things!

## Checklist

## Related issues

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,6 +109,7 @@ Semantic versioning in our case means:
 - Forbids inconsistent structuring of multiline comprehensions
 - Forbids to use unpythonic getters and setters such as `get_attribute` or `set_attribute`
 - Now `credits`, `license`, and `copyright` builtins are free to shadow
+- Forbids the reassignment of variables that affect the generator expression
 
 ### Bugfixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,13 @@ Semantic versioning in our case means:
   change the client facing API, change code conventions signigicantly, etc.
   
   
+# 0.16.0
+
+### Features
+
+- Forbids the reassignment of variables that affect the generator expression
+
+
 ## 0.15.3 WIP
 
 ### Bugfixes
@@ -109,7 +116,6 @@ Semantic versioning in our case means:
 - Forbids inconsistent structuring of multiline comprehensions
 - Forbids to use unpythonic getters and setters such as `get_attribute` or `set_attribute`
 - Now `credits`, `license`, and `copyright` builtins are free to shadow
-- Forbids the reassignment of variables that affect the generator expression
 
 ### Bugfixes
 
--- a/tests/test_visitors/test_ast/test_loops/test_comprehensions/test_unsafe_generator_expression.py
+++ b/tests/test_visitors/test_ast/test_loops/test_comprehensions/test_unsafe_generator_expression.py
@@ -15,9 +15,9 @@ sum_result = sum(expression) # noqa: WPS363
 """
 
 no_variables = """
-a = 1
-b = (i for i in range(5))
-c = sum(b)
+first_value = 1
+result = (index for index in range(5))
+c = sum(result)
 """
 
 single_use = """
@@ -31,9 +31,42 @@ result = sum((first_value * i for i in range(5)))
 first_value = 2
 """
 
+unsafe_string = """
+line = ['child','tall']
+suffix = 'ish'
+converted_words = (word + suffix for word in line)
+suffix = 'ash'
+print(list(converted_words))
+"""
+
+safe_string = """
+line = ['hello', 'world']
+suffix = 'rar'
+converted_words = (word + suffix for word in line)
+print(list(converted_words))
+"""
+
+unsafe_change_index = """
+values = [1, 2, 3, 4, 5]
+exponent = 5
+power_expr = (value * exponent for value in values)
+exponent = 10
+print(list(power_expr))
+"""
+
+safe_change_index = """
+values = [1, 2, 3, 4, 5]
+exponent = 5
+power_expr = (value * exponent for value in values)
+print(list(power_expr))
+exponent = 10
+"""
+
 
 @pytest.mark.parametrize('code', [
     unsafe_function,
+    unsafe_string,
+    unsafe_change_index,
 ])
 def test_unsafe_gen_expression(
     assert_errors,
@@ -53,6 +86,8 @@ def test_unsafe_gen_expression(
     single_use,
     single_use_variables_assigned,
     no_variables,
+    safe_string,
+    safe_change_index,
 ])
 def test_safe_gen_expression(
     assert_errors,
--- a/tests/test_visitors/test_ast/test_loops/test_comprehensions/test_unsafe_generator_expression.py
+++ b/tests/test_visitors/test_ast/test_loops/test_comprehensions/test_unsafe_generator_expression.py
@@ -78,12 +78,43 @@ print(list(power_expr))
 exponent += 1
 """
 
+unsafe_self = """
+values = [1, 2, 3, 4, 5]
+self.exponent = 5
+power_expr = (value * self.exponent for value in values)
+self.exponent = 1
+print(list(power_expr))
+"""
+
+safe_self = """
+values = [1, 2, 3, 4, 5]
+self.exponent = 5
+power_expr = (value * self.exponent for value in values)
+print(list(power_expr))
+self.exponent = 1
+"""
+
+unsafe_append = """
+values = [1, 2, 3, 4, 5]
+power_expr = (value for value in values)
+values.append(6)
+print(list(power_expr))
+"""
+
+safe_append = """
+values = [1, 2, 3, 4, 5]
+power_expr = (value for value in values)
+print(list(power_expr))
+values.append(6)
+"""
+
 
 @pytest.mark.parametrize('code', [
     unsafe_function,
     unsafe_string,
     unsafe_change_index,
     unsafe_aug_value,
+    unsafe_self,
 ])
 def test_unsafe_gen_expression(
     assert_errors,
@@ -106,6 +137,7 @@ def test_unsafe_gen_expression(
     safe_string,
     safe_change_index,
     safe_aug_value,
+    safe_self,
 ])
 def test_safe_gen_expression(
     assert_errors,
--- a/tests/test_visitors/test_ast/test_loops/test_comprehensions/test_unsafe_generator_expression.py
+++ b/tests/test_visitors/test_ast/test_loops/test_comprehensions/test_unsafe_generator_expression.py
@@ -62,11 +62,28 @@ print(list(power_expr))
 exponent = 10
 """
 
+unsafe_aug_value = """
+values = [1, 2, 3, 4, 5]
+exponent = 5
+power_expr = (value * exponent for value in values)
+exponent += 1
+print(list(power_expr))
+"""
+
+safe_aug_value = """
+values = [1, 2, 3, 4, 5]
+exponent = 5
+power_expr = (value * exponent for value in values)
+print(list(power_expr))
+exponent += 1
+"""
+
 
 @pytest.mark.parametrize('code', [
     unsafe_function,
     unsafe_string,
     unsafe_change_index,
+    unsafe_aug_value,
 ])
 def test_unsafe_gen_expression(
     assert_errors,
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -109,6 +109,7 @@ Semantic versioning in our case means:
 - Forbids inconsistent structuring of multiline comprehensions
 - Forbids to use unpythonic getters and setters such as `get_attribute` or `set_attribute`
 - Now `credits`, `license`, and `copyright` builtins are free to shadow
+- Forbids the reassignment of variables that affect the generator expression
```
sobolevn, line 112: This would be released under `0.16.0`, not `0.15.0`

wemake_python_styleguide/violations/consistency.py
```diff
@@ -2343,3 +2345,37 @@ class AssignToSliceViolation(ASTViolation):
...
+
+    Solution:
+        Either do not reassign new values to the variables that affect the
+        generator expression or implement it before reassigning.
+
+    Example:
+        # Correct
+        a = 1
+        b = (a * i for i in range(5))
+        c = (sum(b))
```
sobolevn, line 2368:
```suggestion
        c = sum(b)
```

wemake_python_styleguide/violations/consistency.py
```diff
@@ -2343,3 +2345,37 @@ class AssignToSliceViolation(ASTViolation):
...
+        Reassigning values to variables that affect a generator expression
+        before the generator expression is used will produce a different
+        result from what it is expected.
+
+    Solution:
+        Either do not reassign new values to the variables that affect the
+        generator expression or implement it before reassigning.
+
+    Example:
+        # Correct
```
sobolevn, line 2365: Can you please use more meaningful names here? We try to write our correct examples as close to the real-world code as possible.

tests/test_visitors/test_ast/test_loops/test_comprehensions/test_unsafe_generator_expression.py
```diff
@@ -0,0 +1,68 @@
...
+    UnsafeGeneratorExpressionViolation,
+)
+from wemake_python_styleguide.visitors.ast.statements import (
+    UnsafeGeneratorExpressionVisitor,
+)
+
+unsafe_function = """
+first_value = 1
+expression = (first_value * index for index in range(5))
+first_value = 2
```
sobolevn, line 13: What about changing `index`? Let's test that as well.

wemake_python_styleguide/violations/consistency.py
```diff
@@ -2343,3 +2345,37 @@ class AssignToSliceViolation(ASTViolation):
...
+        b = (a * i for i in range(5))
+        c = (sum(b))
+
+        # Incorrect
+        a = 1
+        b = (a * i for i in range(5))
+        a = 2
+        c = (sum(c))
+
+    .. versionadded:: 0.15.0
```
sobolevn, line 2376: 0.16.0

wemake_python_styleguide/visitors/ast/statements.py
```diff
@@ -498,3 +508,108 @@ def _check_tuple_arguments_types(
...
+
+
+@final
+class UnsafeGeneratorExpressionVisitor(BaseNodeVisitor):
+    """Checks if a generator expression is safe."""
+
+    def __init__(self, *args, **kwargs) -> None:
+        """Creates the lists that would hold the variables."""
+        super().__init__(*args, **kwargs)
+        self.inside_variables: List[str] = []
```
sobolevn, line 520: We use `_inside_variables` (protected notation) for visitor attributes.

wemake_python_styleguide/visitors/ast/statements.py
```diff
@@ -498,3 +508,108 @@ def _check_tuple_arguments_types(
...
+        self.created_variables: List[str] = []
+        self.affecting_variables: List[str] = []
+
+    def visit(self, node: ast.AST) -> None:
```
sobolevn, line 524: Why don't you use `visit_any_assign` here? It does the same thing.

wemake_python_styleguide/visitors/ast/statements.py
```diff
@@ -498,3 +508,108 @@ def _check_tuple_arguments_types(
...
+        """Checks if the generator expression has affecting variables."""
+        assignments = [
+            sub_node
+            for sub_node in ast.walk(node)
+            if isinstance(sub_node, ast.Assign)
```
sobolevn, line 529: You only check for `Assign`, what about `AugAssign` like `+= 1` or assigning expression like `(x := 1)`? We need to support them as well.

wemake_python_styleguide/visitors/ast/statements.py
```diff
@@ -498,3 +508,108 @@ def _check_tuple_arguments_types(
...
+                created = self.created_variables
+
+                self.affecting_variables = ([
+                    name
+                    for name in inside
+                    if name not in inside or name not in created
+                ])
+
+                if self.affecting_variables:
+                    self._check_gen_usage(node, assignments)
```
sobolevn, line 552: You are missing `self.generic_visit(node)` call at the bottom. It is required.

wemake_python_styleguide/visitors/ast/statements.py
```diff
@@ -498,3 +508,108 @@ def _check_tuple_arguments_types(
...
+
+    def _get_inside_variables(self, node: ast.GeneratorExp) -> None:
+        sub_node = node.elt
+        self.inside_variables = [
```
sobolevn, line 556: We already have this logic as a function: https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/logic/walk.py#L56

wemake_python_styleguide/visitors/ast/statements.py
```diff
@@ -498,3 +508,108 @@ def _check_tuple_arguments_types(
...
+        for generator in assignments:
+            if isinstance(generator.value, ast.GeneratorExp):
+                for target in generator.targets:
+                    if isinstance(target, ast.Name):
+                        gen_target_id_lineno = self._get_id_lineno(target)
+                        gen_positions.append(gen_target_id_lineno)
+
+        return gen_positions
+
+    def _get_id_lineno(self, node: ast.Name):
```
sobolevn, line 589: Missing return type

wemake_python_styleguide/visitors/ast/statements.py
```diff
@@ -498,3 +508,108 @@ def _check_tuple_arguments_types(
...
+        return (node.id, node.lineno)
+
+    def _check_gen_usage(self, node: ast.AST, assignments: List[ast.Assign]):
```
sobolevn, line 592: Missing return type

sobolevn (changes requested):
Great work! Thanks! 💪 

It needs a bit of polishing and I would be happy to merge it.

GibranHL0:
Sure, I'll be working on them today, and I hope later today or tomorrow early commit the requested changes. 🤓

tests/test_visitors/test_ast/test_loops/test_comprehensions/test_unsafe_generator_expression.py
GibranHL0: I believe that the index cannot be changed as the variable is load within the generator expression. Even though, I'll include a test

wemake_python_styleguide/visitors/ast/statements.py
GibranHL0: I didn't consider it 🤔, I'm changing it!

wemake_python_styleguide/violations/consistency.py
```diff
@@ -2363,17 +2363,17 @@ class UnsafeGeneratorExpressionViolation(ASTViolation):
 
     Example:
         # Correct
-        a = 1
-        b = (a * i for i in range(5))
-        c = (sum(b))
+        suffix = 'ish'
```
sobolevn, line 2366: Great one! 👍

wemake_python_styleguide/visitors/ast/statements.py
GibranHL0: Sure, it needs to have AugAssign support, but it'll add too much complexity with the current Visitor.

In the case of the assigning expression, there is a Violation that prevents the usage of `(x := 1)` , WPS332 Found Walrus operator

GibranHL0:
I uploaded a new version of the UnsafeGeneratorExpressionVisitor that fixes the issues with complexity and style. But I'd like to add a new visitor that adds support for the AugAssign. I'll upload it in less than a day! 👍

sobolevn:
> But I'd like to add a new visitor that adds support for the AugAssign

This sounds like not the best idea. 
The better way: move logic to `logic/` and use  a single visitor with all the types you need to call it.

GibranHL0:
I figured out a new way to support `AugAssign` within the visitor and not creating a logic method outside. I also included new tests.(/7-7)/

tests/test_visitors/test_ast/test_loops/test_comprehensions/test_unsafe_generator_expression.py
```diff
@@ -62,11 +62,28 @@
 exponent = 10
 """
 
+unsafe_aug_value = """
+values = [1, 2, 3, 4, 5]
+exponent = 5
+power_expr = (value * exponent for value in values)
+exponent += 1
```
sobolevn, line 69: What about changing `values` here? Like `values.append(1)`
Should we allow it?

sobolevn (commented):
Great progress! 👍

tests/test_visitors/test_ast/test_loops/test_comprehensions/test_unsafe_generator_expression.py
GibranHL0: I think we should not because it affects the expected behavior of the expression. Can I create a new rule request to prevent it? 🤔

sobolevn: I am not sure that it is "expected" because some people might be getting an error here. Let me give you an example:

```python
>>> values = [1, 2, 3]
>>> listcomp = [a * 2 for a in values]
>>> values.append(0)
>>> values, listcomp
([1, 2, 3, 0], [2, 4, 6])
```

Here's how it works with lists: no unexpected problems.
Now, the same with `gen`:

```python
>>> values = [1, 2, 3]
>>> gen = (a * 2 for a in values)
>>> values.append(0)
>>> values, list(gen)
([1, 2, 3, 0], [2, 4, 6, 0])
```

Now, we can see that changing `values` changes how `gen` works. It is not consistent with list comprehensions. If a persons just refactored list comprehension to be a generator expression with a code like this - there would be an error.

And we can catch it!

I recommend to try to visit `ast.Name` with usage / reassining context. It might be easier that our current appoach.

GibranHL0: Yes, I think that we should prevent it!

But, with the current visitor, it would be difficult to implement without making it complex because the `ast.Name` of values have `ctx` `ast.Load` in comparisons, functions or creating other list comprehensions and gens that not affect the `gen`.
```python
>>> values = [1, 2, 3, 4, 5]
>>> gen = (a * 2 for a in values)
>>> if 2 in values:
>>>    addition = sum(values)
>>> values, list(gen)
([1, 2, 3], [2, 4, 6]
```
Then we should check that the `ast.Attribute` append of `gen` is not being used within assignment and usage. That would need at least 2 new methods `visit_Attribute` and `_check_append` but the current visitor is at maximum allowed method capacity.

Also, I believe that a new violation would make the error easier to understand as the current violation only explains the variables that affect a generator expression. The `UnsafeAppendUsageViolation` would point out the forbidden usage of append within assignment and call for lists that affect a generator expression.

sobolevn: Will the current implementation work for:

```python
self.x = 1
gen = (self.x * 2 for a in values)
self.x = 2
...
```

?

> That would need at least 2 new methods visit_Attribute and _check_append but the current visitor is at maximum allowed method capacity.

Don't worry about it for now, let's concentrate on making this work for all general and edge cases first, then we can refactor the working solution into a simplier version.

GibranHL0: Sure then, I'll be working on append support. 📝

On the other hand, the current implementation doesn't work with `self` variables. So, I'll update it as well with new tests 🤠

sobolevn:
Thanks for the PR, but it now has merge conflicts.

PR #2087, opened 2021-07-04
Fix `StaticMethodViolation` `See also` link

monosans:


wemake_python_styleguide/violations/oop.py
```diff
@@ -164,9 +164,6 @@ class StaticMethodViolation(ASTViolation):
     .. versionadded:: 0.1.0
     .. versionchanged:: 0.11.0
 
-    See also:
```
sobolevn, line 167: Can you please change url to `https://bit.ly/3AtnUKE`?
The article has moved.

sobolevn (changes requested):

sobolevn (approved):
Thanks!

PR #2096, opened 2021-07-11
Replace if statement with if expression

monosans:


sobolevn (approved):
Thanks!

PR #2108, opened 2021-07-14
Replace if statement with if expression

monosans:
## Checklist


sobolevn (commented):
Hey! Thanks! 

Are there any other cases worth refactoring? 
If yes, let's combine all of them together.

monosans:
> Are there any other cases worth refactoring?

Hey! I think that's all for now.

sobolevn (approved):
Great thanks!

PR #2111, opened 2021-07-16
Fix WPS510 violation in example

monosans:
- [X] I have double checked that there are no unrelated changes in this pull request


sobolevn (approved):
Great catch! 👀

Issue #2127, opened 2021-07-30
WPS000 on several code fragments


GirZ0n:
### What's wrong

If I run WPS on such code fragments, then the error WPS000 appears, although the code is correct.

**Fragment №1:**
```python
number = int(input())
if number == 1:
    print("1")
else:
    if number == 2: print("2")
```

**Output №1:**
```
Traceback (most recent call last):
  File "/home/ilya/Desktop/venv-wps-15.2/lib/python3.8/site-packages/wemake_python_styleguide/checker.py", line 154, in run
    visitor.run()
  File "/home/ilya/Desktop/venv-wps-15.2/lib/python3.8/site-packages/wemake_python_styleguide/visitors/base.py", line 285, in run
    self.visit(token)
  File "/home/ilya/Desktop/venv-wps-15.2/lib/python3.8/site-packages/wemake_python_styleguide/visitors/base.py", line 279, in visit
    method(token)
  File "/home/ilya/Desktop/venv-wps-15.2/lib/python3.8/site-packages/wemake_python_styleguide/visitors/tokenize/conditions.py", line 56, in visit_name
    self._check_implicit_elif(token)
  File "/home/ilya/Desktop/venv-wps-15.2/lib/python3.8/site-packages/wemake_python_styleguide/visitors/tokenize/conditions.py", line 135, in _check_implicit_elif
    self._check_complex_else(next_tokens, next_token, index)
  File "/home/ilya/Desktop/venv-wps-15.2/lib/python3.8/site-packages/wemake_python_styleguide/visitors/tokenize/conditions.py", line 110, in _check_complex_else
    complex_else = self._if_has_code_below(tokens[index + 1:])
  File "/home/ilya/Desktop/venv-wps-15.2/lib/python3.8/site-packages/wemake_python_styleguide/visitors/tokenize/conditions.py", line 89, in _if_has_code_below
    while remaining_tokens[index - 1].exact_type != tokenize.INDENT:
IndexError: list index out of range

../df_test.py:0:1: WPS000 Internal error happened, see log. Please, take some time to report it
../df_test.py:1:1: D100 Missing docstring in public module
../df_test.py:1:14: WPS421 Found wrong function call: input
../df_test.py:3:5: WPS421 Found wrong function call: print
../df_test.py:3:11: Q000 Remove bad quotes
../df_test.py:5:19: E701 multiple statements on one line (colon)
../df_test.py:5:21: WPS421 Found wrong function call: print
[...]

PR #2132, opened 2021-08-08
Issue 2127 B

danielboghean:
# I have made things!

## Checklist

## Related issues


danielboghean:
I took a stab at fixing the second fragment issue from #2127. Please let me know if there are any comments or advice!

Thank you!

danielboghean:
@sobolevn For this one, there's two issues I can't figure out. The first is that flake8 gives an "B009  Do not call getattr with a constant attribute value, it is not any safer than normal property access." This is something that was done before, and I had just edited it to make it work with my code. Not sure how to proceed with this one.

The second is that running pytest results in all tests passed, but 99.97% coverage. Any thoughts on what I'm doing wrong to not get 100% coverage?

Thank you in advance! I appreciate your patience.

sobolevn:
> B009 Do not call getattr with a constant attribute value, it is not any safer than normal property access.

It means that somewhere in your code you use `getattr(obj, 'prop')`. It is the same as writting `obj.prop` which is the recommended way. 

> The second is that running pytest results in all tests passed, but 99.97% coverage. Any thoughts on what I'm doing wrong to not get 100% coverage?

Whenever I have this issue - I tend to open `htmlcov/index.html` (generated after each `pytest` run) in my browser. It shows nice visual report to understand what is not covered. Then - you can understand how to cover the missing part.

I hope this helps! 👍

danielboghean:
@sobolevn Thank so much, super helpful! `htmlcov/index.html` is great! 

I made the necessary changes and passed all the pytest and flake8 tests locally, so hopefully this is good to go.

Thanks!

sobolevn:
We now need to fix `mypy` 🙂 

```
wemake_python_styleguide/logic/tree/loops.py:54: error: Statement is unreachable
wemake_python_styleguide/logic/tree/loops.py:80: error: Incompatible types in "yield" (actual type "stmt", expected type "Union[For, List[stmt]]")
wemake_python_styleguide/logic/tree/loops.py:81: error: Incompatible types in assignment (expression has type "stmt", variable has type "Union[For, List[stmt]]")
wemake_python_styleguide/visitors/ast/conditions.py:115: error: "If" has no attribute "wps_parent"
Found 4 errors in 2 files (checked 141 source files)
```

You can solve `"If" has no attribute "wps_parent"` by using `get_parent()` function.

danielboghean:
@sobolevn  Sorry for the delay. I went through and hopefully fixed all the mypy issues and reformatted everything to be clean. I ran pytest and flake8 as well, and everything seemed to be working. Hopefully that's the case on here as well!

sobolevn:
@danielboghean great job! Thank you! 👍

danielboghean:
@sobolevn Thanks for all of your patience! I think I finally figured it out and simplified things a bit. Let me know if this works!

wemake_python_styleguide/visitors/ast/conditions.py
```diff
@@ -111,7 +118,12 @@ def _check_simplifiable_returning_if(self, node: ast.If) -> None:
                 if keywords.is_simple_return(else_body):
                     self.add_violation(SimplifiableReturningIfViolation(node))
                 return
-            body = getattr(node, 'wps_parent').body  # TODO: refactor
+
+            parent = getattr(node, 'wps_parent', None)
```
sobolevn, line 122: This is not right for two reasons:
1. It uses `getattr` and not `get_parent`
2. It has `None` as the default value, but later it is used as `parent.body`. It will fail if `parent` is `None`. We don't check for this case

wemake_python_styleguide/visitors/ast/conditions.py
```diff
@@ -102,6 +102,13 @@ def _check_useless_len(self, node: AnyIf) -> None:
             if given_function_called(node.test, {'len'}):
                 self.add_violation(UselessLenCompareViolation(node))
 
+    def _check_else_body(self, node: ast.AST) -> Optional[List[ast.AST]]:
+        else_body = getattr(node, 'orelse', None)
+        else_nodes = (ast.For, ast.While, ast.Try)
```
sobolevn, line 107: Let's move it to a module level as `ELSE_NODES: Final = ...`

wemake_python_styleguide/visitors/ast/conditions.py
```diff
@@ -102,6 +102,13 @@ def _check_useless_len(self, node: AnyIf) -> None:
             if given_function_called(node.test, {'len'}):
                 self.add_violation(UselessLenCompareViolation(node))
 
+    def _check_else_body(self, node: ast.AST) -> Optional[List[ast.AST]]:
```
sobolevn, line 105: What about `ast.AsyncFor` here? We also need a test for it.

wemake_python_styleguide/visitors/ast/conditions.py
```diff
@@ -111,7 +118,12 @@ def _check_simplifiable_returning_if(self, node: ast.If) -> None:
                 if keywords.is_simple_return(else_body):
                     self.add_violation(SimplifiableReturningIfViolation(node))
                 return
-            body = getattr(node, 'wps_parent').body  # TODO: refactor
+
+            parent = getattr(node, 'wps_parent', None)
+            body = parent.body
+            if self._check_else_body(parent):
```
sobolevn, line 124: That's strange. We use `self._check_else_body` as a boolean function, but it is not.
It returns `node.body + else_body | None`. But, we don't reuse it. We do the same thing this function does once again with `body = parent.body + parent.orelse`

sobolevn (changes requested):
Awesome! There are some things to improve in the proposed solution 👍

wemake_python_styleguide/visitors/ast/conditions.py
sobolevn: Consider adding it to `ELSE_NODES`

danielboghean:
@sobolevn Thank you so much for the comments. I've made the changes, but ran into one issue.

I tried to use `get_parent` but the issue is that it returns an `ast.AST` node or None. I then run into an issue where I can't get the body attribute because `AST` doesn't have that attribute. I haven't been able to figure out how to fix this without adding the body type hint to the `AST` class. I apologize if this is an easy solution, I'm just not sure how to deal with it.

The only solution I've found that worked so far is to use `getattr` but then I run into the mypy issue which isn't ideal.

wemake_python_styleguide/visitors/ast/conditions.py
```diff
@@ -1,11 +1,12 @@
 import ast
 from collections import defaultdict
 from functools import reduce
-from typing import ClassVar, DefaultDict, List, Mapping, Optional, Set, Type
+from typing import ClassVar, DefaultDict, Final, List, Mapping, Set, Type
```
sobolevn, line 4: We import `Final` from `typing_extensions`, because we support `python3.6` and `python3.7`

sobolevn (approved):
LGTM! 🎉

sobolevn:
Thanks a lot @danielboghean for your awesome work! 👍

sobolevn:
Refs #2127

PR #2183, opened 2021-09-15
Issue 1974

### LINKED ISSUE #1974: Detect bare `raise` outside of `except` block
# Rule request

## Thesis

Report `raise` statements outside of `except` block

```python
# bad
def smth():
    raise

# good
def smth():
    try:
        ...
    except:
        raise
```

## Reasoning

It causes runtime error:

```
>>> smth()
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in smth
RuntimeError: No active exception to reraise
```

It is possible that the function `smth` is meant to be called only from `except` block and so it won't fail but I'd say it is quite obscure and is a bad practice.

danielboghean:
# I have made things!

## Checklist

- [ X  ] I have double checked that there are no unrelated changes in this pull request (old patches, accidental config files, etc)
- [ X ] I have created at least one test case for the changes I have made
- [ X ] I have updated the documentation for the changes I have made
- [ X ] I have added my changes to the `CHANGELOG.md`

## Related issues
- Fixes #1974

### DIFF
--- a/wemake_python_styleguide/violations/best_practices.py
+++ b/wemake_python_styleguide/violations/best_practices.py
@@ -83,6 +83,7 @@ Summary
    EmptyCommentViolation
    BitwiseAndBooleanMixupViolation
    NewStyledDecoratorViolation
+   BareRaiseViolation
 
 Best practices
 --------------
@@ -154,6 +155,7 @@ Best practices
 .. autoclass:: EmptyCommentViolation
 .. autoclass:: BitwiseAndBooleanMixupViolation
 .. autoclass:: NewStyledDecoratorViolation
+.. autoclass:: BareRaiseViolation
 
 """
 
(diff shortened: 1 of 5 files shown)


danielboghean:
@sobolevn Can you take a look at this when you get a chance. Thanks!

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -2572,3 +2574,35 @@ def my_function(): ...
...
     code = 466
+
+
+@final
+class BareRaiseViolation(ASTViolation):
+    """
+    Forbid using a bare ``raise`` keyword outside of ``except``.
+
+    Reasoning:
+       Bad practice to use a bare ``raise`` outside of an ``except`` block
```
sobolevn, line 2585: It is not just a bad practice, it is a runtime error

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -2572,3 +2574,35 @@ def my_function(): ...
...
+            try:
+                ...
+            except:
+                raise
+
+        # Wrong:
+        def smth():
+            raise
+
+    .. versionadded:: 0.15.0
```
sobolevn, line 2603: 0.16.0

wemake_python_styleguide/visitors/ast/keywords.py
```diff
@@ -65,6 +67,12 @@ def _check_exception_type(self, node: ast.Raise) -> None:
                 BaseExceptionRaiseViolation(node, text=exception_name),
             )
 
+    def _check_bare_raise(self, node: ast.Raise) -> None:
+        parent_except = walk.get_closest_parent(node, ast.ExceptHandler)
```
sobolevn, line 71: We should first check that `node.exc is None`.
Because, this operation is quite heavy, we don't waste it on `raise` with arguments

sobolevn (changes requested):
Thanks a lot! Great work!

wemake_python_styleguide/visitors/ast/keywords.py
danielboghean: Ah yeah, that makes a lot of sense. Thanks!

sobolevn (approved):
👍

PR #2192, opened 2021-09-25
Issue Error in tests fixtures #2191 - Fix rewritten variable

### LINKED ISSUE #2191: Error in tests fixtures
### What's wrong

In test fixtures, one variable is rewritten.

tests/test_visitors/test_ast/test_conditions/test_useless_else/test_useless_loop_else.py:28

```
correct_example4 = """
while ...:
    ...
raise ...
"""

correct_example4 = """
def function():
    for x in ...:
        ...
    else:
        raise ...
"""
```

and also correct examples from 1 to 4(first 4 :) ) do not contain else, so may be we do not need them. But i'm not sure.

### How it should be

correct_example4 should be assigned once

may be correct_example1 - correct_example4(first of them) should be removed. 

not important for issue

### pip information

not important for issue

### OS information

not important for issue

whysage:
# I have made things!

## Checklist

- [ ] I have created at least one test case for the changes I have made - **no need**.
- [ ] I have updated the documentation for the changes I have made  - **no need**.

## Related issues
Closes #2191

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ Semantic versioning in our case means:
 - Fixes that `range(len(x))` was not allowed even outside `for` loops #1883
 - Fixes `UselessReturningElseViolation` to not report `else` with `break` #2187
   (even if we have `except` in loop)
+- Fixes fixture in `UselessReturningElseViolation` #2191
 
 ### Misc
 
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Thanks! 

Btw, you can check other tests as well. This might be a common problem.
You can even write a custom lint rule for it. See https://github.com/wemake-services/wemake-python-styleguide/blob/master/scripts/check_generic_visit.py

PR #2214, opened 2021-10-07
Adds a new rule to forbid to raise an exception from itself

### LINKED ISSUE #2133: Forbid to `raise` exception from itself
# Rule request

We need to forbid code like `raise exc from exc`, because it does not make much sense and works the same way:

```python
>>> e = Exception('m')
>>> raise e
  File "ex.py", line 1, in <module>
    e = Exception("m"); raise e
Exception: m
```

And with `from`:

```python
>>> e = Exception('m')
>>> raise e from e
  File "ex.py", line 1, in <module>
    e = Exception("m"); raise e
Exception: m
```

We need to explicitly test two cases:
1. New `raise`
2. Re-raising inside `except`

thepabloaguilar:
# Adds a new rule to forbid to raise an exception from itself

## Checklist

## Related issues

Closes #2133


thepabloaguilar:
@sobolevn I've made the changes!

PR #2215, opened 2021-10-10
Introduces markdown linter to CI

### LINKED ISSUE #1940: Introduce markdown linter to CI
## Thesis

I propose adding markdown linter to project CI. Initial discussion was [here](https://github.com/wemake-services/wemake-python-styleguide/pull/1927#discussion_r591257409).

One possibility is to add it to Makefile, the second is to add github action, as was suggested [here](https://github.com/wemake-services/wemake-python-styleguide/pull/1927#discussion_r591458412).

Also, there is a question which rules to enable. Linter could check all rules that not warn on current markdown state or enable just some selected rules.

## Reasoning

There is at least one rule of writing project MD files that can be checked by linter, but is checked manually now.
Using linter will reduce toil.

thepabloaguilar:
# Introduces markdown linter to CI

I've chose to use `remark-lint` instead of `markdownlint` because it's extensibility.
I added some rules, almost the sames used by [dry-python/functional-jargon-python](https://github.com/dry-python/functional-jargon-python)!

The rule that originated the issue (lists must use `-`) was added too.

## Checklist

## Related issues

Closes #1940

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1037,7 +1037,7 @@ to the project during `#hactoberfest`. List of awesome people:
 - [@roxe322](https://github.com/roxe322)
 - [@geoc0ld](https://github.com/geoc0ld)
 - [@lensvol](https://github.com/lensvol)
-- [@SheldonNunes ](https://github.com/SheldonNunes)
+- [@SheldonNunes](https://github.com/SheldonNunes)
 - [@tommbee](https://github.com/tommbee)
 - [@valignatev](https://github.com/valignatev)
 - [@vsmaxim](https://github.com/vsmaxim)


thepabloaguilar:
You can check [here](https://github.com/wemake-services/wemake-python-styleguide/pull/2215/checks?check_run_id=3850322935) the failed pipeline because I have use `*` instead of `-` in a unordered list!

thepabloaguilar:
I tried to introduce `remark-lint` with `reviewdog` but its actions has something really annoying [here](https://github.com/reviewdog/action-remark-lint/blob/main/entrypoint.sh#L200), it use just ONE preset and I couldn't subscribe that config even using an configuration file! But the check I've introduced makes its job good

PR #2281, opened 2021-12-02
2268 only disallow certain magic attributes

### LINKED ISSUE #2268: Additional allowed magic attributes; consider 'blocklist' over 'allowlist' approach (WPS609)
I'd like to suggest some exceptions to rule WPS609 (disallow accessing magic attributes/methods):
- `__bases__` (the base classes of a class)
- `__path__` (present in Python packages)
- `__dict__` (present in all non-slots classes)
- `__module__` (present on all Python classes, functions)

These methods are part of established Python standards and are explicitly documented as being readable (and sometimes even settable!). 

I could add these to `_allowed_magic_attributes`, but... I'd like to go one step further and suggest _only_ generating errors for magic methods _which have a possible alternative_. For a few reasons:
1. Currently we force people to `noqa` without offering a workaround (e.g. how do you get a class bases without using `__bases__`?)
2. Some modules define their own (documented) magic attributes (like `enum` defining `__members__`) . This style guide cannot predict what the allowed/disallowed behavior is for all other libraries (current and future).
3. By keeping a 'blocklist' we still prevent all known egregious violations, like calling `__eq__`, `__iter__`, `__delitem__`, etc. directly. We can add more as we encounter them.

Also: If forbidding introspection is what we're after, it would perhaps be better to create a separate violation (so that it can be toggled more easily for introspection-heavy projects)

related:
- #1314 
- #494

ariebovenberg:
# I have made things!

See #2268 

## Checklist

## Related issues

Closes #2268


wemake_python_styleguide/visitors/ast/attributes.py
```diff
@@ -23,14 +23,136 @@ class WrongAttributeVisitor(BaseNodeVisitor):
...
 
-    _allowed_magic_attributes: ClassVar[FrozenSet[str]] = frozenset((
-        '__class__',
-        '__name__',
-        '__qualname__',
-        '__doc__',
-        '__subclasses__',
-        '__mro__',
-        '__version__',
+    _disallowed_magic_attributes: ClassVar[FrozenSet[str]] = frozenset((
```
sobolevn, line 26: Let's move this to `constants.py`, it is rather long!

It would also be a good idea to link to it from the violation's docs.

Thanks a lot for your time and effort 👍

sobolevn (changes requested):

ariebovenberg:
@sobolevn note I added some missing magic methods (like `__next__`) to the already existing `ALL_MAGIC_METHODS` constant. This may affect other violations (in a good way -- it's more complete now). Do we want to mention this as a fix in the changelog as well?

wemake_python_styleguide/constants.py
```diff
@@ -263,6 +266,38 @@
...
+    '__nonzero__',
+    '__unicode__',
+    '__cmp__',
+
+    # copy
+    '__copy__',
+    '__deepcopy__',
+
+    # dataclasses
+    '__post_init__',
```
sobolevn, line 291:
```suggestion
    '__post_init__',

    # attrs:
    '__attrs_pre_init__',
    '__attrs_init__',
    '__attrs_post_init__',
```

sobolevn:
> Do we want to mention this as a fix in the changelog as well?

Yes, please! 👍 

I've also added `attrs` magic methods, because they are special enough to cover 🙂

sobolevn (approved):

ariebovenberg:
Added the changelog entry. I set this PR as a ref since this is where we discussed the fix.

sobolevn:
Thanks again!

PR #2364, opened 2022-02-28
Do not complain about -1 overusage

dreamflasher:
Negative constants are already handled by "Found magic number".

## Checklist

## Related issues

Closes https://github.com/wemake-services/wemake-python-styleguide/issues/2362

### DIFF
--- a/wemake_python_styleguide/logic/complexity/overuses.py
+++ b/wemake_python_styleguide/logic/complexity/overuses.py
@@ -105,8 +105,10 @@ def is_unary_minus(node: ast.AST) -> bool:
     Although negative constants like ``-1``
     should raise violation to force naming them.
     """
-    return (isinstance(node, ast.UnaryOp) and
-            isinstance(node.op, ast.USub) and
-            (not isinstance(node.operand, (Constant, ast.Num)) or
-            node.operand.n == -1)
+    return (
+        isinstance(node, ast.UnaryOp) and
+        isinstance(node.op, ast.USub) and
+        (not isinstance(node.operand, (Constant, ast.Num)) or
+         node.operand.n == -1
+         )
     )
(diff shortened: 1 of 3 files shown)


wemake_python_styleguide/logic/complexity/overuses.py
```diff
@@ -101,13 +101,9 @@ def is_unary_minus(node: ast.AST) -> bool:
 
     We use this predicate to allow values
     like ``-some_value`` to be overused.
-
-    Although negative constants like ``-1``
```
sobolevn, line 105: `-1` is fine. But, things like `-5` and `-37` should be reported.

sobolevn (commented):
Thank you!

wemake_python_styleguide/logic/complexity/overuses.py
dreamflasher: But they are already reported by "Found magic number" – should they be really reported again here?

sobolevn: Yes, complexity rules are different from "Found magic number"

dreamflasher: I am looking at this: https://github.com/wemake-services/wemake-python-styleguide/blob/2377314181bc795e83ad4618a95a188cea86225b/wemake_python_styleguide/constants.py#L395

Should all negatives of those be excluded or only -1?

dreamflasher: @sobolevn what do you think?

sobolevn: Only `-1`

sobolevn (changes requested):
Thank you! Please, add a test case for this.

PR #2411, opened 2022-04-29
Add __init_subclass__ to the methods order

### LINKED ISSUE #2410: Add `__init_subclass__` to the order of methods in WPS338
# Rule request

## Thesis

`__init_subclass__` ([PEP 487](https://peps.python.org/pep-0487/)) should be added to `WPS338` order. The resulting list would look like this:
- `__init_subclass__`
- `__new__`
- `__init__`
- `__call__`
- `__await__`
- public and magic methods
- protected methods
- private methods (we discourage using them)

## Reasoning
In current order I see that construction and initialisation methods are given priority and ordered from instance construction to instance initialisation.

Chronologically, before all that come class construction and its initialisation. First happens in meta class, second can also be done in `__init_subclass__`. Therefore it naturally should come before class `__new__` and `__init__` methods.

## Second thoughts
It might be worth considering doing the same with `__class_getitem__`, since it mostly used for generics construction. But since it might not be even that interesting from user perspective, I'm hesitant at proposing it.

It also should be possible to ignore specific methods by placing #noqa WPS338 at the line of their definition. But that means that reporting will also be done per method instead of current per class reporting, which can be both good (you can see what method is causing the error) and bad (there may be too many errors per class)

Bobronium:
Add `__init_subclass__`  to `WPS338` order.

New allowed methods order:
- `__init_subclass__`
- `__new__`
- `__init__`
- `__call__`
- `__await__`
- 
## Checklist

## Related issues
Closes #2410

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ Semantic versioning in our case means:
   But, in the future we might change the configuration names/logic,
   change the client facing API, change code conventions significantly, etc.
 
+## 0.16.2
+
+### Features
+- Adds `__init_subclass__` in the beginning of accepted methods order as per WPS338
 
 ## 0.16.1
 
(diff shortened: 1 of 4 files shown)


sobolevn (approved):
Thank you!

PR #2414, opened 2022-05-18
Replace typing.re import

srittau:
`typing.re` has been deprecated since Python 3.8 and will be removed in Python 3.12. But `Pattern` can also be imported directly from `typing`.

## Checklist

Should this have a CHANGELOG entry?

## Related issues

python/cpython#92873


sobolevn (approved):
Hi, @srittau! Glad to see you here :)

Thanks for the fix.

PR #2422, opened 2022-06-01
Added `WPS473` rule that disallows usage of `Union` and `Optional`.

### LINKED ISSUE #2421: Add an option to enforce new `Union` typing syntax `|`.
# Rule request

I'd like to propose addition of a new [configuration rule](https://wemake-python-stylegui.de/en/latest/pages/usage/configuration.html) that disallows usage of the `typing.Union` and enforces usage of the new `|` syntax ([PEP604](https://peps.python.org/pep-0604/)).

The new option should be set to `False` by default since Python <3.10 versions do not support this syntax without `from __future__ import annotations` import.

## Thesis

I think that having two different ways to denote union type does not comply with the "There should be one -- and preferably only one -- obvious way to do it" rule.

## Reasoning

The project I've been working on is using `wemake-python-styleguide` and has many contributors - not all of them are accustomed to the new syntax yet, but since we have decided to stick with it we would like to have an option to enforce it. Lacking this option makes reviewing PRs less effective because we have to concentrate on the typing style checking which we tried to minimize by introducing flake.

## Misc.

I'd gladly try my best to prepare PR for that, but I'd like to know if there are any objections.

Waszker:
# I have made things!

I have added a rule that disallows union types using the old `typing.Union` and/or `typing.Optional` notation when using Python 3.10 or newer.
This PR changed the scope of the issue - instead of introducing configuration option, the new rule has been created.

## Checklist

## Related issues

- Closes #2421

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Semantic versioning in our case means:
   But, in the future we might change the configuration names/logic,
   change the client facing API, change code conventions significantly, etc.
 
-## 0.17.2
+## 0.18.0
 
 ### Features
 - Adds `disallow_union_type` configuration option
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,11 @@ Semantic versioning in our case means:
   But, in the future we might change the configuration names/logic,
   change the client facing API, change code conventions significantly, etc.
 
+## 0.17.2
+
+### Features
+- Adds `disallow_union_type` configuration option
+
 ## 0.17.1
 
 ### Features
@@ -24,7 +29,7 @@ Semantic versioning in our case means:
 ## 0.16.2
 
 ### Misc
-- Adds full violation codes to docs and `BaseViolation.full_code` #2409 
+- Adds full violation codes to docs and `BaseViolation.full_code` #2409
 
 ## 0.16.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ Semantic versioning in our case means:
 ## 0.18.0
 
 ### Features
-- Adds `disallow_union_type` configuration option
+- Adds `WPS473` check for `typing.Union` and `typing.Optional` usages
 
 ## 0.17.1
 
--- a/tests/test_checker/test_noqa.py
+++ b/tests/test_checker/test_noqa.py
@@ -378,7 +378,6 @@ def test_noqa_fixture_disabled(
             '--ignore',
             ','.join(IGNORED_VIOLATIONS),
             '--disable-noqa',
-            '--disallow-union-type',
             '--isolated',
             '--select',
             'WPS',
--- a/tests/test_checker/test_noqa.py
+++ b/tests/test_checker/test_noqa.py
@@ -23,6 +23,7 @@ import pytest
 _PY_OLD = sys.version_info < (3, 8)
 _PY38 = (3, 8) <= sys.version_info < (3, 9)
 _PY39 = (3, 9) <= sys.version_info < (3, 10)
+_PYGTE310 = sys.version_info >= (3, 10)
 
 #: Used to find violations' codes in output.
 ERROR_PATTERN = re.compile(r'(WPS\d{3})')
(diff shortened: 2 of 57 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,11 @@ Semantic versioning in our case means:
   But, in the future we might change the configuration names/logic,
   change the client facing API, change code conventions significantly, etc.
 
+## 0.17.2
```
sobolevn, line 19: This is a new major version feature.

tests/test_checker/test_noqa.py
```diff
@@ -378,6 +378,7 @@ def test_noqa_fixture_disabled(
             '--ignore',
             ','.join(IGNORED_VIOLATIONS),
             '--disable-noqa',
+            '--disallow-union-type',
```
sobolevn, line 381: We don't use configuration options for things like that.

You either requires some style, or you don't.
We only allow complexity options.

tests/test_visitors/test_ast/test_annotations/test_union_annotations.py
```diff
@@ -0,0 +1,37 @@
+import pytest
+
+from wemake_python_styleguide.violations.best_practices import (
+    DisallowUnionTypeViolation,
+)
+from wemake_python_styleguide.visitors.ast.annotations import (
+    WrongAnnotationVisitor,
+)
+
+functions_breaking_rule = [
```
sobolevn, line 10: This can be use directly in `@pytest.mark.parametrize`

tests/test_visitors/test_ast/test_annotations/test_union_annotations.py
```diff
@@ -0,0 +1,37 @@
...
+    """def function(a: Union[int, str]): ...""",
+    """def function(a: typing.Union[int, str]): ...""",
+    """def function(a: t.Union[int, str]): ...""",
+    """def function(a: List[Union[int, str]]): ...""",
+    """def function(a: int) -> Union[int, str]: ...""",
+    """def function(a: Optional[int]) -> None: ...""",
+]
+
+
+@pytest.mark.parametrize('function', functions_breaking_rule)
```
sobolevn, line 20: You should test `python3.9-` here explicitly.

wemake_python_styleguide/visitors/ast/annotations.py
```diff
@@ -44,3 +51,42 @@ def _check_arg_annotation(self, node: ast.arg) -> None:
...
+
+    def _has_union_annotation_been_used(
+        self,
+        node: Union[AnyFunctionDef, ast.arg],
+        annotation_node: Optional[ast.expr],
+    ) -> bool:
+        if not isinstance(annotation_node, ast.Subscript):
+            return False
+
+        union_names = ('Union', 'Optional')
```
sobolevn, line 74: This should be a class-level variable.

sobolevn (commented):
Thanks a lot!

wemake_python_styleguide/visitors/ast/annotations.py
Waszker: Done :heavy_check_mark:

tests/test_visitors/test_ast/test_annotations/test_union_annotations.py
Waszker: Done :heavy_check_mark:

CHANGELOG.md
Waszker: Done :heavy_check_mark:

tests/test_visitors/test_ast/test_annotations/test_union_annotations.py
Waszker: Done by checking the Python version and expecting no errors if the version is <3.10

tests/test_checker/test_noqa.py
sobolevn: > My way of ensuring that Union/Optional keywords are not used is by checking the name of the annotation - this will obviously not work if somebody writes the following from typing import Union as UnionType - do you have any suggestion how to solve this problem? Does similar check exist somewhere in the current code so that I could look at it?

We can probably ignore that!

> By turning this check on for the Python 3.10+ versions the project is not passing flake checks if being run on Python 3.10 interpreter - should I update all occurrences of Union and Optional in the code?

Or you can ignore this violation. It is up to you :)

Waszker: > > My way of ensuring that Union/Optional keywords are not used is by checking the name of the annotation - this will obviously not work if somebody writes the following from typing import Union as UnionType - do you have any suggestion how to solve this problem? Does similar check exist somewhere in the current code so that I could look at it?
> 
> We can probably ignore that!

OK thanks :) I know that this solution is far from ideal but I guess it will have to stay for now.

> 
> > By turning this check on for the Python 3.10+ versions the project is not passing flake checks if being run on Python 3.10 interpreter - should I update all occurrences of Union and Optional in the code?
> 
> Or you can ignore this violation. It is up to you :)

I've decided to update the codebase where it was possible - sorry for making the PR so big!

Waszker:
@sobolevn the PR is ready for the re-review. I've tried to minimize the amount of changes but it is still fairly big.
The `make test` passed on my machine (tested using Python 3.10 and Python 3.9).

Waszker:
My bad, I missed the fact that we are testing the library against Python 3.6 too... No point in running my PR since `from __future__ import annotations` will fail.

I have the question then about the intended strategy. Should I:
- revert all the typing changes (revert instances of `|` to `Union` and `Optional`) and add `noqa: WPS473` everywhere
- add some project-wide `noqa: WPS473` so that my PR does not contain so many changes (what about the test_noqa.py though?)
- add conditional `from __future__` import (via if _PYGT36 or similar) and convert all the types containing new syntax to string representation?

Waszker:
My bad, I missed the fact that we are testing the library against Python 3.6 too... No point in running my PR since `from __future__ import annotations` will fail.

I have the question then about the intended strategy. Should I:
- revert all the typing changes (revert instances of `|` to `Union` and `Optional`) and add `noqa: WPS473` everywhere
- add some project-wide `noqa: WPS473` so that my PR does not contain so many changes (what about the test_noqa.py though?)
- add conditional `from __future__` import (via if _PYGT36 or similar) and convert all the types containing new syntax to string representation?

Waszker:
Closing the PR as discussed in https://github.com/wemake-services/wemake-python-styleguide/issues/2426.

Issue #2492, opened 2022-09-12
Docker example for running the linter does not work


dashmug:
### What's wrong

I was trying to test the `wemake-python-styleguide` using the docker command to see how it compares with my own linting setup.

```
$ docker run --rm wemakeservices/wemake-python-styleguide .
```

resulted into:

```
Unable to find image 'wemakeservices/wemake-python-styleguide:latest' locally
latest: Pulling from wemakeservices/wemake-python-styleguide
5843afab3874: Pull complete
2dfaacf7024e: Pull complete
2dbba127c6aa: Pull complete
1a467efa2204: Pull complete
728751b4406e: Pull complete
b6c2e1f1bb4c: Pull complete
820fac86edf7: Pull complete
2f68a5740778: Pull complete
2434db522449: Pull complete
4d31acd1acb8: Pull complete
Digest: sha256:7e3f99893f4dd3f31fc2e1d390435688a48199cfd50a5511b34782f4d40fa3a4
Status: Downloaded newer image for wemakeservices/wemake-python-styleguide:latest
Using reporter:
Linting path:
flake8 --version:
3.9.2 (flake8-bandit: 2.1.2, flake8-broken-line: 0.3.0, flake8-bugbear:
21.4.3, flake8-comprehensions: 3.5.0, flake8-darglint: 1.8.0, flake8-debugger:
4.0.0, flake8-docstrings: 1.6.0, pydocstyle: 6.1.1, flake8-eradicate: 1.1.0,
flake8-string-format: 0.3.0, flake8_commas: 2.0.0, flake8_isort: 4.0.0,
flake8_quotes: 3.2.0, mccabe: 0.6.1, naming: 0.11.1, pycodestyle: 2.7.0,
pyflakes: 2.3.1, rst-docstrings: 0.2.3, wemake_python_styleguide: 0.15.3)
CPython 3.9.6 on Linux
=================================

::set-output name=output::Invalid action reporter specified:
=================================

Invalid action reporter specified:
Process failed with the status code: 1
```

### How it should be

I'd expect it to run the linter on my project.

### Flake8 version and plugins

N/A. I am using the docker image.

### pip information

N/A. I am using the docker image.

### OS information

I am using the docker image.

But if it's relevant I am on Darwin 21.6.0 (MacOS 12.5.1).

sobolevn:
Thanks for the report! 

Does https://github.com/wemake-services/wemake-python-styleguide/pull/2493 solve your problem?

PR #2493, opened 2022-09-12
Fix dockerfile direct invocation

sobolevn:
Closes https://github.com/wemake-services/wemake-python-styleguide/issues/2492


sobolevn:
I verify that it works locally.

PR #2495, opened 2022-09-16
Do not check | in type hints in WPS226

dolfinus:
# I have made things!

In my project I use a lot of new-style annotations - `Type1 | Type2` instead of `Union[Type1, Type2]` provided by Python 3.10. To get these annotations work on previous Python versions (e.g. 3.7) I use `from __future__ import annotations`, which converts all type hints to strings, so it is actually stored in `__annotations__` like `"Type1 | Type2"`.

If module contains a lot of optional annotations, such as:
```python
class MyClass:
  field1: int | None
  field2: str | None
  fiels3: datetime | int | None

...
class MyClass2(MyClass):
  fiels4: SomeType | None
```
and so on, WPS226 starts to fail:
```
WPS226 Found string literal over-use: |
```

In this PR I've added `|` literal to list of ignored strings which should fix the issue.

## Checklist

- [X I have double checked that there are no unrelated changes in this pull request (old patches, accidental config files, etc)

## Related issues

### DIFF
--- a/wemake_python_styleguide/visitors/ast/complexity/overuses.py
+++ b/wemake_python_styleguide/visitors/ast/complexity/overuses.py
@@ -48,6 +48,7 @@ class StringOveruseVisitor(base.BaseNodeVisitor):
         '\n',
         '\r\n',
         '\t',
+        '|',
         b' ',
         b'.',
         b',',
@@ -55,6 +56,7 @@ class StringOveruseVisitor(base.BaseNodeVisitor):
         b'\n',
         b'\r\n',
         b'\t',
+        b'|',
     ))
 
     def __init__(self, *args, **kwargs) -> None:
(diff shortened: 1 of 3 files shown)


wemake_python_styleguide/visitors/ast/complexity/overuses.py
```diff
@@ -48,13 +48,15 @@ class StringOveruseVisitor(base.BaseNodeVisitor):
...
         '\t',
+        '|',
         b' ',
         b'.',
         b',',
         b'',
         b'\n',
         b'\r\n',
         b'\t',
+        b'|',
```
sobolevn, line 59: I don't think we should add this one. Annotations are always strings.

sobolevn, line 59:
```suggestion
```

sobolevn (commented):
Thanks, good idea!

sobolevn (approved):
Thanks!

wemake_python_styleguide/visitors/ast/complexity/overuses.py
dolfinus: Ok

dolfinus:
PR is approved, how about merging?

PR #2525, opened 2022-10-10
docs: warn about darglint perf issues (#2287)

### LINKED ISSUE #2287: Document `darglint`'s performance issues
### What's wrong

While integrating WPS into yet another project, I've hit a case where it caused flake8 to run for over 11 minutes on a relatively performant laptop. Having not seen this before, it was a bit challenging to figure out the cause but I've armed myself with py-spy, and after an hour of staring at the terminal and helpful visualizations, I've discovered that some of the `darglint`'s style parsers are slow, in particular in the corner case I've hit where it took it about 10 minutes to lint a single file with its default `google` style linter. Turned out that I usually set it to `sphinx` style which is why I've never seen this performance disaster in the past.
On my machine with the affected file, `google` was 4360x slower than `sphinx`. And `numpy` was 3420x slower than `sphinx`.

The upstream issue and the adventure details are here: https://github.com/terrencepreilly/darglint/issues/186.

### How it should be

This should be documented and linked to the upstream issue. Alternatively, this linter could be made optional to allow for not using it.
A suggestion could be made to use `sphinx` style if the problem is hit.

Doesn't matter.

### pip information

Doesn't matter.

### OS information

Gentoo Linux, but the performance problem has been observed on the computers of my colleagues running macOS and Fedora. As well as in the CI — pre-commit.ci GitHub App and Zuul (with Fedora/CentOS IIRC).

tales-aparecida:
# Add warning about darglint performance issues with `numpy` and `google` styles

## Checklist

- [n/a] I have created at least one test case for the changes I have made
- [n/a] I have added my changes to the `CHANGELOG.md`

## Related issues

- Closes #2287

I wasn't sure about the desired phrasing here, feel free to make suggestions! I based my suggestions on https://github.com/cjolowicz/cookiecutter-hypermodern-python-instance/pull/822 , where they disabled darglint on commits (pre-commit hooks). 
image

### DIFF
--- a/docs/pages/usage/configuration.rst
+++ b/docs/pages/usage/configuration.rst
@@ -73,6 +73,17 @@ following settings:
 Our `darglint.toml <https://github.com/wemake-services/wemake-python-styleguide/blob/master/styles/darglint.toml>`_
 file is available with the core settings for ``isort``.
 
+.. warning::
+
+  There is a `known issue <https://github.com/terrencepreilly/darglint/issues/186>`_
+  with ``darglint``'s performance when using ``google`` or ``numpy``
+  documentation style, if you face long running times during the linting
+  process you can use the ``sphinx`` style by setting
+  ``docstring-style = sphinx`` in the ``["setup.cfg".flake8]`` section in a
+  nitpick configuration file. Otherwise, you can run ``darglint`` manually and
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thanks!

PR #2529, opened 2022-10-12
Allow calling 'magic' methods inside 'magic' methods.

### LINKED ISSUE #2381: 609: Allow __enter__() inside an __enter__()
### What's wrong

One design pattern is to wrap a context manager. It would be nice to avoid WPS609 errors with this code, which seems to require accessing the direct magic methods.

### How it should be

Allow code like:
```
class Foo:
    ...

    def __enter__(self):
        self._conn.__enter__()
        return self

    def __exit__(self, exc_type, exc_value, traceback):
        self._conn.__exit__(exc_type, exc_value, traceback)
```

I guess the same for aenter/aexit as well.

lensvol:
Resolves #2381.

# I have made things!

## Checklist

## Related issues

### DIFF
--- a/tests/test_visitors/test_ast/test_attributes/test_magic_attributes.py
+++ b/tests/test_visitors/test_ast/test_attributes/test_magic_attributes.py
@@ -1,5 +1,6 @@
 import pytest
 
+from wemake_python_styleguide.constants import ALL_MAGIC_METHODS
 from wemake_python_styleguide.violations.oop import (
     DirectMagicAttributeAccessViolation,
 )
@@ -105,6 +106,35 @@ class Test(object):
     @classmethod
     def method(cls):
         super().{0}()
+"""
+
+magic_wrapper_method = """
+class Test(object):
+
+    def {0}(cls):
+        self.conn.{0}()
+"""
+
+magic_wrapper_method_inside_stacked_cls = """
+class StackedTest(object):
+
+    def something(cls):
+        class Test(object):
+            def {0}(cls):
+                self.conn.{0}()
+
+        anything()
+"""
+
+magic_wrapper_method_with_stacked_wrapper_cls = """
+class TwiceStackedTest(object):
+
+    def {0}(cls):
+        class Test(object):
+            def {0}(cls):
+                self.conn.{0}()
+
+        self.anything.{0}()
 """
 
 
@@ -225,7 +255,7 @@ def test_magic_attribute_correct_contexts(
     magic_super_cls_attribute,
     magic_super_cls_method,
 ])
-def test_other_magic_attributs_allowed(
+def test_other_magic_attributes_allowed(
     assert_errors,
     parse_ast_tree,
     code,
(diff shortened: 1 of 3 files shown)


tests/test_visitors/test_ast/test_attributes/test_magic_attributes.py
```diff
@@ -107,6 +108,35 @@ def method(cls):
...
+
+    def something(cls):
+        class Test(object):
+            def {0}(cls):
+                self.conn.{0}()
+
+        anything()
+"""
+
+magic_wrapper_method_with_stacked_wrapper_cls = """
```
sobolevn, line 129: Is this used?

wemake_python_styleguide/visitors/ast/attributes.py
```diff
@@ -24,6 +30,19 @@ class WrongAttributeVisitor(BaseNodeVisitor):
         'mcs',
     ))
 
+    _func_stack: List[str] = []
```
sobolevn, line 33: Any reason to use a class var?

sobolevn: I think we can get context nodes and see if we are in a magic method that is inside a class.

tests/test_visitors/test_ast/test_attributes/test_magic_attributes.py
lensvol: No, removed it in a separate commit.

wemake_python_styleguide/visitors/ast/attributes.py
lensvol: Interesting, I will try reworking using that.

tests/test_visitors/test_ast/test_attributes/test_magic_attributes.py
```diff
@@ -107,6 +108,24 @@ def method(cls):
         super().{0}()
 """
 
+magic_wrapper_method = """
```
sobolevn, line 111: Should we allow cases like:

```python
def __init__(...):
    other.__init__(...)
```

?

I think it makes sense for `__getattr__` and `__dir__` which can be used on a module level. But, not for anything else. Do you agree?

lensvol:
@sobolevn I've rewritten it using context, thanks for the suggestion!

tests/test_visitors/test_ast/test_attributes/test_magic_attributes.py
lensvol: Do you want to allow only a limited subset of _magic_ methods here? With an explicitly provided list?

I do not see how that can significantly improve the utility of this feature.

wemake_python_styleguide/visitors/ast/attributes.py
```diff
@@ -50,6 +51,15 @@ def _check_protected_attribute(self, node: ast.Attribute) -> None:
 
     def _check_magic_attribute(self, node: ast.Attribute) -> None:
         if access.is_magic(node.attr):
+            # If "magic" method being called has the same name as
+            # the enclosing function, then it is a "wrapper" and thus
+            # a "false positive".
+
+            ctx = nodes.get_context(node)
+            if isinstance(ctx, (ast.FunctionDef, ast.AsyncFunctionDef)):
```
sobolevn, line 59: Use `from wemake_python_styleguide.compat.aliases import FunctionNodes`

sobolevn (commented):
Ok, fair enough! Only one comment left :)

wemake_python_styleguide/visitors/ast/attributes.py
lensvol: See new commit.

sobolevn (approved):
Thanks!

PR #2535, opened 2022-10-20
Use new github actions API to set output

sobolevn:
Old one is deprecated.

Docs https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter


LaurenzReitsam:
Will there be any release/new tag in the next time with these changes included?
`0.17.0` still raises deprecation warning.

PR #2769, opened 2023-11-06
Forbid importing the same object under different aliases

### LINKED ISSUE #2768: Forbid importing the same object with two different names.
# Rule request

## Thesis

Flake8 will forbid this:

```python
from datetime import datetime
from datetime import datetime  # redefinition of unused...
```

WPS will also forbid this:

```python
from datetime import datetime, datetime as datetime # block variables overlap
```

And this:

```python
from datetime import datetime, date as datetime # block variables overlap
```

However, there is no rule that detect this (assuming that both names are used):

```python
from datetime import datetime, datetime as date_time
```

The code above should be forbidden. Either of these is fine:

```python
from datetime import datetime as date_time
# or
from datetime import datetime
# or
from datetime import datetime, date as date_time
```

## Reasoning

Consistency. In a single file, there should be only one name used for the same imported object.

sobolevn:
Closes #2768

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ Semantic versioning in our case means:
 
 - `wemake` output formatter now respects `NO_COLOR=1` option
   to disable text highlighting. See https://no-color.org
+- Add `ImportObjectCollisionViolation` to detect
+  the same objects imported under different aliases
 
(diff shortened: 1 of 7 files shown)


orsinium (approved):
❤️

Issue #3043, opened 2024-08-27
Docs: error in section Formatter.


imtoopunkforyou:
### What's wrong

I noticed an error in the documentation in the [Formatter](https://wemake-python-styleguide.readthedocs.io/en/latest/pages/usage/formatter.html) section (Showing statistic).

### How it should be

Inside the setup.cfg file you need to specify:
```
[flake8]
statistics = True
```
It is necessary with **s** at the end. Otherwise it will not work.

The `--statistic` and `--statistics` options work the same in the terminal, but I think it's better to specify `--statistics` in the documentation.

`--show-statistic` option doesn't exist at all.

### Flake8 version and plugins

{
  "platform": {
    "python_implementation": "CPython",
    "python_version": "3.10.12",
    "system": "Linux"
  },
  "plugins": [
    {
      "plugin": "darglint",
      "version": "1.8.1"
    },
    {
      "plugin": "flake8-bandit",
      "version": "4.1.1"
    },
    {
      "plugin": "flake8-broken-line",
      "version": "1.0.0"
    },
    {
      "plugin": "flake8-bugbear",
      "version": "24.8.19"
    },
    {
      "plugin": "flake8-commas",
      "version": "4.0.0"
    },
    {
      "plugin": "flake8-comprehensions",
      "version": "3.15.0"
    },
    {
      "plugin": "flake8-debugger",
      "version": "4.1.2"
    },
    {
      "plugin": "flake8-docstrings",
      "version": "1.7.0"
    },
    {
      "plugin": "flake8-eradicate",
      "version": "1.5.0"
    },
    {
      "plugin": "flake8-isort",
      "version": "6.1.1"
    },
    {
      "plugin": "flake8-pytest-style",
      "version": "2.0.0"
    },
    {
      "plugin": "flake8-quotes",
      "version": "3.4.0"
    },
    {
      "plugin": "flake8-rst-docstrings",
      "version": "0.3.0"
    },
    {
      "plugin": "flake8-string-format",
      "version": "0.3.0"
    },
    {
      "plugin": "mccabe",
      "version": "0.7.0"
    },
    {
      "plugin": "nitpick",
      "version": "0.35.0"
    },
    {
      "plugin": "pep8-naming",
      "version": "0.14.1"
    },
    {
      "plugin": "pycodestyle",
      "version": "2.12.0"
    },
    {
      "plugin": "pyflakes",
      "version": "3.2.0"
    },
    {
[...]

PR #3044, opened 2024-08-27
Fix(docs): error in section Formatter

imtoopunkforyou:
### Problem
Error in the documentation in the [Formatter](https://wemake-python-styleguide.readthedocs.io/en/latest/pages/usage/formatter.html) section (Showing statistic).

### Related issues
- Refs #3043


sobolevn (approved):
Thanks!

PR #3131, opened 2024-12-12
Drops python3.9 support

sobolevn:


Sxderp:
Why is 3.9 being dropped? EOL is 2025?

sobolevn:
I drop 3.9 on all of my projects due to its typing system. `0.19` will still support 3.9

PR #3212, opened 2024-12-24
Fixes `WPS314` not detecting `match` statements

### LINKED ISSUE #3207: Detect constant conditions in `match`
This `match` condition does not report any violations right now:

```python
match 1:
    case SomeClass(field): ...
```

However, this does report a violation:

```python
if 1: ...
```

```
  1:4      WPS314 Found conditional that always evaluates the same
  if 1: ...
     ^
```

What needs to be done?
- Change docs here: https://github.com/wemake-services/wemake-python-styleguide/blob/e878f3a0cd17c612ad1bc4b442bdcd1c005d20c2/wemake_python_styleguide/violations/consistency.py#L664
- Add new `visit_Match` to https://github.com/wemake-services/wemake-python-styleguide/blob/e878f3a0cd17c612ad1bc4b442bdcd1c005d20c2/wemake_python_styleguide/visitors/ast/compares.py#L169

with something like:

```python
    def visit_Match(self, node: ast.Match) -> None:
        self._check_constant_condition(node.subject)
        self.generic_visit(node)
```
- Add tests to https://github.com/wemake-services/wemake-python-styleguide/blob/master/tests/test_visitors/test_ast/test_compares/test_conditionals.py
- Update `CHANGELOG.md`

MarshalX:
# Fixes `WPS314` not detecting `match` statements

## Checklist

## Related issues

fixes #3207

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,6 +132,7 @@ Semantic versioning in our case means:
 - Fixes several bugs in `WPS322` with multiline strings detection
 - Fixes several violations not been detected in `case:` statements
 - Fixes `WPS434` not detecting reassignments in `:=` case
+- Fixes `WPS314` not detecting `match` statements
 
 ### Misc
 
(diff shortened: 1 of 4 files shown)


sobolevn (approved):
Thanks a lot! Great work!

PR #3311, opened 2025-02-15
Issue 3310 fix

HighBornHoney:
The job is done. This simple code successfully fixes the issue. But, by the way, I'm not sure if a new test is required in this situation. Should I create one?


sobolevn (commented):
Yes, please add a test case.

sobolevn:
closed by https://github.com/wemake-services/wemake-python-styleguide/pull/3335

PR #3325, opened 2025-03-03
Fix "Has a lot of plugins?" info for ruff

ulgens:
ruff has [a very long list of rule sets](https://docs.astral.sh/ruff/rules/) that can be used to extend existing rules. It's not fair to say it doesn't have a lot of plugins. Even [the ruff config in the repo](https://github.com/wemake-services/wemake-python-styleguide/blob/master/pyproject.toml) uses many of them, configure WPS itself as a ruff plugin.

And I don't think this style guide and ruff should be and can be compared with these data points. They are not really alternatives to each other.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ to make all people write **exactly** the same `Python` code.
 | Finds bugs?                |   🤔   |   ✅   |   ❌  |  ✅  |  ✅  |            ✅            |
 | Finds complex code?        |   ❌   |   🤔   |   ❌  |  ❌  |  ✅  |            ✅            |
 | Has a lot of strict rules? |   ❌   |   🤔   |   ❌  |  ❌  |  ✅  |            ✅            |
-| Has a lot of plugins?      |   ✅   |   ❌   |   ❌  |  🤔  |  ❌  |            ✅            |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -90,7 +90,7 @@ to make all people write **exactly** the same `Python` code.
 | Finds bugs?                |   🤔   |   ✅   |   ❌  |  ✅  |  ✅  |            ✅            |
 | Finds complex code?        |   ❌   |   🤔   |   ❌  |  ❌  |  ✅  |            ✅            |
 | Has a lot of strict rules? |   ❌   |   🤔   |   ❌  |  ❌  |  ✅  |            ✅            |
-| Has a lot of plugins?      |   ✅   |   ❌   |   ❌  |  🤔  |  ❌  |            ✅            |
+| Has a lot of plugins?      |   ✅   |   ❌   |   ❌  |  🤔  |  ✅  |            ✅            |
```
sobolevn, line 93: Let's use 🤔 as the status, since it fits better

ulgens: I agree it's better than ❌ but I'm missing why ✅ doesn't fit.

sobolevn: because there are no plugins that you can install, only bundled ones. which is not exactly fit the term "plugin"

Issue #3310, opened 2025-02-13
`_SELF` name should not trigger `WPS117`


sobolevn:
```
server/settings/components/csp.py

  11:1     WPS117 Found name reserved for first argument: _SELF
  _SELF = "'self'"
  ^

```

We also need to check names like `_Self` and `Self`, they should be fine.

ApostolFet:
It seems sufficient to replace the use of the logic.is_wrong_name function with the `in` operator https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/visitors/ast/naming/validation.py#L116

Burka111:
I've encountered a similar issue when working with Django models. Implementing a custom validator function resolved the problem for me. Has anyone else tried this approach? 😄

Whitris:
Is this issue still open? May I try working on it?

ApostolFet:
> Is this issue still open? May I try working on it?

we have a related pr but it seems the author has disappeared. 
https://github.com/wemake-services/wemake-python-styleguide/pull/3311

Whitris:
> > Is this issue still open? May I try working on it?
> 
> we have a related pr but it seems the author has disappeared. 
> https://github.com/wemake-services/wemake-python-styleguide/pull/3311
> 

How do you usually proceed in this situations? Should I wait for him to answer? Can I just open another PR?

sobolevn:
@Whitris please, feel free to open a new PR with the commit from #3311

Whitris:
@sobolevn I don't know if this is the right place to write about this. In case, please direct me to the right place.

I followed the contributing guidelines, but after running `make test`, i get the following error:

```
ERROR: Coverage failure: total of 99 is less than fail-under=100
/home/sa/.cache/pypoetry/virtualenvs/wemake-python-styleguide-IKN65Jdc-py3.10/lib/python3.10/site-packages/pytest_cov/plugin.py:355: CovFailUnderWarning: Coverage failure: total of 99 is less than fail-under=100
  warnings.warn(CovFailUnderWarning(message), stacklevel=1)
                                                                                                                                                     [100%]

---------- coverage: platform linux, python 3.10.12-final-0 ----------
Name                                                                                         Stmts   Miss Branch BrPart  Cover   Missing
----------------------------------------------------------------------------------------------------------------------------------------
[...]

PR #3333, opened 2025-03-06
Fixed bug that caused incomplete code coverage by modifying coverage rule

Whitris:
Now the whole function test_numbers_do_not_error is excluded from code coverage.

## Related issues

- Refs #3310


sobolevn (approved):
Thank you!

PR #3335, opened 2025-03-09
Issue 3310: _SELF name should not trigger WPS117

Whitris:
I modified the naming logic validation so that variations of reserved argument names do not trigger WPS117 for simple variables.

I added a test case that checks all patterns related to simple variables.

## Checklist

## Related issues

- Closes #issue-3310
- Refs #issue-3310

### DIFF
--- a/tests/test_visitors/test_ast/test_naming/test_naming_rules/test_reserved_argument.py
+++ b/tests/test_visitors/test_ast/test_naming/test_naming_rules/test_reserved_argument.py
@@ -5,6 +5,7 @@ import pytest
 from wemake_python_styleguide.constants import SPECIAL_ARGUMENT_NAMES_WHITELIST
 from wemake_python_styleguide.violations.naming import (
     ReservedArgumentNameViolation,
+    TrailingUnderscoreViolation,
 )
 from wemake_python_styleguide.visitors.ast.naming.validation import (
     WrongNameVisitor,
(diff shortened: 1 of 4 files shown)


tests/test_visitors/test_ast/test_naming/test_naming_rules/test_reserved_argument.py
```diff
@@ -75,3 +75,24 @@ def test_correct_argument_name(
     visitor.run()
 
     assert_errors(visitor, [])
+
+
+@pytest.mark.parametrize(
+    'variable_name',
+    ['_SELF', '_Self', 'Self', '_CLS', 'cLs', '_clS', '_MCS', 'mcS', '_mCs'],
```
sobolevn, line 82: Please, test a magic name like `__self__`;

and `cls_` :)

sobolevn (commented):
Thanks! Just one suggestion. Plus, add this change to `CHANGELOG.md` :)

tests/test_visitors/test_ast/test_naming/test_naming_rules/test_reserved_argument.py
Whitris: No problem with the magic name, but I actually have an issue with `cls_` because it raises `WPS120: TrailingUnderscoreViolation`. This rule checks: variables, attributes, functions, methods, and classes.

This is due to the fact that trailing underscores are allowed only when shadowing built-ins or keywords, but `cls` is considered to be neither.

Is this expected or is this another kind of bug?

sobolevn: It is correct, I think. You can ignore this violation with `ignored_types=` (or something like that)

sobolevn (approved):
Thank you!

PR #3399, opened 2025-04-24
Fix int literal

### LINKED ISSUE #3397: WPS432 should ignore literal types
# Rule request

(I am not sure whether this is a bug or a rule request, sorry if I picked the wrong one.)

## Thesis

WPS432 currently is applied to magic numbers within literal types.

````python
foobar: Literal[9001] = 9001
````

In this example, `9001` inside of `Literal` is being reported. This behavior should be changed such that this type annotations are ignored.

## Reasoning

With how the WPS rules currently work, you are allowed to create a `Literal` with a `str` constant, but not with an `int` or `float`.

````python
foobar: Literal[9001] = 9001 # not allowed
barfoo: Literal["something"] = "something" # allowed
````

The only way to have my typed constant is to either use `noqa` or use `int` and `float` as the type, which don't give me the same static guarantees than `Literal`.

More generally, a magic number in the domain of types is different from a magic number in the domain of values. In the case of `Literal` the semantic meaning of the number is given by the way it is used to build the whole type. I think this opens up the question how literals should be treated within type annotations.

### PR DESCRIPTION
**Pull request by LordGvozd**
Fixes #3397

### DIFF
--- a/tests/test_visitors/test_ast/test_builtins/test_numbers/test_magic_numbers.py
+++ b/tests/test_visitors/test_ast/test_builtins/test_numbers/test_magic_numbers.py
@@ -34,6 +34,11 @@ literal_type_hint_with_typing_module = 'code: typing.Literal[{0}] = {0}'
 literal_type_hint_with_typing_exts_module = (
     'code: typing_extensions.Literal[{0}] = {0}'
 )
+literal_type_hint_without_assign = 'code: Literal[{0}]'
+literal_type_hint_typing_without_assign = 'code: typing.Literal[{0}]'
+literal_type_hint_typing_ext_without_assign = (
+    'code: typing_extensions.Literal[{0}]'
+)
 
 
 # Wrong usages:
@@ -80,6 +85,7 @@ foo: Bar[{0}] = {0}
 """
 
 literal_type_with_not_typing_module = 'foo: bar.Literal[{0}] = {0}'
+literal_type_not_typing_module_no_assign = 'foo: bar.Literal[{0}]'
 
 
 @pytest.mark.parametrize(
@@ -99,6 +105,9 @@ literal_type_with_not_typing_module = 'foo: bar.Literal[{0}] = {0}'
         literal_type_hint,
         literal_type_hint_with_typing_module,
         literal_type_hint_with_typing_exts_module,
+        literal_type_hint_without_assign,
+        literal_type_hint_typing_without_assign,
+        literal_type_hint_typing_ext_without_assign,
     ],
 )
 @pytest.mark.parametrize(
@@ -214,6 +223,7 @@ def test_magic_number_whitelist(
         dict_key,
         not_literal_type_hint,
         literal_type_with_not_typing_module,
+        literal_type_not_typing_module_no_assign,
     ],
 )
 @pytest.mark.parametrize(
@@ -266,6 +276,7 @@ def test_magic_number_warning(
         dict_key,
         not_literal_type_hint,
         literal_type_with_not_typing_module,
+        literal_type_not_typing_module_no_assign,
     ],
 )
 @pytest.mark.parametrize(
--- a/tests/test_visitors/test_ast/test_builtins/test_numbers/test_magic_numbers.py
+++ b/tests/test_visitors/test_ast/test_builtins/test_numbers/test_magic_numbers.py
@@ -29,6 +29,8 @@ dict_definition_value = '{{"first": {0}}}'
 set_definition = '{{"first", {0}, "other"}}'
 tuple_definition = '({0}, )'
 
+literal_type_hint = 'code: Literal[{0}] = {0}'
+
 # Wrong usages:
 
 assignment_binop = 'final = {0} + 1'
@@ -68,6 +70,10 @@ some_dict = {{11: 12, 13: 14}}
 some_dict[{0}]
 """
 
+not_literal_type_hint = """
+foo: Bar[{0}] = {0}
+"""
+
 
 @pytest.mark.parametrize(
     'code',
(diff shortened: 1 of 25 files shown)

### CODE REVIEW

File: wemake_python_styleguide/visitors/ast/builtins.py [hunk 1]
```diff
@@ -189,15 +189,15 @@ def visit_Num(self, node: ast.Constant) -> None:
...
-            return
 
-        if isinstance(node.value, int) and node.value <= self._non_magic_modulo:
+        if any((
+            isinstance(parent, self._allowed_parents),
+            node.value in constants.MAGIC_NUMBERS_WHITELIST,
+            isinstance(node.value, int)
+            and node.value <= self._non_magic_modulo,
+            self._check_is_number_in_typing_literal(parent),
+        )):
```
sobolevn, line 199:
```suggestion
        if (
            isinstance(parent, self._allowed_parents)
            or node.value in constants.MAGIC_NUMBERS_WHITELIST
            or (isinstance(node.value, int)
            and node.value <= self._non_magic_modulo)
            or self._check_is_number_in_typing_literal(parent)
        ):
```

tests/test_visitors/test_ast/test_builtins/test_numbers/test_magic_numbers.py
```diff
@@ -29,6 +29,8 @@ def function_name(param1, param2: int = {0}):
 set_definition = '{{"first", {0}, "other"}}'
 tuple_definition = '({0}, )'
 
+literal_type_hint = 'code: Literal[{0}] = {0}'
```
sobolevn, line 32: Please, also test `typing.Literal` and `typing_extensions.Literal`

sobolevn (commented):
Thank you! Please, also add a changelog entry

tests/test_visitors/test_ast/test_builtins/test_numbers/test_magic_numbers.py
LordGvozd: Should we check `import as`? For example:

```python
import bar as typing

foo = typing.Literal[418] = 418  # Wrong?

```

sobolevn: nope, we should not. it is overly complex to do and very rare in practice.

LordGvozd: Ok, thanks

wemake_python_styleguide/visitors/ast/builtins.py
LordGvozd: Not on the topic of PR, but don't we need an analogue of WPS222, but for `any()` and `all()`? Because wouldn't `if+any()` with multiple conditions be as hard to read as a regular `if` condition with `or`?
```python
if (
    smth
    and boo
    and ctx
    and va
    and (1 - 1 or (not 2 + 2 == 3))
): # WPS222
    ...
if any((
    smth, 
    boo,
    ctx,
    va,
    (1 - 1 or (not 2 + 2 == 3))
)): # No violations
    ...
```

sobolevn: Good idea, lets create a new issue for it! 👍

LordGvozd: Great, I'll close PR and work on that)

tests/test_visitors/test_ast/test_builtins/test_numbers/test_magic_numbers.py
```diff
@@ -29,6 +29,13 @@ def function_name(param1, param2: int = {0}):
...
+literal_type_hint_with_typing_module = 'code: typing.Literal[{0}] = {0}'
+literal_type_hint_with_typing_exts_module = (
```
sobolevn, line 34: Please, also test `code: Literal[{0}]`

wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -212,6 +215,19 @@ def _check_is_magic(self, node: ast.Constant) -> None:
             best_practices.MagicNumberViolation(node, text=real_value),
         )
 
+    def _check_is_number_in_typing_literal(self, node: ast.AST | None) -> bool:
```
sobolevn, line 218: please, move this to `logic/`, example https://github.com/wemake-services/wemake-python-styleguide/blob/master/wemake_python_styleguide/logic/naming/enums.py

sobolevn (commented):
👍

wemake_python_styleguide/visitors/ast/builtins.py
LordGvozd: Should I put this in `logic/complexity/annotations.py` or create a new one in `logic/complexity/builtins.py`?

sobolevn: `logic/complexity/annotations.py` seems fine 👍

wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -198,7 +201,9 @@ def _check_is_magic(self, node: ast.Constant) -> None:
             isinstance(parent, self._allowed_parents)
             or node.value in constants.MAGIC_NUMBERS_WHITELIST
             or is_non_magic
-            or self._check_is_number_in_typing_literal(parent)
+            or check_is_node_in_specific_annotation(
+                parent, 'Literal', {'typing', 'typing_extensions'}
```
sobolevn, line 205: Please, move `{'typing', 'typing_extensions'}` to class level and make it a `frozenset`

sobolevn (approved):

sobolevn:
Thank you!

Issue #3420, opened 2025-05-16
Issue with the SHOULD_BE_RAISED from test_noqa.py


MIXgleb:
### What's wrong

I checked each WPS rule from the SHOULD_BE_RAISED dict from the test_noqa.py and found a discrepancy in the number for some of them, although the total number converges.

Rule WPS432: found 1, expected 2
Rule WPS517: found 1, expected 2
Rule WPS604: found 2, expected 1
Rule WPS614: found 2, expected 1

I analyzed the noqa.py file using `flake8 . --select=WPS` and did not find any rules other than listed in the IGNORED_VIOLATIONS tuple.
I also found that one of the two mentioned in noqa.py rules 604 and 614 are useless. And when the file was analyzed again, after deleting them, the result did not change and we still see only the IGNORED_VIOLATIONS rules.

Therefore, I think that it is necessary to correct the number of these rules and remove unnecessary rules from the noqa.py so that they show the correct essence.

### How it should be

1) SHOULD_BE_RAISED = types.MappingProxyType(
    {
        - 'WPS432': 2,
        - 'WPS517': 2,
        'WPS604': 1,
        'WPS614': 1,
        + 'WPS432': 1,
        + 'WPS517': 1,
    },
)

2) remove unnecessary useless rules WPS604, WPS614 from the noqa.py

### Flake8 version and plugins

no matter

### pip information

no matter

### OS information

no matter

PR #3419, opened 2025-05-16
Fixes #3417 #3420

MIXgleb:
# I have made things!

## Checklist

- [x] I have double checked that there are no unrelated changes in this pull request (old patches, accidental config files, etc)
- [x] I have created at least one test case for the changes I have made
- [x] I have updated the documentation for the changes I have made
- [x] I have added my changes to the `CHANGELOG.md`

## Related issues

Closes #3417 #3420 

Hi,
- I've removed the unnecessary WPS604 and WPS614 rules from the noqa.py
- I've added variation to the rule `WPS466`.
Now, for `python312+` it is allowed to use [] in decorators due to the [PEP 695](https://peps.python.org/pep-0695/ "https://peps.python.org/pep-0695/")

PR #3423, opened 2025-05-19
Fixes #3417 , #3420

### LINKED ISSUE #3417: WPS466 NewStyledDecoratorViolation issue
### What's wrong

I agree that decorators should use only simple grammar.

But the current check overlaps with the new Generic designation from version 3.12 like
DecoratorClass[T, **P]

This is not a getitem call, this is a new notation, which in my opinion is a good approach.

for example, in a real project, I use such a construction for a specific decorator. It requires to specify the type of generic in order for the typechecker to work correctly.

@decorator[str | None, ...](var1=1, var2=2)
def foo() -> str | None: ...

### How it should be

raise an error for complex grammar (@some.decorator['method'] + other), but skip the notation for generics (@decorator[str | int](var1=1)).

no matter

### pip information

no matter

### OS information

no matter

MIXgleb:
The clone of the PR #3419 with clean commits and fixes from the maintainers.

## Checklist

## Related issues

- Closes #3417 , #3420 

Hi,

- I've removed the unnecessary WPS604 and WPS614 rules from the noqa.py, #3420 
- I've added variation to the rule WPS466:
Now, for `python312+` it is allowed to use `[]` in the decorator due to the [PEP-695](https://peps.python.org/pep-0695/), #3417

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Semantic versioning in our case means:
 
 ### Bugfixes
 
+- Removes unnecessary WPS604 and WPS614 rules from the noqa.py, #3420
 - Fixes `WPS115` false-positive on `StrEnum`, `IntEnum`, `IntFlag` attributes, #3381
 - Fixes `WPS432`, now it ignores magic numbers in `Literal`, #3397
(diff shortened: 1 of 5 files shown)


MIXgleb:
Added tests for full coverage

PR #3487, opened 2025-07-12
Issue 3458

### LINKED ISSUE #3458: Forbid "fat" finally
# Rule request

Allow `finally` block to have simple content only.
It can contain few lines of code, but it shouldn't contain untrivial logic, which can not be understand very quickly at code reading.

## Thesis

Correct `finally` block:
1) One-two statements.
2) One `if` with `elif` and `else` w/o sub-blocks.

Wrong `finally` block:
One bloated with business logic and hard to read.

It's arguable when exactly the watershed should be located.
I propose something like this for correct `finally`:
1) Up to 10 statements.
2) Up to 2 levels of indentation (i.e. sub-blocks)
3) _Maybe_ up to 10 basic tokens in each statement.

## Reasoning

IMO, `finally` block best suites for code snippets like this:

```python

try:
    x = open_something()
    y = x.read_value()
    ...
except SomeError as e:
   ...
finally:
    close_something(x)
```

It's easy to read and understand what `finally` block stands for.
But code with big and difficult `finally` block smells.

My arguments are here:
1. Code in `finally` block should basically do nothing other than some "cleaning" or "finishing" after `try` and provide only necessary things. If some code can be located after `try-except-finally`, it should be located there, and not in `finally` itself.
2. There is an accepted PEP 765 (https://peps.python.org/pep-0765/) which disallow `return/break/continue` in `finally` block for 3.14+. Thus `finally` is already block with very special rules, we just want to make those rules to be more strict.
3. Generated bytecode has two copies of `finally` block content on latest CPython versions. As a result, big `finally` block provide larger code objects.

NikitaSemenovAiforia:
# I have made things!

## Checklist

## Related issues

- Closes #3458

## Comment

The previous solution used a number of lines to determine finally block complexity, which works purely with a function with multiple arguments or a function with a long name. New logic uses the same logic as try body violation: at most one statement.

## Docs
I also changed `poetry shell` command in docs cause it was [change](https://python-poetry.org/blog/announcing-poetry-2.0.0/#poetry-export-and-poetry-shell-only-available-via-plugins) since 2.0.0 version


CHANGELOG.md
```diff
@@ -23,6 +23,12 @@ Semantic versioning in our case means:
 ### Features
 - Adds more names to `WPS110`: `spam`, `ham`, `tmp`, `temp`, `arr`
 
+## 1.2.1
+
+### Bugfixes
+
+- Fixes `WPS243` to use number of statements in `finally` body instead of a number of lines, #3458
+- Rename `--max-lines-in-finally` option to `--max-finally-body-length`, #3458
```
sobolevn: This is a breaking change :(
Sadly, we can't do that. Please, revert.

wemake_python_styleguide/violations/complexity.py
```diff
@@ -1402,22 +1402,25 @@ class TooManyMatchCaseViolation(ASTViolation):
...
     Solution:
-        Simplify the ``finally`` block. Use context managers, use ``ExitStack``.
+        Move things out of the ``finally`` block or create new functions.
+        The fewer lines you have in your ``finally`` block -
+        the safer you are from accidental errors.
+        Use context managers, use ``ExitStack``.
 
     Configuration:
-        This rule is configurable with ``--max-lines-in-finally``.
+        This rule is configurable with ``--max-finally-body-length``.
```
sobolevn: We can solve this problem with better docs. We can explain what "line" means here. It basically means "statement"

sobolevn (commented):
Thanks!

CHANGELOG.md
NikitaSemenovAiforia: May I keep the old name for the option but use the new one inside code? The new name follows the same pattern as try violation. Or rename internal variable as well?

NikitaSemenovAiforia: I also had to set the default value = 2 to support backward compatibility. It is slightly looser strictness than for the try body.

wemake_python_styleguide/violations/complexity.py
NikitaSemenovAiforia: done

CHANGELOG.md
NikitaSemenovAiforia: Fixed

NikitaSemenovAiforia: fixed one more time =) now should be fine.

sobolevn (approved):
Thank you for the quick fix! I will release this tomorrow 🤝

PR #3502, opened 2025-08-03
Allow set comps

### LINKED ISSUE #3500: Allow set comprehensions in `WPS335`
We don't allow `list` and `set` to be used as iterable parts of `for`, because we only allow `tuple`s there.

We need to allow `SetComp` in `for` iterable: 

```
  118:22   WPS335 Found incorrect `for` loop iter type
  for queue in {track.id for track in expected}:
               ^
```

Because it is a completely different logic. I can't replace it with a tuple.

markparonyan:
- [x] I have added my changes to the CHANGELOG.md

closes #3500

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Semantic versioning in our case means:
 
 - Allows `__init__.py` files that consist only of imports, #3486
 - Adds `--max-conditions` option, #3493
+- Allow set comprehensions in WPS335, #3500
 
 ### Misc
 


sobolevn (commented):
Looks like I am going a bit insane :)

You did exceletly well, but why did I just create a new variable for this case? It would solve my problem and not produce any violations. And will be very consistent.

I am very sorry that you had to fix a task that I created for no reason 🙏 

Please, tell me to buy you a beer whenever you see me ❤️

PR #3526, opened 2025-09-12
Issue 3520

### LINKED ISSUE #3520: Prefer `if` over `match`
# Rule request

## Thesis

We need to force users to prefer `if` over simple `match` statements.

Example:

```python
match state:
    case EventType.REJECT:
        user = 'rejected'
    case _:
        user = 'active'
```

Can be rewritten as:

```python
if state == EventType.REJECT:
   user = 'rejected'
else:
   user = 'active'
```

## Reasoning

Simple `if` is better, less nested, and more performant then `match`.

When can we raise this violation?
1. When there are just two cases in `match`
2. When the first `case` is simple, matching a single value or several values `EventType.REJECT | EventType.REJECT2 | EventType.REJECT3`
3. When the second `case` is `_` (so it can be converted to `else` easily`)

ZhdanovAM72:
# I have made things!
Adds a new consistency violation to enforce using `if` instead of simple two-case `match` statements with a wildcard `_`.

> - Implemented `SimplifiableMatchViolation` with code `WPS365`
> - Created `SimplifiableMatchVisitor` to detect:
>   - Exactly two cases
>   - Second case is `case _:`
>   - First case uses a simple pattern (e.g. `CONST`, `ENUM.VALUE`, `1`, or `A | B`)

## Checklist

## Related issues

Closes #3520

### DIFF
--- a/tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
+++ b/tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
@@ -10,7 +10,7 @@ from wemake_python_styleguide.visitors.ast.conditions import (
 # Wrong:
 simplifiable_match_match = """
 match subject:
-    case {0}:
+    case {0}{1}:
         pass
     case _:
         pass
@@ -18,19 +18,11 @@ match subject:
 
 simplifiable_union_match_match = """
 match subject:
-    case {0} | {1}:
+    case {0} | {1}{2}:
         pass
     case _:
         pass
 """
-
-simplifiable_with_const_as_binding_match = """
-    match subject:
-        case State.REJECTED as status:
-            pass
-        case _:
-            pass
-    """
 
 # Correct:
 complex_match = """
@@ -93,61 +85,63 @@ match subject:
 
 
 @pytest.mark.parametrize(
-    'code',
+    ('code', 'as_binding'),
     [
-        '1',
-        'True',
-        'None',
-        '"string"',
-        'ns.CONST',
-        'State.REJECTED',
-        'x if x > 0',
+        ('1', ''),
+        ('True', ''),
+        ('None', ''),
+        ('"string"', ''),
+        ('ns.CONST', ''),
+        ('State.REJECTED', ''),
+        ('x if x > 0', ''),
+        ('2', ' as x'),
+        ('False', ' as x'),
+        ('"string"', ' as x'),
+        ('ns.CONST', ' as x'),
+        ('State.REJECTED', ' as x'),
     ],
 )
 def test_simplifiable_single_match(
     code: str,
+    as_binding: str,
     assert_errors,
     parse_ast_tree,
     default_options,
 ):
     """Test that simple single-case match raises a violation."""
-    tree = parse_ast_tree(simplifiable_match_match.format(code))
+    tree = parse_ast_tree(simplifiable_match_match.format(code, as_binding))
     visitor = SimplifiableMatchVisitor(default_options, tree=tree)
     visitor.run()
     assert_errors(visitor, [SimplifiableMatchViolation])
 
 
 @pytest.mark.parametrize(
-    ('left', 'right'),
+    ('left', 'right', 'as_binding'),
     [
-        ('1', '2'),
-        ('True', 'False'),
-        ('"a"', '"b"'),
-        ('State.OK', 'State.ERROR'),
-        ('"first" | "second"', '"third"'),
+        ('1', '2', ''),
+        ('True', 'False', ''),
+        ('"a"', '"b"', ''),
+        ('State.OK', 'State.ERROR', ''),
+        ('"first" | "second"', '"third"', ''),
+        ('3', '4', ' as x'),
+        ('False', 'True', ' as x'),
+        ('"c"', '"d"', ' as x'),
+        ('State.BAD', 'State.DEBUG', ' as x'),
+        ('"first" | "second"', '"third"', ' as x'),
     ],
 )
 def test_simplifiable_union_match(
     left: str,
     right: str,
+    as_binding: str,
     assert_errors,
     parse_ast_tree,
     default_options,
 ):
     """Test that union pattern raises violation."""
-    tree = parse_ast_tree(simplifiable_union_match_match.format(left, right))
-    visitor = SimplifiableMatchVisitor(default_options, tree=tree)
-    visitor.run()
-    assert_errors(visitor, [SimplifiableMatchViolation])
-
-
-def test_simplifiable_with_const_as_binding(
-    assert_errors,
-    parse_ast_tree,
-    default_options,
-):
-    """Test that `case CONST as name:` is still simplifiable."""
-    tree = parse_ast_tree(simplifiable_with_const_as_binding_match)
+    tree = parse_ast_tree(
+        simplifiable_union_match_match.format(left, right, as_binding)
+    )
     visitor = SimplifiableMatchVisitor(default_options, tree=tree)
     visitor.run()
     assert_errors(visitor, [SimplifiableMatchViolation])
--- a/tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
+++ b/tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
@@ -7,6 +7,7 @@ from wemake_python_styleguide.visitors.ast.conditions import (
     SimplifiableMatchVisitor,
 )
 
+# Wrong:
 simplifiable_match_template = """
 match subject:
     case {0}:
@@ -23,6 +24,15 @@ match subject:
         pass
 """
 
+simplifiable_with_const_as_binding_template = """
+    match subject:
+        case State.REJECTED as status:
+            pass
+        case _:
+            pass
+    """
+
+# Correct:
 complex_match = """
 match subject:
     case State.FIRST:
@@ -41,33 +51,41 @@ match subject:
         pass
 """
 
-sequence_match = """
-match subject:
-    case [1, 2]:
-        pass
-    case _:
-        pass
-"""
-
-mapping_match = """
-match subject:
-    case {"status": "ok"}:
-        pass
-    case _:
-        pass
-"""
-
-class_match = """
-match subject:
-    case SomeClass():
-        pass
-    case _:
-        pass
-"""
-
 guard_match = """
 match subject:
     case x if x > 0:
+        pass
+    case _:
+        pass
+"""
+
+class_with_args_match = """
+match subject:
+    case SomeClass(x):
+        pass
+    case _:
+        pass
+"""
+
+class_with_kwargs_match = """
+match subject:
+    case SomeClass(name=x):
+        pass
+    case _:
+        pass
+"""
+
+sequences_match = """
+match subject:
+    case [x, *rest]:
+        pass
+    case _:
+        pass
+"""
+
+mappings_match = """
+match subject:
+    case {"key": x}:
         pass
     case _:
         pass
@@ -83,6 +101,8 @@ match subject:
         '"string"',
         'ns.CONST',
         'State.REJECTED',
+        '[1, 2]',
+        '{"status": "ok"}',
     ],
 )
 def test_simplifiable_single_match(
@@ -128,14 +148,7 @@ def test_simplifiable_with_const_as_binding(
     default_options,
 ):
     """Test that `case CONST as name:` is still simplifiable."""
-    code = """
-    match subject:
-        case State.REJECTED as status:
-            pass
-        case _:
-            pass
-    """
-    tree = parse_ast_tree(code)
+    tree = parse_ast_tree(simplifiable_with_const_as_binding_template)
     visitor = SimplifiableMatchVisitor(default_options, tree=tree)
     visitor.run()
     assert_errors(visitor, [SimplifiableMatchViolation])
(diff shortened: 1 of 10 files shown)


tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
```diff
@@ -0,0 +1,165 @@
+import pytest
+
+from wemake_python_styleguide.violations.consistency import (
+    SimplifiableMatchViolation,
+)
+from wemake_python_styleguide.visitors.ast.conditions import (
+    SimplifiableMatchVisitor,
+)
+
+simplifiable_match_template = """
```
sobolevn: Please, separate `# Correct: ` and `# Wrong:` examples, so it would be easy to understand which are which.

tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def test_simplifiable_with_const_as_binding(
+    assert_errors,
+    parse_ast_tree,
+    default_options,
+):
+    """Test that `case CONST as name:` is still simplifiable."""
+    code = """
+    match subject:
```
sobolevn: Please, move this example to the top level.

tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
```diff
@@ -0,0 +1,165 @@
...
+    visitor.run()
+    assert_errors(visitor, [SimplifiableMatchViolation])
+
+
+@pytest.mark.parametrize(
+    'template',
+    [
+        complex_match,
+        no_wildcard_match,
+        sequence_match,
```
sobolevn: I guess we can also simply `[]` and `{}` and classes? What do you think?

wemake_python_styleguide/violations/consistency.py
```diff
@@ -2415,3 +2415,46 @@ class NotInWithUnaryOpViolation(ASTViolation):
...
+            user = 'active'
+
+        # Wrong:
+        match state:
+            case EventType.REJECT:
+                user = 'rejected'
+            case _:
+                user = 'active'
+
+    .. versionadded:: 1.5.0
```
sobolevn, line 2455:
```suggestion
    .. versionadded:: 1.5.0

```

wemake_python_styleguide/violations/consistency.py
```diff
@@ -2415,3 +2415,46 @@ class NotInWithUnaryOpViolation(ASTViolation):
...
+    """
+
+    error_template = (
+        'Found simple ``match ... case _`` constructs, use ``if ... else``'
```
sobolevn:
```suggestion
        'Found simplifiable `match` statement that can be just `if`'
```

tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
ZhdanovAM72: I added checks for simple [] and {} patterns, but for classes

I think:
```
match x:
    case SomeClass():
    ...
```
works like:
`if isinstance(x, SomeClass):`

 not like:
`if x == SomeClass():`

tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
```diff
@@ -0,0 +1,179 @@
...
+
+simplifiable_union_match_template = """
+match subject:
+    case {0} | {1}:
+        pass
+    case _:
+        pass
+"""
+
+simplifiable_with_const_as_binding_template = """
```
sobolevn: this is not a template.

tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
sobolevn: Ok, let's start small. We can add some patterns later. Please, open a new issue about this, so we won't forget about it.

wemake_python_styleguide/violations/consistency.py
```diff
@@ -2415,3 +2415,47 @@ class NotInWithUnaryOpViolation(ASTViolation):
 
     error_template = 'Found legacy `not ... in`, use `... not in` instead'
     code = 364
+
+
+@final
+class SimplifiableMatchViolation(ASTViolation):
+    """
+    Match statement can be simplified to `if`.
```
sobolevn:
```suggestion
    Some ``match`` statements can be simplified to ``if`` statements.
```

wemake_python_styleguide/violations/consistency.py
```diff
@@ -2415,3 +2415,47 @@ class NotInWithUnaryOpViolation(ASTViolation):
...
+
+    Reasoning:
+        Using ``match`` for simple two-case conditions
+        (including wildcard ``_``)
+        is unnecessarily verbose and less performant than a plain ``if/else``.
+        Simple conditions should prefer readability and simplicity.
+
+    Solution:
+        Replace simple ``match ... case _`` constructs with ``if ... else``.
```
sobolevn:
```suggestion
        Replace violating ``match ... case _`` statements with ``if ... else``.
```

wemake_python_styleguide/violations/consistency.py
```diff
@@ -2415,3 +2415,47 @@ class NotInWithUnaryOpViolation(ASTViolation):
...
+
+    When is this violation not issued?
```
sobolevn: It is better to describe when it is raised. Not when it is not raised.

wemake_python_styleguide/visitors/ast/conditions.py
```diff
@@ -188,3 +189,82 @@ def visit_Compare(self, node: ast.Compare) -> None:
...
         self.generic_visit(node)
+
+
+@final
+class SimplifiableMatchVisitor(BaseNodeVisitor):
+    """Checks for match statements that can be simplified to if/else."""
+
+    def visit_Match(self, node: ast.Match) -> None:
+        """Checks match statements."""
+        cases = node.cases
```
sobolevn, line 205: Please, move this logic to `_check_simplifiable_match` method and only keep top level logic in `visit_` method.

wemake_python_styleguide/visitors/ast/conditions.py
```diff
@@ -188,3 +189,82 @@ def visit_Compare(self, node: ast.Compare) -> None:
...
+
+    def _is_simple_sequence(self, pattern: ast.MatchSequence) -> bool:
+        """Check [1, 2] but not [x, *rest]."""
+        return all(
+            self._is_simple_pattern(sub) for sub in pattern.patterns
+        ) and not any(
+            isinstance(sub, ast.MatchStar) for sub in pattern.patterns
+        )
+
+    def _is_simple_mapping(self, pattern: ast.MatchMapping) -> bool:
```
sobolevn: You can move this logic to `logic/` :)

tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
ZhdanovAM72: #3527

tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
```diff
@@ -0,0 +1,177 @@
...
+
+no_wildcard_match = """
+match subject:
+    case State.FIRST:
+        pass
+    case State.SECOND:
+        pass
+"""
+
+guard_match = """
```
sobolevn, line 54: This must be simplifiable. This is literally `if x > 0:`

wemake_python_styleguide/violations/consistency.py
```diff
@@ -2415,3 +2415,49 @@ class NotInWithUnaryOpViolation(ASTViolation):
...
+    Reasoning:
+        Using ``match`` for simple two-case conditions
+        (including wildcard ``_``)
+        is unnecessarily verbose and less performant than a plain ``if/else``.
+        Simple conditions should prefer readability and simplicity.
+
+    Solution:
+        Replace violating ``match ... case _`` statements with ``if ... else``.
+
+    When is this violation is issued?
```
sobolevn:
```suggestion
    When is this violation is raised?
```

we use "raise" in this context

wemake_python_styleguide/violations/consistency.py
```diff
@@ -2415,3 +2415,49 @@ class NotInWithUnaryOpViolation(ASTViolation):
...
+        - When there are exactly two cases
```
sobolevn:
```suggestion
        - When there are exactly two ``case`` statements
```

tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
```diff
@@ -0,0 +1,178 @@
...
+match subject:
+    case {0}:
+        pass
+    case _:
+        pass
+"""
+
+simplifiable_union_match_match = """
+match subject:
+    case {0} | {1}:
```
sobolevn: One last thing, please add more test with `as` keyword.

For example: 

```suggestion
    case {0} | {1}{2}:
```

where `{2}` can be `''` or `' as x'` during parametrizarion.

And other cases need `as` in different places for better coverage :)

wemake_python_styleguide/violations/consistency.py
```diff
@@ -2415,3 +2415,49 @@ class NotInWithUnaryOpViolation(ASTViolation):
...
+        (including wildcard ``_``)
+        is unnecessarily verbose and less performant than a plain ``if/else``.
+        Simple conditions should prefer readability and simplicity.
+
+    Solution:
+        Replace violating ``match ... case _`` statements with ``if ... else``.
+
+    When is this violation is raised?
+        - When there are exactly two ``case`` statements
+        - When the first case uses a simple pattern (e.g. liter, const, enums)
```
sobolevn:
```suggestion
        - When the first case uses a literal pattern
```

wemake_python_styleguide/violations/consistency.py
```diff
@@ -2415,3 +2415,49 @@ class NotInWithUnaryOpViolation(ASTViolation):
...
+        - When the second case is a wildcard: ``case _:``
+        - When the first case has simple guard (e.g. ``x if x > 0``)
```
sobolevn: What does "simple" mean here exactly?

sobolevn (commented):
👍

wemake_python_styleguide/violations/consistency.py
ZhdanovAM72: it's mean "literal" (i'm fixed it).

tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
```diff
@@ -93,61 +85,63 @@
...
-        'State.REJECTED',
-        'x if x > 0',
+        ('1', ''),
+        ('True', ''),
+        ('None', ''),
+        ('"string"', ''),
+        ('ns.CONST', ''),
+        ('State.REJECTED', ''),
+        ('x if x > 0', ''),
+        ('2', ' as x'),
```
sobolevn: This should be done via a second `@pytest.mark.parametrize` :)

- one for values
- one for `as`

```suggestion
        '2',
```

tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
```diff
@@ -93,61 +85,63 @@
...
+        ('False', ' as x'),
+        ('"string"', ' as x'),
+        ('ns.CONST', ' as x'),
+        ('State.REJECTED', ' as x'),
     ],
 )
 def test_simplifiable_single_match(
     code: str,
+    as_binding: str,
```
sobolevn: We don't use annotations in tests, so please remove them for consistency.

wemake_python_styleguide/visitors/ast/conditions.py
```diff
@@ -188,3 +190,34 @@ def visit_Compare(self, node: ast.Compare) -> None:
...
+                pattern_matching.is_wildcard_pattern(second)
+                and first.guard is None
+                and pattern_matching.is_simple_pattern(first.pattern)
+            ):
+                self.add_violation(consistency.SimplifiableMatchViolation(node))
+                return
+
+            if (
+                pattern_matching.is_wildcard_pattern(second)
+                and first.guard is not None
```
sobolevn:
```suggestion
```

I think that we can drop any `guard` references. When patterns are simple, we can simplify them with and without guards. When patterns are complex - we also don't care about guards.

So, the whole code can be:

```python
if not pattern_matching.is_wildcard_pattern(second):
    return

if (
    pattern_matching.is_irrefutable_binding(first.pattern)
    or pattern_matching.is_simple_pattern(first.pattern)
):
    self.add_violation(consistency.SimplifiableMatchViolation(node))
```

wemake_python_styleguide/violations/consistency.py
```diff
@@ -2415,3 +2415,49 @@ class NotInWithUnaryOpViolation(ASTViolation):
...
+        Simple conditions should prefer readability and simplicity.
+
+    Solution:
+        Replace violating ``match ... case _`` statements with ``if ... else``.
+
+    When is this violation is raised?
+        - When there are exactly two ``case`` statements
+        - When the first case uses a literal pattern
+        - When the second case is a wildcard: ``case _:``
+        - When the first case has literal guard (e.g. ``x if x > 0``)
```
sobolevn:
```suggestion
```

PR #3544, opened 2025-10-13
Feat/issue 3527

### LINKED ISSUE #3527: Prefer if over match simple sequences and mappings
# Rule request

## Thesis

Example:
```python
match data:
    case [1, 2]:
        handle_pair()
    case _:
        ignore()
```
Can be rewritten as:
```python
if data == [1, 2]:
    handle_pair()
else:
    ignore()
```

## Reasoning

When should this be allowed?

Only when:

Not allowed if:
- Has guards (`if` ...)

ZhdanovAM72:
# I have made things!

## Checklist

## Related issues
Closes #3527 

/wip


danak495-creator (approved):

sobolevn:
Fixed :)

PR #3559, opened 2025-11-11
feat: add SimplifiableIfMatchViolation for single-case matches

### LINKED ISSUE #3528: Prefer if over match with single-case
# Rule request

## Thesis

Example:
```python
match x:
    case 1:
        do_something()
```

Can be rewritten as:
```python
if x == 1:
    do_something()
```

## Reasoning

When should this be allowed?

Only when:
1. There is exactly one case

CommitToday:
## Overview
This PR adds a new violation `SimplifiableIfMatchViolation` that flags single-case `match` statements that can be simplified to `if` statements. This helps reduce unnecessary complexity and cognitive load when a simple `if` condition would suffice.

## Checklist
- [x] Code changes implemented
- [x] Tests added for new violation
- [x] Documentation included in violation docstring

## Proof
The new violation is tested with various cases in `test_simplified_match_conditions.py`:
- Single-case matches with simple literals (e.g., `case 1:`, `case True:`, `case "string":`) trigger the violation
- Single-case matches with as-binding also trigger the violation
- Multi-case matches, matches with guards, wildcard cases, and complex patterns do not trigger the violation
- The violation correctly identifies when a single-case match can be simplified to an if statement

Closes #3528

### DIFF
--- a/tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
+++ b/tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
@@ -1,6 +1,7 @@
 import pytest
 
 from wemake_python_styleguide.violations.consistency import (
+    SimplifiableIfMatchViolation,
     SimplifiableMatchViolation,
 )
 from wemake_python_styleguide.visitors.ast.conditions import (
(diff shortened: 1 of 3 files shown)


danak495-creator (approved):

sobolevn:
Fixed :)

PR #3560, opened 2025-11-11
feat: add check for simplifiable match with sequence or mapping patterns

### LINKED ISSUE #3527: Prefer if over match simple sequences and mappings

CommitToday:
## Overview
This PR adds a new check for simplifiable `match` statements that use simple sequences or mappings (like lists or dicts with literal values). The check identifies when a `match` statement can be replaced with a simpler `if` statement using equality comparison.

## Checklist
- [x] Code changes implemented
- [x] Tests added and passing
- [x] Documentation updated in violation docstring

## Proof
The new visitor `SimplifiableMatchWithSequenceOrMappingVisitor` checks for `match` statements that:
- Have exactly two cases
- The first case uses a simple sequence or mapping pattern (with literals, no variable bindings)
- The second case is a wildcard (`case _:`)
- No guards or starred patterns are present

When these conditions are met, it raises `SimplifiableMatchWithSequenceOrMappingViolation` with code 366. The test file `test_simplified_match_with_sequence_or_mapping.py` validates both simplifiable and non-simplifiable patterns.

## Closes #3527

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -212,3 +212,6 @@ fabric.properties
 # Used for testing:
 ex.py
 setup.py
+
+# CodeRabbit logs:
+logs/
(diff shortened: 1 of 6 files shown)


sobolevn:
Fixed :)

PR #3590, opened 2026-01-30
Fix wps430 false positive

### LINKED ISSUE #3589: False positive for `WPS430` with whitelisted name
```python
def wrap_handler(
    method: _MethodSyncHandler | _MethodAsyncHandler,
) -> SyncErrorHandlerT | AsyncErrorHandlerT:
    if inspect.iscoroutinefunction(method):

        @wraps(method)
        async def decorator(
            endpoint: 'Endpoint',
            controller: 'Controller[BaseSerializer]',
            exc: Exception,
        ) -> HttpResponse:
            return await method(  # type: ignore[no-any-return]
                controller.active_blueprint,
                endpoint,
                controller,
                exc,
            )

    else:

        @wraps(method)
        def decorator(
            endpoint: 'Endpoint',
            controller: 'Controller[BaseSerializer]',
            exc: Exception,
        ) -> HttpResponse:
            return method(  # type: ignore[return-value]
                controller.active_blueprint,
                endpoint,
                controller,
                exc,
            )

    return decorator
```

This code falsy raises two `WPS430` violations.

```
  69:9     WPS430 Found nested function: factory
  async def factory(
  ^

  84:9     WPS430 Found nested function: factory
  def factory(
  ^
```

But, the name is correct. `decorator` name is in whitelist of names. This is a false positive.
We need to fix it and add a test case.

albertedwardson:
# I have made things!

## Checklist

## Related issues

Closes #3589 

Actually, there was already a test case for this, but it was wrong and assumed that deep nested whitelisted functions must be reported.

### DIFF
--- a/tests/test_visitors/test_ast/test_complexity/test_nested/test_nested_functions.py
+++ b/tests/test_visitors/test_ast/test_complexity/test_nested/test_nested_functions.py
@@ -257,8 +257,7 @@ def test_deep_whitelist_nested_functions(
     visitor = NestedComplexityVisitor(default_options, tree=tree)
     visitor.run()
 
-    assert_errors(visitor, [NestedFunctionViolation])
-    assert_error_text(visitor, whitelist_name)
+    assert_errors(visitor, [])
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


sobolevn (commented):
Please, add a test case :)

PR #3591, opened 2026-01-31
Fix `WPS430` false positive on whitelisted deep nested functions

### LINKED ISSUE #3589: False positive for `WPS430` with whitelisted name

albertedwardson:
# I have made things!

## Checklist

## Related issues

Closes #3589 

Actually, there was already a test case for this `/wemake-python-styleguide/tests/test_visitors/test_ast/test_complexity/test_nested/test_nested_functions.py`, but it was wrong and assumed that deep nested whitelisted functions must be reported.

### DIFF
--- a/tests/fixtures/noqa/noqa.py
+++ b/tests/fixtures/noqa/noqa.py
@@ -122,20 +122,13 @@ def some():  # noqa: WPS110
 
     def nested():  # noqa: WPS430
         ...
-    
-    def factory():  # ok
-        if some_condition():
-            def deep_nested():  # noqa: WPS430
-                    ...
-        else:
-            def wrapper(): # ok
-                ...
-            if some_other_condition():
-                def deep_nested():  # noqa: WPS430
-                    ...
-            else:
-                def decorator(): # ok
-                    ...
+
+    if some_condition():
+        def deep_nested():  # noqa: WPS430
+            ...
+    else:
+        async def deep_nested():  # noqa: WPS430
+            ...
 
 
 del {'a': 1}['a']  # noqa: WPS420
--- a/tests/fixtures/noqa/noqa.py
+++ b/tests/fixtures/noqa/noqa.py
@@ -122,6 +122,20 @@ def some():  # noqa: WPS110
 
     def nested():  # noqa: WPS430
         ...
+    
+    def factory():  # ok
+        if some_condition():
+            def deep_nested():  # noqa: WPS430
+                    ...
+        else:
+            def wrapper(): # ok
+                ...
+            if some_other_condition():
+                def deep_nested():  # noqa: WPS430
+                    ...
+            else:
+                def decorator(): # ok
+                    ...
 
 
 del {'a': 1}['a']  # noqa: WPS420
--- a/tests/test_checker/test_noqa.py
+++ b/tests/test_checker/test_noqa.py
@@ -197,7 +197,7 @@ SHOULD_BE_RAISED = types.MappingProxyType(
         'WPS427': 1,
         'WPS428': 0,  # disabled since 1.0.0
         'WPS429': 1,
-        'WPS430': 1,
+        'WPS430': 3,
         'WPS431': 2,
         'WPS432': 2,
         'WPS433': 0,  # disabled since 1.0.0
--- a/tests/test_visitors/test_ast/test_complexity/test_nested/test_nested_functions.py
+++ b/tests/test_visitors/test_ast/test_complexity/test_nested/test_nested_functions.py
@@ -257,8 +257,7 @@ def test_deep_whitelist_nested_functions(
     visitor = NestedComplexityVisitor(default_options, tree=tree)
     visitor.run()
 
-    assert_errors(visitor, [NestedFunctionViolation])
-    assert_error_text(visitor, whitelist_name)
+    assert_errors(visitor, [])
 
 
 @pytest.mark.parametrize(
--- a/tests/test_visitors/test_ast/test_complexity/test_nested/test_nested_functions.py
+++ b/tests/test_visitors/test_ast/test_complexity/test_nested/test_nested_functions.py
@@ -234,6 +234,34 @@ def test_whitelist_nested_functions(
     [
         nested_function_in_if,
         nested_function_in_if_else,
+    ],
+)
+def test_deep_whitelist_nested_functions_allowed(
+    assert_errors,
+    assert_error_text,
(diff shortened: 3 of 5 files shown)


albertedwardson:
@sobolevn hi! I opened a new PR with correct branch name and other edits

tests/fixtures/noqa/noqa.py
```diff
@@ -122,6 +122,20 @@ class Nested:  # noqa: WPS431
 
     def nested():  # noqa: WPS430
         ...
+    
+    def factory():  # ok
```
sobolevn: No need to put this into `noqa.py`. Here we test that `# noqa` comments work correctly ;)

tests/test_checker/test_noqa.py
```diff
@@ -197,7 +197,7 @@
         'WPS427': 1,
         'WPS428': 0,  # disabled since 1.0.0
         'WPS429': 1,
-        'WPS430': 1,
+        'WPS430': 3,
```
sobolevn, line 200: Let's revert these changes.

```suggestion
        'WPS430': 1,
```

sobolevn (commented):
Let's only allow `def` in a single function-level `if`. Nothing more.

Please, also write unit-tests :)

albertedwardson:
> Let's only allow `def` in a single function-level `if`. Nothing more.

Ok, sure, didn't catch that from the issue itself, thought that it is ok to define whitelisted functions regardless of nesting and what causes that nesting 

> Please, also write unit-tests :)

Don't `tests/test_visitors/test_ast/test_complexity/test_nested/test_nested_functions.py` already cover that case? Or do you mean I should extend it with a case where whitelisted name in, e.g. `try/except` or `for` block, that will be reported as violation, and separate them from single `if` block, where it is allowed? If so, should I also explicitly document this behavior in documentation and add an example with `inspect.iscoroutinefunction`?

tests/test_visitors/test_ast/test_complexity/test_nested/test_nested_functions.py
```diff
@@ -257,8 +257,7 @@ def test_deep_whitelist_nested_functions(
     visitor = NestedComplexityVisitor(default_options, tree=tree)
     visitor.run()
 
-    assert_errors(visitor, [NestedFunctionViolation])
-    assert_error_text(visitor, whitelist_name)
+    assert_errors(visitor, [])
```
albertedwardson: @sobolevn 

> Nope, we modified code and none tests failed :)

Not because I actually changed assertion in that case in this PR? As I meantioned before, this test assumed that *all* deep nested function definitions with whitelisted names should be reported as violation, regardless what causes that indentation


Did you mean that we just need exclude 

https://github.com/wemake-services/wemake-python-styleguide/blob/5198e4051fe60292e77d4f17979814e1d4576525/tests/test_visitors/test_ast/test_complexity/test_nested/test_nested_functions.py#L235-L236

into a separate test case that asserts there is no violation with `if` and `if/else` blocks, while keeping other deep nesting cases reported as violations?

Sorry for asking so many questions and not understanding what needed to be done from the first try :(

tests/test_checker/test_noqa.py
albertedwardson: I've added minimal reproducible example as part of integration tests (if I didn't misunderstood what `CONTRIBUTING.md` states)

wemake_python_styleguide/visitors/ast/complexity/nested.py
```diff
@@ -56,13 +56,23 @@ def visit_Lambda(self, node: ast.Lambda) -> None:
...
 
-        is_direct = isinstance(get_parent(node), FunctionNodes)
-        is_bad = is_direct and node.name not in NESTED_FUNCTIONS_WHITELIST
+        parent = get_parent(node)
 
-        if is_bad or (is_inside_function and not is_direct):
-            self.add_violation(NestedFunctionViolation(node, text=node.name))
+        is_direct = isinstance(parent, FunctionNodes)
+
+        is_single_if = isinstance(parent, ast.If) and isinstance(
```
sobolevn: `get_parent(parent) is get_context(node)` that's the correct check :)

wemake_python_styleguide/visitors/ast/complexity/nested.py
```diff
@@ -56,13 +56,23 @@ def visit_Lambda(self, node: ast.Lambda) -> None:
...
+            get_parent(parent), FunctionNodes
+        )
+
+        if node.name not in NESTED_FUNCTIONS_WHITELIST or not (
```
sobolevn: Let's use early returns here

albertedwardson: like this? 
```python
        if node.name not in NESTED_FUNCTIONS_WHITELIST:
            self.add_violation(
                NestedFunctionViolation(node, text=node.name),
            )
            return

        if not is_direct and not is_single_if:
            self.add_violation(
                NestedFunctionViolation(node, text=node.name),
            )
```

sobolevn: `if node.name in NESTED_FUNCTIONS_WHITELIST and (is_direct or is_single_if): return` :)

sobolevn (approved):
Great work! Thanks!

PR #3602, opened 2026-02-16
Remove WPS354

### LINKED ISSUE #3601: Remove `WPS354`
This rules has more problems that it solves:
1. Two `yield`s and `yield from` are two different things
2. Async iterators cannot use `yield from`, but `WPS354` still raises even in `async def`

So, it is time to remove this rule. I was wrong about that :(
See other rules with `disabled_since` attribute on how to disable a rule.

j2cry:
# I have made things!
Hi! I tried to do this.

## Checklist

## Related issues
Closes #3601


sobolevn (approved):
Great! Thank you! 🤝

PR #3604, opened 2026-02-22
fix: WPS457 false positive when `while True` is inside `try/except`

### LINKED ISSUE #3603: `WPS457` false positive with `while` in `try` / `except`
I have the following code:

```python
def factory() -> Iterator[bytes]:
    try:
        while True:  # noqa: WPS457
            yield async_to_sync(async_anext)(iterator)
    except StopAsyncIteration:
        pass
```

It should not raise `WPS457`, because `while` is in `try` / `except`.
It can clearly raise and we handle the error. So, it is not infinite.

PRs are welcome! 👍

stakeswky:
Fixes #3603

## Problem

`WPS457` (InfiniteWhileLoopViolation) was incorrectly triggered when a `while True` loop was wrapped in a `try/except` block:

```python
def factory() -> Iterator[bytes]:
    try:
        while True:
            yield async_to_sync(async_anext)(iterator)
    except StopAsyncIteration:
        pass
```

This is not an infinite loop — the `except` handler provides a clear termination path. The previous implementation only checked for `break`/`raise`/`return`/`ExceptHandler` *inside* the `while` body, but missed the case where the enclosing `try/except` itself acts as the exit mechanism.

## Fix

Added `is_in_try_except()` helper in `wemake_python_styleguide/logic/tree/loops.py` that walks up the AST parent chain to detect if the `while` node is directly inside a `try` block that has at least one `except` handler.

In `_check_infinite_while_loop`, if `is_in_try_except()` returns `True`, the violation is suppressed.

**Note:** A `try/finally` without `except` still triggers WPS457, since `finally` does not catch exceptions and cannot terminate the loop.

## Tests

Added three new correct cases (`correct_while8/9/10`) covering the reported pattern, and one new wrong case (`wrong_while3`) for `try/finally` without `except`.

### DIFF
--- a/tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
+++ b/tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
@@ -126,6 +126,43 @@ while True:
         ...
 """
 
+wrong_while3 = """
+try:
+    while True:
+        do_something()
+finally:
+    cleanup()
+"""
+
+# Correct: while inside try/except
+
+correct_while8 = """
+def gen():
+    try:
+        while True:
+            yield some()
+    except StopIteration:
+        pass
+"""
+
+correct_while9 = """
+def factory():
+    try:
+        while True:
+            yield async_to_sync(anext)(iterator)
+    except StopAsyncIteration:
+        pass
+"""
+
+correct_while10 = """
+try:
+    while 1:
+        do_something()
+except SomeError:
+    handle()
+"""
+
+
 
 @pytest.mark.parametrize(
     'template',
@@ -173,6 +210,9 @@ def test_correct_while_loops_with_statements(
         correct_while5,
         correct_while6,
         correct_while7,
+        correct_while8,
+        correct_while9,
+        correct_while10,
     ],
 )
 def test_correct_while_loops_with_try(
@@ -259,6 +299,7 @@ def test_wrong_while_loops(
     [
         wrong_while1,
         wrong_while2,
+        wrong_while3,
     ],
 )
 def test_wrong_while_loops_with_try(
--- a/tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
+++ b/tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
@@ -163,7 +163,6 @@ except SomeError:
 """
 
 
-
 @pytest.mark.parametrize(
     'template',
     [
--- a/tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
+++ b/tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
@@ -105,6 +105,36 @@ async def worker():
         await asyncio.sleep(1)
 """
 
+# Correct: while inside try/except
+
+correct_while8 = """
+def gen():
+    try:
+        while True:
+            yield some()
+    except StopIteration:
+        pass
+"""
+
+correct_while9 = """
+try:
+    while 1:
+        do_something()
+except SomeError:
+    handle()
+"""
+
+correct_while10 = """
+try:
+    while True:
+        try:
+            do_something()
+        except InnerError:
+            pass
+except OuterError:
+    handle()
+"""
+
 # Do raise:
 
 wrong_while1 = """
@@ -132,34 +162,6 @@ try:
         do_something()
 finally:
     cleanup()
-"""
-
-# Correct: while inside try/except
-
-correct_while8 = """
-def gen():
-    try:
-        while True:
-            yield some()
-    except StopIteration:
-        pass
-"""
-
-correct_while9 = """
-def factory():
-    try:
-        while True:
-            yield async_to_sync(anext)(iterator)
-    except StopAsyncIteration:
-        pass
-"""
-
-correct_while10 = """
-try:
-    while 1:
-        do_something()
-except SomeError:
-    handle()
 """
 
 
--- a/tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
+++ b/tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
@@ -108,12 +108,12 @@ async def worker():
 # Correct: while inside try/except
 
 correct_while8 = """
-def gen():
-    try:
+try:
+    if some:
         while True:
-            yield some()
-    except StopIteration:
-        pass
+            yield some
+except StopIteration:
+    pass
 """
 
 correct_while9 = """
@@ -132,6 +132,17 @@ try:
         except InnerError:
             pass
 except OuterError:
+    handle()
+"""
+
+# Wrong: try/except is outside a scope boundary, should not exempt the while
+
+wrong_while4 = """
+try:
+    def inner():
+        while True:
+            do_something()
+except SomeError:
     handle()
 """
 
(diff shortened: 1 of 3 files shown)


tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
```diff
@@ -126,6 +126,42 @@ async def worker():
...
+try:
+    while True:
+        do_something()
+finally:
+    cleanup()
+"""
+
+# Correct: while inside try/except
+
+correct_while8 = """
```
sobolevn: Please, move correct examples to correct examples' section.

tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
```diff
@@ -126,6 +126,42 @@ async def worker():
...
+def gen():
+    try:
+        while True:
+            yield some()
+    except StopIteration:
+        pass
+"""
+
+correct_while9 = """
```
sobolevn: All these examples are the same from the AST's point of view.

Please, remove duplicates.

wemake_python_styleguide/logic/tree/loops.py
```diff
@@ -20,6 +21,21 @@ def _does_loop_contain_node(
     return any(to_check is inner_node for inner_node in ast.walk(loop))
 
 
+def is_in_try_except(node: ast.AST) -> bool:
+    """Checks whether a node is directly inside a ``try/except`` block."""
+    parent = get_parent(node)
```
sobolevn: Why not using `get_closest_parent()`?

tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
```diff
@@ -126,6 +126,42 @@ async def worker():
...
+def factory():
+    try:
+        while True:
+            yield async_to_sync(anext)(iterator)
+    except StopAsyncIteration:
+        pass
+"""
+
+correct_while10 = """
+try:
```
sobolevn: Please, add more nesting in the tests. But, keep them minimal.

sobolevn (commented):
Thanks! If you are using AI, please, double check the produced results :)

stakeswky (commented):
Thanks for the detailed review! I've addressed all four points:

1. **`get_closest_parent()`**: Refactored `is_in_try_except()` to use `walk.get_closest_parent(node, (*_SCOPE_BOUNDARIES, ast.Try))` — finds the nearest `Try` or scope-boundary parent in one call, then checks if it's a `Try` with handlers.

2. **Correct examples section**: Moved `correct_while8`/`9`/`10` above the `# Do raise:` section where they belong.

3. **Removed duplicate**: `correct_while9` (same AST as `correct_while8` — both are `def fn(): try: while True: yield; except: pass`) has been removed and replaced with a simpler top-level `try/except` case.

4. **More nesting**: Added `correct_while10` with a nested `try/except` inside the `while True` body, inside an outer `try/except`.

tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
stakeswky: Done — moved correct examples up to the correct section.

tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
stakeswky: Good point. Removed the duplicates — kept one simple case (`correct_while9`) and replaced the other with a nested variant (`correct_while10`) so they're structurally distinct.

wemake_python_styleguide/logic/tree/loops.py
stakeswky: Refactored to use `get_closest_parent()` — much cleaner. Thanks for the pointer.

tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
stakeswky: Added `correct_while10` with nested try/except inside the while loop to cover deeper nesting, kept it minimal.

tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
```diff
@@ -105,6 +105,36 @@ async def worker():
...
 
+# Correct: while inside try/except
+
+correct_while8 = """
+def gen():
+    try:
+        while True:
+            yield some()
+    except StopIteration:
+        pass
```
sobolevn:
```suggestion
try:
    if some:
        while True:
            yield some
except StopIteration:
    pass
```

wemake_python_styleguide/logic/tree/loops.py
```diff
@@ -1,9 +1,11 @@
 import ast
 
 from wemake_python_styleguide.compat.aliases import ForNodes
-from wemake_python_styleguide.logic.nodes import get_parent
+from wemake_python_styleguide.logic.walk import tree as walk
 from wemake_python_styleguide.types import AnyLoop, AnyNodes
 
+_SCOPE_BOUNDARIES = (ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)
```
sobolevn: You don't need a variable for a single use of values.

tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
```diff
@@ -105,6 +105,36 @@ async def worker():
...
+
+correct_while9 = """
+try:
+    while 1:
+        do_something()
+except SomeError:
+    handle()
+"""
+
+correct_while10 = """
```
sobolevn, line 127: you never test `try` being further in hierarchy then scope boundaries.

tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
stakeswky: Applied your suggestion — replaced with `try: if some: while True: yield some; except StopIteration: pass` to test deeper nesting.

wemake_python_styleguide/logic/tree/loops.py
stakeswky: Done — removed `_SCOPE_BOUNDARIES` and inlined the tuple directly in the `get_closest_parent()` call.

tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
stakeswky: Added `wrong_while4`: a `while True` inside a function that is itself inside a `try/except` — the scope boundary (function def) should block the try from exempting the loop, so it raises `InfiniteWhileLoopViolation`. Included in `test_wrong_while_loops_with_try`.

tests/test_visitors/test_ast/test_loops/test_loops/test_infinite_while_loops.py
```diff
@@ -135,6 +135,17 @@ def gen():
     handle()
 """
 
+# Wrong: try/except is outside a scope boundary, should not exempt the while
```
sobolevn: This comment is not required.

stakeswky: Removed.

sobolevn (approved):

PR #3608, opened 2026-02-25
WPS237: forbid complex f-string format specifiers

### LINKED ISSUE #3491: Make sure that we forbid all f-string formatting
https://fstrings.wtf/

We need to be sure that no formatting for `f` strings are allowed. Maybe except `=` part.

In case it is already disallowed, we need to improve tests for it.

j2cry:
# I have made things!

Hi!

This forbids all f-string format patterns, except the following
```python
f'{value!r}'
f'{value!a}'
f'{value:x}'
f'{value:X}'
f'{value:#x}'
f'{value:#X}'
f'{value:c}'
f'{value:.4f}'  # it can be any number, not necessarily 4
f'{value:.4e}'  # it can be any number, not necessarily 4
f'{value:<4}'
```

The next one I suggest as a workaround for complex formatting.
It keeps f-string readable and preserves its functionality.
```python
f'{value:{fmt}}'
```
Since the processing rules for the string and the format specifier are the same,
the following expressions are also valid:
```python
f'{value:{fmt[0]}}'
f'{value:{fmt.attr}}'
f'{value:{fmt[None]}}'
f'{value:{fmt[True]}}'
f'{value:{fmt()}}'
# etc
```
Format patterns consisting of multiple elements are banned.

I still think that this is not a strict enough rule, but finding a balance on this issue is not easy.
At least I tried 🥲.

## Checklist

## Related issues
Closes #3491

### DIFF
--- a/wemake_python_styleguide/visitors/ast/builtins.py
+++ b/wemake_python_styleguide/visitors/ast/builtins.py
@@ -16,6 +16,7 @@ from wemake_python_styleguide.logic.complexity.annotations import (
 from wemake_python_styleguide.logic.tree import (
     attributes,
     operators,
+    strings,
     variables,
 )
 from wemake_python_styleguide.types import (
@@ -82,7 +83,7 @@ class WrongStringVisitor(base.BaseNodeVisitor):
 
 
 @final
-class WrongFormatStringVisitor(base.BaseNodeVisitor):  # noqa: WPS214
+class WrongFormatStringVisitor(base.BaseNodeVisitor):
     """Restricts usage of ``f`` strings."""
 
     _valid_format_index: ClassVar[AnyNodes] = (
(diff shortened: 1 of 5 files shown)


wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -81,7 +82,7 @@ def _check_is_alphabet(
 
 
 @final
-class WrongFormatStringVisitor(base.BaseNodeVisitor):
+class WrongFormatStringVisitor(base.BaseNodeVisitor):  # noqa: WPS214
```
sobolevn: Let's refactor it :)

wemake_python_styleguide/visitors/ast/builtins.py
```diff
@@ -101,24 +102,62 @@ class WrongFormatStringVisitor(base.BaseNodeVisitor):
...
+        ):
+            # check if a string component contains a valid format pattern
+            matched = re.match(
+                r'^(\.\d+(?:f|e)|[<^>]\d+|#?[x|X]|[c,])$',
+                string_component.value,
+            )
+            return bool(matched)
+        return True
+
+    def _has_conversion(self, string_component: ast.AST) -> bool:
```
sobolevn: For example: these methods can be in `logic/`

sobolevn (commented):
please, explicitly test `=` to be allowed. I might have missed it in many different cases :)

sobolevn:
I would love to merge it and release `1.6.0` with it :)

sobolevn (approved):
Great job! Release is on its way.

PR #3612, opened 2026-02-27
Update docs to reflect removed config knob

Maroloccio:
Update documentation to reflect the fact that there is no configuration setting for the check anymore.


sobolevn (approved):
Thanks!

PR #3614, opened 2026-02-28
WPS330: fix false positive when alternating operators

### LINKED ISSUE #3594: WPS330 false positive
Hi! While working on #3593 I found this case
```python
value = -~-7  # -> -6
value = ~-~7  # -> -9
```
Why is it marked as unnecessary operator (WPS330)?
I think this code should definitely be banned, but WPS330 is not applicable here.
Maybe only for `ast.IAdd` operator, because it doesn't affect the value.
```python
value = +~+++~++7  # -> 7
```

Can we use something like the following instead of `count_unary_operator()` to fix it?
```python
# SingleUnaryOperator: TypeAlias = type[ast.UAdd] | type[ast.USub] | type[ast.Invert] | type[ast.Not]

def count_consecutive_unary_operator(
    node: ast.AST,
    operator: type[ast.unaryop],  # I have doubts about the correctness of this annotation, but can imagine that someday we will need to count i.e. UAdd and USub together
    # operator: SingleUnaryOperator,  # alternative
) -> int:
    """Counts the maximum number of consecutive identical operators"""
    counter = 0
    max_counter = 0
    parent = node
    while True:
        parent = get_parent(parent)
        if not isinstance(parent, ast.UnaryOp):
            break
        
        if isinstance(parent.op, operator):
            counter += 1
        elif counter > max_counter:
            max_counter = counter
            counter = 0
        else:
            counter = 0

    return max(max_counter, counter)
```
or in a recursive form (which one is preferred?).

What do you think about this?

j2cry:
# I have made things!

## Checklist

## Related issues
Closes #3594

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Semantic versioning in our case means:
 ### Bugfixes
 
 - Fixes false positive `WPS366` allowing the use of a single constant in `or`, #3610
+- Fixes the false positive `WPS330` when alternating unary operators, #3594
 
 ## 1.6.0
 
--- a/tests/test_visitors/test_ast/test_operators/test_useless_operators_before_numbers.py
+++ b/tests/test_visitors/test_ast/test_operators/test_useless_operators_before_numbers.py
@@ -50,13 +50,15 @@ usages = (
         '---5',
         '~~5',
         '~~~5',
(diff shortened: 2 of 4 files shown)


sobolevn (approved):
Good idea, thanks! I will release 1.6.1 today :)

PR #3634, opened 2026-04-23
fix: WPS529 false positive

### LINKED ISSUE #3501: False positive for `WPS529`
Code like this

```python
def func(**kwargs):
    if 'a' in kwargs:
        assert 'a_kwargs' not in kwargs
    else:
        kwargs['a'] = other(
            **kwargs.pop('a_kwargs', {}),
        )
```

raises `WPS529`

I don't think that it is correct, because we can't replace it with `.get`.

```python
    # Correct:
    value = collection.get(key)
    if value is not None:
        print(value)

    # Wrong:
    if key in collection:
        print(collection[key])
```

vldmrdev:
# I have made things!
1. Modified `ImplicitDictGetVisitor._check_implicit_get` and add `_walk_body ` func, `else` exluded from check scope.
2. Added negative test and test case for `else` 

## Checklist

## Related issues

Closes #3501 

### DIFF
--- a/tests/test_visitors/test_ast/test_subscripts/test_implicit_get.py
+++ b/tests/test_visitors/test_ast/test_subscripts/test_implicit_get.py
@@ -44,6 +44,13 @@ elif other:
     ...
 """
 
+else_subscripts = """
+if 'a' in kwargs:
+    pass
+else:
+    kwargs['a'] = other(kwargs)
+"""
+
 
 @pytest.mark.parametrize(
     'template',
@@ -117,6 +124,27 @@ def test_correct_if(
 ):
     """Testing that correct `if` can be used."""
     tree = parse_ast_tree(template.format(compare, expression))
+
+    visitor = ImplicitDictGetVisitor(default_options, tree=tree)
+    visitor.run()
+
+    assert_errors(visitor, [])
+
+
+@pytest.mark.parametrize(
+    'code',
+    [
+        else_subscripts,
+    ],
+)
+def test_no_implicit_dict_get_in_else_branch(
+    assert_errors,
+    parse_ast_tree,
+    default_options,
+    code,
+):
+    """Testing that subscripts in `else` branch do not trigger violation."""
+    tree = parse_ast_tree(code)
 
     visitor = ImplicitDictGetVisitor(default_options, tree=tree)
     visitor.run()
--- a/wemake_python_styleguide/violations/refactoring.py
+++ b/wemake_python_styleguide/violations/refactoring.py
@@ -1135,9 +1135,6 @@ class ImplicitDictGetViolation(ASTViolation):
             print(collection[key])
 
     .. versionadded:: 0.13.0
-    .. versionadded:: 1.7.0
-        Excluded the `else` branch, check scope changed from
-        the whole ast.If to node.body.
     """
 
     error_template = 'Found implicit `.get()` dict usage'
--- a/wemake_python_styleguide/violations/refactoring.py
+++ b/wemake_python_styleguide/violations/refactoring.py
@@ -1135,7 +1135,9 @@ class ImplicitDictGetViolation(ASTViolation):
             print(collection[key])
 
     .. versionadded:: 0.13.0
-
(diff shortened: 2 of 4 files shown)


tests/test_visitors/test_ast/test_subscripts/test_implicit_get.py
```diff
@@ -44,6 +44,30 @@
...
+
+elif_subscripts1 = """
+if 'b' in kwargs:
+    pass
+elif 'a' in kwargs:
+    print(kwargs['a'])
+"""
+
+if_elif_else_subscripts = """
+def func(d):
```
sobolevn: why do we need a function here? we try to keep the examples as minimals as possible.

tests/test_visitors/test_ast/test_subscripts/test_implicit_get.py
```diff
@@ -44,6 +44,30 @@
     ...
 """
 
+else_subscripts1 = """
```
sobolevn: Please, split these examples into `# Correct:` and `# Wrong:` ones.

wemake_python_styleguide/visitors/ast/subscripts.py
```diff
@@ -136,7 +142,7 @@ def _check_implicit_get(self, node: ast.If) -> None:
         checked_key = source.node_to_string(node.test.left)
         checked_collection = source.node_to_string(node.test.comparators[0])
 
-        for sub in ast.walk(node):
+        for sub in self._walk_body(node.body):
```
sobolevn, line 145:
```suggestion
        for body_node in node.body:
            for sub in ast.walk(body_node):
```

wemake_python_styleguide/visitors/ast/subscripts.py
```diff
@@ -127,6 +128,11 @@ def visit_If(self, node: ast.If) -> None:
         self._check_implicit_get(node)
         self.generic_visit(node)
 
+    def _walk_body(self, body: list[ast.stmt]) -> Iterator[ast.AST]:
```
sobolevn, line 131: We can remove this method

wemake_python_styleguide/visitors/ast/subscripts.py
vldmrdev: It was my first solution but I got this error from linter:

`error[C901]: `_check_implicit_get` is too complex (7 > 6)`

So that's why I moved it to an external function `_walk_body`. Should I ignore this error?

sobolevn: ok then :)

tests/test_visitors/test_ast/test_subscripts/test_implicit_get.py
vldmrdev: > Please, split these examples into # Correct: and # Wrong: ones.

I need clarification. This test case - `else_subscripts1` is needed to check for false positives and there are no other options because WPS529 should ignores the use of a collection index inside an `else` block (orelse in AST), excluding new `if` blocks.
I suggest keeping this test case `else_subscripts1` and remove the others, because they repeat the tests in the templates.

wemake_python_styleguide/violations/refactoring.py
```diff
@@ -1135,7 +1135,9 @@ class ImplicitDictGetViolation(ASTViolation):
             print(collection[key])
 
     .. versionadded:: 0.13.0
-
+    .. versionadded:: 1.7.0
```
sobolevn: You can remove this note, it is only needed for functional changes, not bug fixes :)

wemake_python_styleguide/violations/refactoring.py
```diff
@@ -1135,7 +1135,9 @@ class ImplicitDictGetViolation(ASTViolation):
...
+        Excluded the `else` branch, check scope changed from
+        the whole ast.If to node.body.
```
sobolevn:
```suggestion
```

sobolevn (approved):
Thanks a lot!

PR #3695, opened 2026-07-02
add TooManyComments (#3689)

### LINKED ISSUE #3689: add TooManyCommentsViolation
# Rule request

## Thesis

Nowadays, developers often use an excessive number of comments, which leads to unclear code, loss of readability, increased cognitive load, trying replace good documentation.

This rule complements [WPS 402.](https://wemake-python-styleguide.readthedocs.io/en/latest/pages/usage/violations/best_practices.html#wemake_python_styleguide.violations.best_practices.OveruseOfNoqaCommentViolation). 
Let's be honest: people often write comments because their code is less readable and overly complex.
## Reasoning

The most appropriate solution is to introduce a new violation called `TooManyCommentsViolation`. 
This violation would enforce a maximum of `N` comments inside functions or methods (and in the future, we could extend it to all files).

Khabib73:
# I have made things!

## Checklist

## Related issues

- Closes #3689

### DIFF
--- a/tests/fixtures/noqa/noqa.py
+++ b/tests/fixtures/noqa/noqa.py
@@ -625,6 +625,67 @@ match subject:  # noqa: WPS365
     case _:
         ...
 
+def too_many_locals():  # noqa: WPS482
+    # comment 1
+    abcd = 1
+    # comment 2
+    abcdx = 2
+    # comment 3
+    # comment 4
+    # comment 5
+    # comment 6
+    # comment 7
+    # comment 8
+    # comment 9
+    # comment 10
+    # comment 11
+    # comment 12
+    # comment 13
+    # comment 14
+    # comment 15
+    # comment 16
+
+    ...
+
+def too_many_nested():
+    # comment 1
+    abcd = 1
+    # comment 2
+    abcdx = 2
+    def factory():
+        # this is a edge case
+        # comment 2
+        # comment 3
+        # comment 4
+        # comment 5
+        # comment 6
+        # comment 7
+        # comment 8
+        # comment 9
+        # comment 10
+        # comment 11
+        # comment 12
+        # comment 13
+        # comment 14
+        # comment 15
+        ...
+    def decorator():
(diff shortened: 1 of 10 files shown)


Khabib73:
I'll squash the commits at the end, please don't squash on merge :)

Khabib73:
@sobolevn Hi! Could you please take a look at my PR, when you have a moment? Thanks!

tests/fixtures/noqa/noqa.py
```diff
@@ -625,6 +625,67 @@ def many_raises_function(parameter):  # noqa: WPS238
     case _:
         ...
 
+def too_many_locals():  # noqa: WPS482
```
sobolevn, line 628: In `test_noqa` only one example is needed, here we test:
- noqa position
- full plugin integration

wemake_python_styleguide/options/config.py
```diff
@@ -239,6 +242,11 @@ class Configuration:
             defaults.MAX_NOQA_COMMENTS,
             'Maximum amount of `noqa` comments per module.',
         ),
+        _Option(
+            '--max-comments-in-function',
```
sobolevn, line 246: I don't think that it should be this simple. We should count the code / comments rate.

There might be several important cases:
- Where the amount of comments are far greater then the code, it is totally valid, if we need to explain something
- Where there are just some comments, like 1 or 2
- Where the amount of comments are zero

These cases are valid.
What cases are not that valid?

```python
def some():
    x = 1  # define x
    y = 2  # define y
    z = 3  # define z
    return x + y  # return their sum
```

We should also ignore `type: ` comments, `# noqa` comments, `# pyright`, `# ty`, `# ruff`, and `# pyrefly` comments.

So, basically we need to find a lot of slop code, analyze its comments and patterns, try to formalize it, ban it.

This is a very complex task :)

Khabib73: Thanks for the details!

PR #3700, opened 2026-07-04
feat #3528

### LINKED ISSUE #3528: Prefer if over match with single-case

Khabib73:
# I have made things!

I also had to fix tests WPS536, WPS241, because they found several violations at once. Moreover I added new test for WPS365, so now there are two of them :)

## Checklist

## Related issues

Closes #3528

### DIFF
--- a/tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
+++ b/tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
@@ -32,6 +32,30 @@ match subject:
         pass
 """
 
+single_case_sequence = """
+match subject:
+    case [1, 2, 3]:
+        pass
+"""
+
+single_case_wildcard = """
+match subject:
+    case _:
+        pass
+"""
+
+single_case_literal = """
+match subject:
+    case '1':
+        pass
+"""
+
+single_case_guard = """
+match subject:
+    case x if x > 0:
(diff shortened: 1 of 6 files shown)


sobolevn (commented):
Please, also add unit tests for this.

Docs for WPS365:

```
When is this violation is raised?
- When there are exactly two ``case`` statements
- When the first case uses a literal pattern
- When the second case is a wildcard: ``case _:``
```

I think that we should update them as well.

tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
```diff
@@ -32,6 +32,24 @@
         pass
 """
 
+single_case_sequence = """
+match subject:
+    case [1, 2, 3]:
```
sobolevn, line 37: Please, also test `case _: ...`

sobolevn (commented):
Don't forget to update the CHANGELOG as well 😉 
Thanks a lot for your help!

sobolevn (approved):
Thank you!

PR #3701, opened 2026-07-04
feat(consistency): add SingleCaseMatchViolation (WPS367)

### LINKED ISSUE #3528: Prefer if over match with single-case

mmustafasenoglu:
## Description

Adds a new violation `SingleCaseMatchViolation` (WPS367) to detect single-case `match` statements that can be simplified to a plain `if` statement.

## Problem

A `match` statement with only one `case` and no wildcard/default is unnecessarily verbose and less readable than a simple `if` statement:

```python
# Bad (WPS367)
match x:
    case 1:
        do_something()

# Good
if x == 1:
    do_something()
```

## Changes

- **violations/consistency.py**: Added `SingleCaseMatchViolation` class (code 367)
- **visitors/ast/conditions.py**: Added `len(cases) == 1` check to `SimplifiableMatchVisitor._check_simplifiable_match`
- **tests/test_visitors/test_ast/test_conditions/test_single_case_match.py**: Added test cases for violation and non-violation scenarios

## Test Cases

- Single case with literal patterns (1, True, None, strings) -> violation
- Single case with `as` binding -> violation
- Single case with guard (`if check()`) -> no violation
- Multiple cases -> no violation
- Wildcard pattern -> no violation
- Irrefutable binding -> no violation

Closes #3528

### DIFF
--- a/tests/test_visitors/test_ast/test_conditions/test_single_case_match.py
+++ b/tests/test_visitors/test_ast/test_conditions/test_single_case_match.py
@@ -1,0 +1,141 @@
+import pytest
+
+from wemake_python_styleguide.violations.consistency import (
+    SingleCaseMatchViolation,
+)
+from wemake_python_styleguide.visitors.ast.conditions import (
+    SimplifiableMatchVisitor,
+)
+
+# Wrong: single case without guard
+single_case_match = """
+match subject:
+    case {0}:
+        pass
+"""
+
+# Wrong: single case with as binding
+single_case_with_as = """
+match subject:
+    case {0} as x:
+        pass
+"""
+
+# Correct: single case with guard (should not raise)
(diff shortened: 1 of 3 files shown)


Khabib73:
Sorry, but this PR is already being addressed in my PR #3700, which was opened 3 hours earlier. Duplicating work is pointless. I suggest closing this PR, and if you have any suggestions to improve my solution, feel free to leave them in my PR. Moreover it looks like a potential contribution without a human behind it ;)

PR #3711, opened 2026-07-08
Closes #3527

### LINKED ISSUE #3527: Prefer if over match simple sequences and mappings

Khabib73:
# I have made things!

## Checklist

## Related issues

Closes #3527

### DIFF
--- a/tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
+++ b/tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
@@ -24,13 +24,6 @@ match subject:
         pass
 """
 
-simplifiable_guard_match = """
-match subject:
-    case x if x > 0:
-        pass
-    case _:
-        pass
-"""
 
 single_case_sequence = """
 match subject:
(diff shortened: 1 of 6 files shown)


tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
```diff
@@ -115,6 +108,214 @@
...
+match subject:
+    case [True, 1, 'a', False]:
+        pass
+    case _:
+        pass
+"""
+
+sequences_with_constants = """
+match subject:
+    case [some.CONSTANT]:
```
sobolevn: Sorry, this does not seem correct. How can we tell that this is a constant?

tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
```diff
@@ -115,6 +108,214 @@
...
+match subject:
+    case [True, 1, False]:
+        pass
+    case _:
+        pass
+"""
+
+sequences_with_bools_and_strings = """
+match subject:
+    case [True, 'a', False]:
```
sobolevn: Please, also test `None`

tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
```diff
@@ -115,6 +108,214 @@
...
+match subject:
+    case [True, False]:
+        pass
+    case _:
+        pass
+"""
+
+sequences_with_nums_and_strings = """
+match subject:
+    case [1, 'a', 2]:
```
sobolevn: Maybe we can format this one template with different patterns?

tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
```diff
@@ -115,6 +108,214 @@
...
+        sequences_with_bools,
+        sequences_with_constants,
+        sequences_with_nums_and_strings,
+        sequences_with_bools_and_nums,
+        sequences_with_bools_and_strings,
+        sequences_with_bools_and_nums_and_strings,
+    ],
+)
+def test_simplifiable_sequence(
+    assert_errors, parse_ast_tree, default_options, code
```
sobolevn:
```suggestion
    assert_errors, parse_ast_tree, default_options, code,
```

nit

you can enforce this rule in a separate PR, btw: see `COM812`

Khabib73: @sobolevn ruff format conflicting with ruff checking:

```
"COM812", # trailing comma, conflicts with `ruff format`
```

Seems like this problem is still not fixed. Or do you mean to write own rule with WPS?

tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
Khabib73: sure!

tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
sobolevn: Yes, this is fine :)
We use `COM812` everyone in our projects (because this is how it should be by default).

sobolevn (approved):
Thanks!

tests/test_visitors/test_ast/test_conditions/test_simplified_match_conditions.py
sobolevn: You can do that in a separate PR :)

PR #3731, opened 2026-07-20
Add tests for len(generator) usage

### LINKED ISSUE #864: Forbid len from lazy comprehension
# Rule request

## Thesis

Iterators have no `len`, and sometimes I forgetting it.

Bad:

```python
len(1 for el in a if el in b)
```

Better:

```python
len([1 for el in a if el in b])
```

Good:

```python
sum(1 for el in a if el in b)
```

Twice slower, but also ok:

```python
sum(el in b for el in a)
```

## Reasoning

Detect runtime `TypeError` in advance. We could also detect `len` from `yield`-like iterators, but resolving symbols in python always is a non-trivial thing, unfortunately.

Drona-jadhav7:
# I have made things!

## Checklist

## Related issues

Closes #864

## Summary

This pull request adds a new `LenGeneratorViolation` to detect the use of `len()` with generator expressions.

Calling `len()` on a generator expression raises a `TypeError` because generators do not implement `__len__`. This rule encourages using a list comprehension when the length of the generated values is required.

### Example

```python
# Wrong
len(x for x in items)

# Correct
len([x for x in items])
```

The implementation includes:
- A new `LenGeneratorViolation` (`WPS537`)
- A new `_check_len_generator()` method in `WrongFunctionCallContextVisitor`
- Tests covering valid and invalid usages


sobolevn:
Sorry, I would prefer to reject this. `mypy` catches this much better :) 

```
main.py:1: error: Argument 1 to "len" has incompatible type "Generator[int, None, None]"; expected "Sized"  [arg-type]
```

Drona-jadhav7:
> Sorry, I would prefer to reject this. `mypy` catches this much better :)
> 
> ```
> main.py:1: error: Argument 1 to "len" has incompatible type "Generator[int, None, None]"; expected "Sized"  [arg-type]
> ```

Thanks for the feedback! Got it, makes sense since mypy already handles this.
I’ll look for another issue to contribute to.

Issue #3748, opened 2026-07-30
ValueError: ('Error message was not formatted', <wemake_python_styleguide.violations.complexity.OverusedStringViolation object at 0x...>)


sobolevn:
```
» flake8 .
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/Users/sobolev/.pyenv/versions/3.11.9/lib/python3.11/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
                    ^^^^^^^^^^^^^^^^^^^
  File "/Users/sobolev/Desktop/django-modern-rest/.venv/lib/python3.11/site-packages/flake8/checker.py", line 82, in _mp_run
    ).run_checks()
      ^^^^^^^^^^^^
  File "/Users/sobolev/Desktop/django-modern-rest/.venv/lib/python3.11/site-packages/flake8/checker.py", line 526, in run_checks
    self.run_ast_checks()
  File "/Users/sobolev/Desktop/django-modern-rest/.venv/lib/python3.11/site-packages/flake8/checker.py", line 428, in run_ast_checks
    for line_number, offset, text, _ in runner:
  File "/Users/sobolev/Desktop/django-modern-rest/.venv/lib/python3.11/site-packages/wemake_python_styleguide/checker.py", line 165, in run
    yield from (
  File "/Users/sobolev/Desktop/django-modern-rest/.venv/lib/python3.11/site-packages/wemake_python_styleguide/checker.py", line 166, in <genexpr>
    (*error.node_items(), type(self))
      ^^^^^^^^^^^^^^^^^^
  File "/Users/sobolev/Desktop/django-modern-rest/.venv/lib/python3.11/site-packages/wemake_python_styleguide/violations/base.py", line 151, in node_items
    return (*self._location(), self.message())
                               ^^^^^^^^^^^^^^
  File "/Users/sobolev/Desktop/django-modern-rest/.venv/lib/python3.11/site-packages/wemake_python_styleguide/violations/base.py", line 145, in message
    raise ValueError('Error message was not formatted', self)
ValueError: ('Error message was not formatted', <wemake_python_styleguide.violations.complexity.OverusedStringViolation object at 0x1129eb4d0>)
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/sobolev/Desktop/django-modern-rest/.venv/bin/flake8", line 10, in <module>
    sys.exit(main())
[...]

Issue #3745, opened 2026-07-27
Fix `WPS226` error message


sobolevn:
It is now:

`WPS226 Found string literal over-use: any > 3`

It should be:

`WPS226 Found string literal over-use: 'any' 4 > 3`

Khabib73:
Could you please assign this issue to me :)

Khabib73:
Almost done

PR #3751, opened 2026-07-31
Version 1.7.1

sobolevn:


CHANGELOG.md
```diff
@@ -16,11 +16,14 @@ Semantic versioning in our case means:
...
 
+
 ## 1.7.1
 
 ### Bugfixes
 
-- Fixes `WPS226`. Now the error message includes the quoted string value and usage count, #3745
+- Fixes `WPS226`. Now the error message includes
+  the quoted string value and usage count, #3745
+- Fixes crash on overused `'{0}'` string literal, #3748
```
Khabib73, line 26: nit: I actually think that #3748 didn't fix the crash on the overused `'{0}'`, because we just added a new test case there. Real fix for this bug was implemented in #3745 :)

PR #3753, opened 2026-08-05
docs: document recommended per-file-ignores for test files (#1120)

### LINKED ISSUE #1120: Rule 442 conflicts with pytest fixtures usage
When using [pytest fixtures](https://docs.pytest.org/en/latest/fixture.html), you are supposed to just declare them as input arguments, and they will be automagically loaded. For example:

```python
import pytest
from myapp import meaning_of_life


@pytest.fixture
def foo():
    return 42

def test_meaning_of_life(foo):
    meaning = meaning_of_life()
    assert meaning == foo
```

This will violate rule 442, which is not desired in this case.

One could argue that fixtures should live on a different file than tests, but you run on the same issue if you have fixtures dependent on other fixtures:

```python
import pytest


@pytest.fixture
def foo():
    return 42


@pytest.fixture
def bar(foo):
    return foo + 1
```

I think we need to make this feature "pytest aware", or just document that it there is conflict with it and might need to be disabled.

arun-pandian-p:
Closes #1120

Summary of Changes:
- Added a dedicated section under Configuration docs for `per-file-ignores` in test files.
- Provided a standard `setup.cfg` configuration example for pytest test files.
- Explained individual violation rules (`S101`, `WPS432`, `WPS211`, `WPS442`, etc.) that are commonly bypassed in test suites due to pytest framework conventions.

### DIFF
--- a/docs/pages/usage/configuration.rst
+++ b/docs/pages/usage/configuration.rst
@@ -40,8 +40,8 @@ You can ignore:
 3. Some specific violation with the full violation code
 
 Use `per-file-ignores <https://flake8.pycqa.org/en/latest/user/options.html?highlight=per-file-ignores#cmdoption-flake8-per-file-ignores>`_
-option, so it is possible to ignore violations on a per-file bases.
-It means, that you can have different set of violations
+option, so it is possible to ignore violations on a per-file basis.
+It means that you can have a different set of violations
 ignored for different files.
 
 Example:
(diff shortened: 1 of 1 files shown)


Khabib73:
sorry, but this rule has been disabled

arun-pandian-p:
Thanks for pointing that out @Khabib73! I see WPS442 was disabled in #3210.
Should I remove WPS442 from this doc update and keep the rest of the recommended test file ignores (like `S101`, `WPS432`, `WPS211`, etc.), or would you prefer to close #1120 as obsolete?

docs/pages/usage/configuration.rst
```diff
@@ -51,8 +51,23 @@ Example:
...
+    tests/*.py: D103, S101, S105, WPS118, WPS202, WPS210, WPS211, WPS336, WPS432, WPS442
+
+.. rubric:: Ignoring violations in test files
+
+Test files often require different linting rules than production code due to testing frameworks like ``pytest`` (which use fixtures as function arguments, magic numbers for assertions, etc.).
+
+Common violations ignored in test files:
+
+* ``S101``: Allows using ``assert`` statements (required for testing).
+* ``S105``: Allows hardcoded password strings in test data.
```
Khabib73: It seems these are ruff rules. How do they relate to flake8? 🤔

arun-pandian-p: Thanks @Khabib73!

S101 and S105 are actually from flake8-bandit (the bandit plugin for flake8), which is one of the core plugins bundled with wemake-python-styleguide. Ruff adopted bandit's Sxxx prefix convention later on to maintain compatibility with existing flake8 configurations.

In pytest suites, flake8-bandit flags S101 for assert statements and S105 for mock hardcoded credentials in test data.

I will also remove WPS442 from the documentation since it was disabled in #3210, and update the section to clarify which plugin each rule code belongs to (wemake-python-styleguide vs flake8-bandit).

Khabib73: Thanks for the explanation! Just to be safe, let's get a final look from @sobolevn, since I don't maintain this project and want to be sure this change aligns with the maintainer's vision :)

arun-pandian-p: Thanks @Khabib73! 

I've updated the PR with the following changes:
- Removed `WPS442` (since it was disabled in #3210).
- Categorized all recommended test violation codes by their plugin source (`flake8-bandit`, `wemake-python-styleguide`, and `flake8-docstrings`) for clarity.

@sobolevn Whenever you have a moment, please take a look when you get a chance!

sobolevn (commented):
Thanks for the PR!

However, I don't think that it is worth it: `flake8-bandit` is not bundled for 2 years now.

We now use `ruff` for this.

PR #3767, opened 2026-08-17
Update Makefile

Khabib73:
I think this will be useful. Sometimes I have problems with mypy and I have to use `make lint` which runs flake8 and ruff


Makefile
sobolevn:
```suggestion
test: lint type-check unit package ## Run all checks (lint, unit, package)
```

Khabib73: I understand your intent, but if lint already runs mypy, adding type-check to the test target will run mypy twice. I suggest either removing mypy from lint (so type-check runs only once in test) or removing type-check in test (Leave it as it is)

sobolevn (approved):

Khabib73:
Thanks!

PR #3784, opened 2026-08-30
Remove the Spellcheckers section from CONTRIBUTING.md

### LINKED ISSUE #3783: CONTRIBUTING.md: `codespell` command lists `styles`, which no longer exists
i ran the spellcheck commands from `CONTRIBUTING.md`, and the `codespell` one lists a `styles` diirectory:

    codespell -w wemake_python_styleguide tests docs scripts styles *.md --ignore-words ./tests/whitelist.txt

There is no `styles/` in the repo. It was removed in #3132 (commit d646837c) along with the preset files it held. `CONTRIBUTING.md` dates to #1799 and not updated.

tbh nothing actually breaks, `codespell` ignores a path that doesn't exist without any warning and still exits 0, so the command works fine. Tho a new contributor (me for e.g) will get confused when looking in `styles/` in my opinion.

```bash
$ ls -d styles
ls: cannot access 'styles': No such file or directory

$ git log --oneline --diff-filter=D -- 'styles/*'
d646837c Drops `nitpick` support (#3132)

$ codespell styles; echo "exit=$?"
exit=0
```

I would suggest just dropping `styles` from the command. Happy to send a PR if that sounds right.

One thing I noticed while checking this, not part of the above: the `flake8-spellcheck` command in the same section

    flake8 --whitelist ./tests/whitelist.txt .

produces no output, because `setup.cfg` has `select = WPS, E999` and `flake8-spellcheck` reports `SC` codes. Running the same command with `--select=SC` does report `SC200` warnings. I'm not sure what the intended fix is here (adjust the documented command, change `select`, or drop the section), so I left it out — happy to open a separate issue if it's worth tracking.

TokenStyle:
Removes the Spellcheckers section from `CONTRIBUTING.md`.

proposed in #3783, `codespell` runs in `pre-commit` now. Also removed the `flake8-spellcheck` command, which reports nothing because `setup.cfg` has `select = WPS, E999`.

FYI the `pre-commit` hook still uses `[tool.codespell]` and works fine.

## Checklist

  N/A: documentation only, no code changes.
  N/A: documentation only. Happy to add an entry if you'd prefer one.

## Related issues

Closes #3783

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -112,32 +112,6 @@ mypy wemake_python_styleguide
 This step is mandatory during the CI.
 
 
-## Spellcheckers
-
-This project is developed by a diverse and multilingual group of people.
-Many of us are not English native speakers and we also know that people can make mistakes and typos even in the simplest of words.
-
-So, that's why we use a bunch of tools to find and fix spelling and grammar.
-
-You will need to install them manually, because we don't ship them with the dependencies:
-
-```bash
-pip install codespell flake8-spellcheck
-```
-
-And then you can use them:
-
-```bash
-# codespell:
-codespell -w wemake_python_styleguide tests docs scripts styles *.md --ignore-words ./tests/whitelist.txt
-
-# flake8-spellcheck:
-flake8 --whitelist ./tests/whitelist.txt .
-```
-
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thanks!

PR #3804, opened 2026-09-10
Do not count docstrings for WPS226

### LINKED ISSUE #3803: Do not count docstrings for `WPS226`
```
dmr/security/jwt/views.py

  155:9    WPS226 Found string literal over-use: 'Lazy endpoint spec.' 6 > 3
  """Lazy endpoint spec."""
  ^
```

I just got this, where `"""Lazy endpoint spec."""` is an actual docstring.
This is not right.

vyhuholl:
# I have made things!

`WPS226` counted docstrings as string literal usages. Four functions sharing one docstring were enough to trigger it:

```
repro.py:2:5: WPS226 Found string literal over-use: 'Lazy endpoint spec.' 4 > 3
```

The violation asks you to move an overused literal into a named constant. You can't do that to a docstring, so it shouldn't be counted at all.

`logic/tree/strings.py` already has `is_doc_string()`, but it takes the statement (`ast.Expr`), and the visitor works with the `ast.Constant` inside it. That function has four other call sites, so I left it alone and put `is_doc_string_value()` next to it. The definition is the one the old docstring already states: the first statement of the nearest module, class or function.

```python
statement = get_parent(node)
if statement is None or not is_doc_string(statement):
    return False
context = get_context(statement)
return context is not None and context.body[0] is statement
```

`StringOveruseVisitor._check_constant` returns early on it, next to the existing annotation and f-string checks. Bytes go through the same method and always fail the check, since bytes can't be a docstring.

Only `body[0]` is treated as a docstring. A bare string statement sitting after the real docstring still counts, and there's a test for that.

One case this doesn't cover: PEP 258 attribute docstrings are still counted. Different shape, and I'd rather not widen the fix here — I can open a separate issue if you think it's worth handling.

## AI Policy

## Checklist

## Related issues

Closes #3803 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Semantic versioning in our case means:
 ### Bugfixes
 
 - Count overused str and bytes separately for `WPS226`, 3782
+- Do not count docstrings for `WPS226`, #3803
 - Adds direct links to violations when searching by code, #3790
 
 ### Misc
--- a/tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_string.py
+++ b/tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_string.py
@@ -131,6 +131,44 @@ tstring_same_prefix2 = pytest.param(
     ),
 )
 
+# See:
+# https://github.com/wemake-services/wemake-python-styleguide/issues/3803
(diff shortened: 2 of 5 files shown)


sobolevn (approved):
Thanks!

PR #3806, opened 2026-09-10
Do not count attribute docstrings for WPS226

### LINKED ISSUE #3805: `WPS226` counts PEP 258 attribute docstrings
With docstrings excluded from `WPS226` (#3804), functions, classes and modules are fine, but the attribute form still counts:

```python
class Some:
    first = 1
    """Some field docs."""

    second = 2
    """Some field docs."""

    third = 3
    """Some field docs."""

    fourth = 4
    """Some field docs."""
```

```
attr_doc.py:3:5: WPS226 Found string literal over-use: 'Some field docs.' 4 > 3
```

Same reasoning as #3803: extracting that string into a constant stops it from being documentation. PEP 258 calls it an attribute docstring and Sphinx `autodoc` picks it up.

It isn't covered by the fix for #3803 because that one defines a docstring as `body[0]` of the nearest module, class or function, which is what `is_doc_string()` in `logic/tree/strings.py` has always meant. An attribute docstring is the statement after an assignment, so it never matches.

I'm genuinely unsure this is worth fixing. This project documents its own module attributes with `#:` comments (147 of those lines in `wemake_python_styleguide/`), so the style guide doesn't obviously owe support to the other convention. But it's the same false positive as #3803, just one line lower.

If you do want it, the change looks narrow: also skip a string `ast.Expr` that directly follows an assignment in the same body. I can pick it up.

vyhuholl:
# I have made things!

`WPS226` still counted PEP 258 attribute docstrings after #3804. Four class attributes sharing one:

```
attr_doc.py:3:5: WPS226 Found string literal over-use: 'Some field docs.' 4 > 3
```

Same reasoning as #3803. The string documents the assignment above it, so moving it into a named constant would stop it from being documentation.

`is_doc_string_value()` accepted exactly one position, `body[0]` of the nearest module, class or function. It now accepts a second, and the lookup moved into a helper:

```python
def _is_doc_string_place(body, statement) -> bool:
    """Docstrings open a definition's body or follow an assignment."""
    if body[0] is statement:
        return True
    return any(
        current is statement and isinstance(previous, AssignNodes)
        for previous, current in itertools.pairwise(body)
    )
```

`AssignNodes` is the existing `(ast.Assign, ast.AnnAssign)` tuple, so `first = 1`, `second: int = 2` and a bare `third: int` all work. I left `ast.AugAssign` out, since `count += 1` followed by a string isn't documenting anything.

The search stays inside the context's own body. `self.third = 3` in `__init__` is covered, because the context there is the method. A string after an assignment nested inside an `if` still counts, same as before, and so does one that follows anything other than an assignment. Both have tests.

One thing that looks odd in the diff: `is_doc_string_value()` now ends with `context is not None and _is_doc_string_place(...)` instead of an early return. Writing it as `if context is None: return False` drops coverage below 100%, because `get_context()` returns `None` only for the module node itself, and that can never be a docstring statement.

## AI Policy

## Checklist

## Related issues

Closes #3805 

### DIFF
--- a/tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_string.py
+++ b/tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_string.py
@@ -209,7 +209,7 @@ def third():
     {0}
 
 def fourth():
-    print(4)
+    x = 1
     {0}
 """
 
--- a/tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_string.py
+++ b/tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_string.py
@@ -1,6 +1,6 @@
 import pytest
 
-from wemake_python_styleguide.compat.constants import PY314
+from wemake_python_styleguide.compat.constants import PY312, PY314
 from wemake_python_styleguide.violations.complexity import (
     OverusedStringViolation,
 )
@@ -213,6 +213,36 @@ def fourth():
     {0}
 """
 
+# Type aliases are documented the very same way, see:
+# https://discuss.python.org/t/docstrings-for-type-aliases/108901
+explicit_type_alias_docstrings = """
+Timeout: TypeAlias = float
+{0}
+
+Retries: TypeAlias = int
+{0}
+
+Backoff: TypeAlias = float
+{0}
+"""
+
+type_alias_docstrings = pytest.param(
+    """
+    type Timeout = float | None
+    {0}
+
+    type Retries = int
+    {0}
+
+    type Backoff = float
+    {0}
+    """,
+    marks=pytest.mark.skipif(
+        not PY312,
+        reason='`type` aliases are only in Python 3.12+',
+    ),
+)
+
 EXPECTED_LOCATION = (2, 8)
 
 
@@ -419,6 +449,8 @@ def test_common_strings_allowed(
         function_docstrings,
         module_attribute_docstrings,
         class_attribute_docstrings,
+        explicit_type_alias_docstrings,
+        type_alias_docstrings,
     ],
 )
 @pytest.mark.parametrize(
--- a/tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_string.py
+++ b/tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_string.py
@@ -213,6 +213,26 @@ def fourth():
     {0}
 """
 
+# `x.some = 1` defines an attribute of `x`, not of the module, the class,
+# or the instance we are in. So, there's nothing here to document.
+foreign_attribute_docstrings = """
+x.some = 1
+{0}
+
+def first():
+    x.some = 1
+    {0}
+
+class Some:
+    def second(self, x):
+        x.some = 1
+        {0}
+
+    def third(self):
+        self.some.other = 1
+        {0}
+"""
+
 # Type aliases are documented the very same way, see:
 # https://discuss.python.org/t/docstrings-for-type-aliases/108901
 explicit_type_alias_docstrings = """
@@ -482,6 +502,7 @@ def test_docstrings_not_counted(
     [
         not_a_docstring,
         not_an_attribute_docstring,
+        foreign_attribute_docstrings,
     ],
 )
 @pytest.mark.parametrize(
--- a/tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_string.py
+++ b/tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_string.py
@@ -169,6 +169,50 @@ def fourth():
     {0}
 '''
 
+# See:
+# https://github.com/wemake-services/wemake-python-styleguide/issues/3805
+module_attribute_docstrings = """
+first = 1
+{0}
+
+second: int = 2
+{0}
+
+third: int
+{0}
+"""
+
+class_attribute_docstrings = """
+class Some:
+    first = 1
+    {0}
+
+    second: int = 2
+    {0}
+
+    def __init__(self):
+        self.third = 3
+        {0}
+"""
+
+not_an_attribute_docstring = """
+def first():
+    print(1)
+    {0}
+
+def second():
+    print(2)
+    {0}
+
+def third():
+    print(3)
+    {0}
+
+def fourth():
+    print(4)
+    {0}
+"""
+
 EXPECTED_LOCATION = (2, 8)
 
 
@@ -373,6 +417,8 @@ def test_common_strings_allowed(
     [
         module_docstring,
         function_docstrings,
+        module_attribute_docstrings,
+        class_attribute_docstrings,
     ],
 )
 @pytest.mark.parametrize(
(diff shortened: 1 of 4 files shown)


sobolevn (commented):
Please, also test type-aliases, see https://discuss.python.org/t/docstrings-for-type-aliases/108901 for the context

vyhuholl:
All addressed, and two of the notes turned into bigger changes than they looked.

`views.py` is now a package: `base.py`, `body.py`, `cookie.py`, with
`__init__.py` re-exporting everything, so `dmr.security.jwt.views.X` keeps
working. The shared bases moved to `base.py` and lost their leading underscore
(`BaseTokenController`, `BaseRefreshTokenController`), because pyright rejects
importing `_`-prefixed names across modules. They are not re-exported from the
package, so the public surface is unchanged.

`check_auth` now takes `(user: AbstractBaseUser, token: JWToken)` on the cookie
controllers and on the four body ones as well. Two more things were needed
before the mixins actually worked:

- `RefreshTokenSyncController` and `RefreshTokenAsyncController` had no
  `get_user`, the lookup was inlined into `refresh`, and `blocklist()` calls it.
  It is now a method there too, like the verify controllers have.
- `_BaseBlocklistMixin.__init__` assumed an auth object and set
  `self.require_claims`, which a controller does not have. It now skips that for
  controllers. A token without `jti` is still rejected, `token_jti` raises for it.

There are five tests for this in `test_blocklist_views.py`: a blocklisted token
cannot refresh, sync and async, cookie and body flow, and a token without `jti`
is rejected. Documented in the blocklist section with an example.

On `secure`: I would keep `True`. Browsers treat `http://localhost` and
`http://127.0.0.1` as secure contexts, so `Secure` cookies are stored there and
local development works as is. `not settings.DEBUG` would make the security of
the flow depend on a flag that is wrong in staging often enough. The case it
does break is a dev environment served over plain HTTP on some other host name,
and the docs now say to set `jwt_cookie_secure = settings.DEBUG is False`
yourself there.

The rest:

- `reverse_lazy` in the test app. `CookieSpec.path` and `NewCookie.path` now
  accept lazy strings, otherwise users have to hardcode the refresh url the same
  way. The docs examples use `reverse_lazy` too.
- `cast` replaced with `type: ignore` in both `make_api_response`.
- `@sensitive_variables` dropped from the sync `login` and `refresh`, added to
  the async `get_user` and to both logout endpoints.
- The inline code block in `jwt.rst` is a separate example file now.
- Test for a redefined `validate_spec`. Writing it showed that an override had
  no way to keep the `Cache-Control` header without reaching for a private
  constant, so the cookie controllers got `response_headers` and
  `response_headers_spec` next to the cookie spec hooks.

Breaking changes and the new methods are in the changelog.

(CI is failing because it's failing on master now, cause — commit 900b7d77)

sobolevn:
🤔

vyhuholl:
Sorry, that was comment for the wrong PR. 🙈

tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_string.py
```diff
@@ -169,6 +169,50 @@ def fourth():
...
+def second():
+    print(2)
+    {0}
+
+def third():
+    print(3)
+    {0}
+
+def fourth():
+    print(4)
```
sobolevn:
```suggestion
    x = 1
```

tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_string.py
```diff
@@ -169,6 +169,50 @@ def fourth():
...
+def first():
+    print(1)
+    {0}
+
+def second():
+    print(2)
+    {0}
+
+def third():
+    print(3)
```
sobolevn:
```suggestion
    y: int = 0
```

tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_string.py
```diff
@@ -169,6 +169,50 @@ def fourth():
...
+{0}
+
+second: int = 2
+{0}
+
+third: int
+{0}
+"""
+
+class_attribute_docstrings = """
```
sobolevn, line 185: please, don't forget about `type` aliases on `3.12+`, they should also be treated as targets for docstrings

tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_string.py
```diff
@@ -169,6 +169,50 @@ def fourth():
...
+        {0}
+"""
+
+not_an_attribute_docstring = """
+def first():
+    print(1)
+    {0}
+
+def second():
+    print(2)
```
sobolevn:
```suggestion
    x, y = call()
```

tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_string.py
```diff
@@ -169,6 +169,50 @@ def fourth():
...
+    {0}
+
+    def __init__(self):
+        self.third = 3
+        {0}
+"""
+
+not_an_attribute_docstring = """
+def first():
+    print(1)
```
sobolevn:
```suggestion
    x = y = 1
```

tests/test_visitors/test_ast/test_complexity/test_overuses/test_overused_string.py
vyhuholl: Done in [#957cbea3 ](https://github.com/vyhuholl/wemake-python-styleguide/commit/957cbea3cc867b9bc59e0b727587ebd3fdf5b952)

wemake_python_styleguide/logic/tree/strings.py
```diff
@@ -26,12 +33,52 @@ def is_doc_string_value(node: ast.AST) -> bool:
...
+        return isinstance(statement, nodes.TypeAlias)
+    if len(targets) != 1:  # `x = y = 1` defines no single attribute
+        return False
+    if isinstance(context, FunctionNodes):
+        # Locals are not attributes, only `self.some = 1` is one.
+        return isinstance(targets[0], ast.Attribute)
+    return isinstance(targets[0], _AttributeTargets)
+
+
+def _get_assign_targets(statement: ast.stmt) -> list[ast.expr]:
```
sobolevn: Will `get_assign_targets` work here?

vyhuholl: Yes, it works — committed as https://github.com/vyhuholl/wemake-python-styleguide/commit/0e593d0cc76b39532ddce08b1314e6255a6469d4. The private helper was redundant. Thank you for noticing!

wemake_python_styleguide/logic/tree/strings.py
```diff
@@ -26,12 +34,43 @@ def is_doc_string_value(node: ast.AST) -> bool:
...
+def _is_documented(statement: ast.stmt, context: ContextNodes) -> bool:
+    """Only type aliases and single attribute definitions are documented."""
+    if not isinstance(statement, AssignNodes):
+        return isinstance(statement, nodes.TypeAlias)  # `type X = int`
+    targets = get_assign_targets(statement)
+    if len(targets) != 1:  # `x = y = 1` defines no single attribute
+        return False
+    if isinstance(context, FunctionNodes):
+        # Locals are not attributes, only `self.some = 1` is one.
+        return isinstance(targets[0], ast.Attribute)
```
sobolevn: this does not seem fully correct. currently this code:

```python
def some():
    x.y = 1
    """Docstring."""
```

counts as a docstring, while it is clearly not 🤔 
not sure what is the best way forward here. 

maybe we can just test this corner case with a comment that it is not fully correct, but it is hard to define an attribute docstring the right way.

using class as a context is not a solution as well, because this code would pass:

```python
class A:
    def some(self, x):
        x.y = 1
        """Docstring."""
```

btw, let's maybe create a rule for placing docstings? this would solve the issue by enforcing the correct places for docstings.

vyhuholl: Fixed in https://github.com/vyhuholl/wemake-python-styleguide/commit/019a1ea4b1db5e7cce3cafd0cd8e7d03ec880532

A rule for placing docstrings is a genuinely good idea and a much better fix than any of this — enforce placement, and the definition problem goes away. I'll open a new issue about this.

sobolevn (approved):
Now this seems correct, thank you :)

sobolevn:
Please, don't forget about new rule request for `docstring` positions :)

PR #3810, opened 2026-09-12
Add WPS484: forbid strings that document nothing

### LINKED ISSUE #3808: Forbid strings that look like docstrings but are not
# Rule request

## Thesis

Add a violation for a string statement placed where it cannot document anything.

```python
def some(self):
    x.y = 1
    """Documents nothing, `y` belongs to `x`."""

    if condition:
        """Neither does this."""
        return 1
```

A string statement documents something only in these places:

- the first statement of a module, class or function body
- after an assignment to a plain name at module or class level
- after `self.some = ...` inside a method
- after a `type Some = ...` alias on 3.12+, see [the discussion](https://discuss.python.org/t/docstrings-for-type-aliases/108901)

Anywhere else it does nothing at runtime while reading like documentation.

## Reasoning

This came out of review on #3806, [here](https://github.com/wemake-services/wemake-python-styleguide/pull/3806#discussion_r3995371446).

`WPS226` skips docstrings, so it has to decide whether a string documents anything, and `_is_documented()` in `logic/tree/strings.py` became that decision. It handles the cases we tested, but the definition is fuzzy, and a plain function whose first argument is named `self` still fools it.

Nothing catches the misplaced strings today. `StatementHasNoEffectViolation` (`WPS428`) was retired in 1.0.0 and delegated to ruff, and neither of the rules it points at flags a string:

```
$ ruff check --isolated --select B015,B018 misplaced.py
All checks passed!
```

Enforcing placement doesn't make the definition any easier to write. It moves it somewhere visible, where a wrong guess produces an error a user can argue with, rather than silently deciding whether `WPS226` fires. And once placement is enforced, `WPS226` can stop guessing: every string statement that survives the new rule is in a documenting position, so it can exempt all of them.

New violation, so this lands in a minor release rather than a patch.

vyhuholl:
# I have made things!

Adds `WPS484`, the rule from #3808. It flags a string statement placed where it cannot document anything:

```python
some.first = 1
"""Documents nothing, `first` belongs to `some`."""


def some(arg):
    """Function docs."""
    if arg:
        """Neither does this."""
        return arg
    return 0
```

A string documents something in exactly four places:

- opening a module, class or function body
- after an assignment to a plain name in a module or a class
- after `self.some = ...` inside a method
- after a `type Some = ...` alias on `python3.12+`

The check itself is a single condition, because `WPS226` already had to answer this same question in order to skip docstrings:

```python
def _check_doc_string_place(self, node: ast.Expr) -> None:
    if strings.is_doc_string(node) and not strings.is_doc_string_value(
        node.value,
    ):
        self.add_violation(WrongDocStringPlacementViolation(node))
```

That reuse is the point of the rule, not a shortcut. Both rules now read one definition of a documenting position, so refining it later moves them together instead of letting two copies drift.

The violation sits in `best_practices.py` next to `WPS428`, the old "statement has no effect" rule that was retired to ruff's `B015`/`B018` in 1.0.0. Neither of those flags a string, which is how these slip through today.

`wemake-python-styleguide` itself is clean under the new rule.

### Open to review
I picked these three without asking, so say if any is wrong:

- the name, `WrongDocStringPlacementViolation`
- the code, `WPS484`
- counting `type Some = int` docstrings as a valid position, given they're still only a proposal upstream

## AI Policy

## Checklist

## Related issues

Closes #3808

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@ Semantic versioning in our case means:
 
 ## WIP
 
+### Features
+
+- Adds `WPS484`: forbid strings that look like docstrings
+  but document nothing, #3808
+
 ### Bugfixes
 
 - Count overused str and bytes separately for `WPS226`, 3782
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,12 +17,15 @@ Semantic versioning in our case means:
   change the client facing API, change code conventions significantly, etc.
 
 
-## 1.8.1
+## WIP
 
 ### Features
 
 - Adds `WPS484`: forbid strings that look like docstrings
   but document nothing, #3808
+
+
+## 1.8.1
 
 ### Bugfixes
 
--- a/tests/test_visitors/test_ast/test_statements/test_doc_string_placement.py
+++ b/tests/test_visitors/test_ast/test_statements/test_doc_string_placement.py
@@ -8,26 +8,26 @@ from wemake_python_styleguide.visitors.ast.statements import (
     DocStringPlacementVisitor,
 )
 
-module_docstring = "'Docs.'"
+module_docstring = '{0}'
 
 function_docstring = """
 def some():
-    'Docs.'
+    {0}
 """
 
 class_docstring = """
 class Some:
-    'Docs.'
+    {0}
 """
 
 module_attribute = """
 first = 1
-'Docs.'
+{0}
 """
 
 annotated_module_attribute = """
 first: int = 1
-'Docs.'
+{0}
 """
 
 class_attribute = """
@@ -35,7 +35,7 @@ class Some:
     'Class docs.'
 
     first = 1
-    'Docs.'
+    {0}
 """
 
 instance_attribute = """
@@ -43,13 +43,21 @@ class Some:
     def __init__(self):
         'Method docs.'
         self.first = 1
-        'Docs.'
+        {0}
+"""
+
+instance_attribute_in_new = """
+class Some:
+    def __new__(cls):
+        'Method docs.'
+        cls.first = 1
+        {0}
 """
 
 type_alias = pytest.param(
     """
     type Some = int
-    'Docs.'
+    {0}
     """,
     marks=pytest.mark.skipif(
         not PY312,
@@ -59,7 +67,7 @@ type_alias = pytest.param(
 
 foreign_attribute = """
 some.first = 1
-'Docs.'
+{0}
 """
 
 instance_attribute_outside_constructor = """
@@ -67,26 +75,26 @@ class Some:
     def method(self):
         'Method docs.'
         self.first = 1
-        'Docs.'
+        {0}
 """
 
 local_variable = """
 def some():
     'Function docs.'
     first = 1
-    'Docs.'
+    {0}
 """
 
 multiple_targets = """
 first = second = 1
-'Docs.'
+{0}
 """
 
 inside_condition = """
 def some(arg):
     'Function docs.'
     if arg:
-        'Docs.'
+        {0}
         return 1
     return 0
 """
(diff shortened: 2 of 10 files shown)


sobolevn (commented):
The first correction:

> after `self.some = ...` inside a constructor 

Not in every other method

:)

tests/test_visitors/test_ast/test_statements/test_doc_string_placement.py
```diff
@@ -0,0 +1,160 @@
...
+local_variable = """
+def some():
+    'Function docs.'
+    first = 1
+    'Docs.'
+"""
+
+multiple_targets = """
+first = second = 1
+'Docs.'
```
sobolevn: please, test different styles of docstrings: `"Doc"`, `'Doc'`, `"""Doc"""`, `'''Doc'''`

wemake_python_styleguide/logic/tree/strings.py
```diff
@@ -66,9 +67,13 @@ def _is_documented(statement: ast.stmt, context: ContextNodes) -> bool:
         return False
     target = targets[0]
     if isinstance(context, FunctionNodes):
-        # Locals are not attributes. Only `self.some = 1` defines one,
-        # while `x.some = 1` documents an attribute of some other object.
-        return isinstance(target, ast.Attribute) and is_special_attr(target)
+        # Only a constructor defines the instance attributes. Locals are
+        # not attributes, and `x.some = 1` belongs to some other object.
+        return (
+            context.name == INIT
```
sobolevn: `__new__` can also define attributes

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -3117,3 +3117,46 @@ class ForbidMappingProxyTypeViolation(ASTViolation):
         'Found a `types.MappingProxyType` usage, prefer `frozendict` on 3.15+'
     )
     code = 483
+
+
+@final
+class WrongDocStringPlacementViolation(ASTViolation):
+    """
```
sobolevn, line 3124: use `'''` here

wemake_python_styleguide/violations/best_practices.py
```diff
@@ -3117,3 +3117,46 @@ class ForbidMappingProxyTypeViolation(ASTViolation):
...
+
+    Solution:
+        Move the string to a place where it documents something,
+        or turn it into a regular ``#`` comment.
+
+    Example::
+
+        # Correct:
+        first = 1
+        \"\"\"Documents ``first``.\"\"\"
```
sobolevn:
```suggestion
        """Documents ``first``."""
```

so you can use `"""` here

CHANGELOG.md
```diff
@@ -19,6 +19,11 @@ Semantic versioning in our case means:
 
 ## 1.8.1
 
+### Features
```
sobolevn, line 22: This is misplaced, it should start a new `## WIP` section.

sobolevn (commented):
👍

tests/test_visitors/test_ast/test_statements/test_doc_string_placement.py
vyhuholl: Done.

wemake_python_styleguide/violations/best_practices.py
vyhuholl: Done.

CHANGELOG.md
vyhuholl: Moved to a fresh `## WIP` above `1.8.1`.

sobolevn (commented):
Thanks! 

Last fixes: 
- please test dataclass-style attrs explicitly, with and without default values
- there's a common error that people use `"""Docstring"""` before the attribute, not after. we should also explicitly test and report this case

wemake_python_styleguide/logic/tree/strings.py
vyhuholl: Both constructors now count, via a named set rather than a bare comparison. Added tests for the `__new__` constructor too.

vyhuholl:
Added tests for dataclass-style attrs and for attribute docstring placement. Both cases already behaved correctly, so no code change was needed.

sobolevn (approved):
Thanks!

